PR #53, opened 2026-04-07
Use Python logging module everywhere, not print/printf (closes #44)

### LINKED ISSUE #44: Use Python logging module everywhere, not print/printf
work.sh and other bash scripts use \`printf\` to log. The Python side uses the \`logging\` module. Once everything is Python, all logging should go through \`logging\` with:

- Per-thread log prefix: \`[confusio]\`, \`[kennel]\`, \`[supervisor]\`, \`[http]\`, \`[watchdog]\`
- Structured format: \`%(asctime)s %(levelname)-5s [%(threadName)s] %(message)s\`
- File handler: \`~/log/kennel.log\` (single file, all threads)
- Stderr handler when tty
- Log level configurable via \`--log-level\`

Currently kennel's Python code uses \`logging\` but the bash scripts write directly to files (\`~/log/fido.log\`, \`~/log/sync-tasks.log\`, \`~/log/watchdog.log\`). After the rewrite, all logs converge into one \`~/log/kennel.log\` via the logging module.

## Milestone
v1

FidoCanCode:
Sniffs out every `print()` call in our Python modules and replaces them with proper `logging` usage — structured, leveled, and ready for production. The bash scripts still use `printf` for now since they're on the chopping block anyway, but all Python code will speak fluent `logging` by the time this branch is done. Woof! 🐕

Fixes #44.

---

## Work queue

- [x] Replace diagnostic print-to-stderr calls in cli.py with log.* and add caplog assertions to tests
- [x] use module-level logger names derived from file names
</details>

### DIFF
--- a/kennel/cli.py
+++ b/kennel/cli.py
@@ -10,7 +10,7 @@ from pathlib import Path
 from kennel import tasks as _tasks_mod
 from kennel.github import GitHub
 
-log = logging.getLogger("kennel")
+log = logging.getLogger(__name__)
 
 
 class Cmd:
--- a/kennel/config.py
+++ b/kennel/config.py
@@ -5,7 +5,7 @@ import logging
 from dataclasses import dataclass
 from pathlib import Path
 
-log = logging.getLogger("kennel")
+log = logging.getLogger(__name__)
 
 
 @dataclass(frozen=True)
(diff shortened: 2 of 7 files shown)


FidoCanCode:
Good catch! I'll use `logging.getLogger(__name__)` so each module gets a logger named after its file path (like `kennel.server`, `kennel.events`, etc.) — clean hierarchy, easy to filter. On it! 🐕

PR #57, opened 2026-04-07
Fetch sibling review threads for context on terse comments (closes #28)

### LINKED ISSUE #28: Fetch sibling review threads for context on terse comments
When a review comment is terse (e.g. "Here, too.", "Same.", "Ditto", "^"), fido has no idea what it means because the webhook only contains that one comment's context.

## Problem
Reviewer leaves a detailed comment on one function, then "Here, too." on another instance of the same pattern. Fido sees "Here, too." with a diff hunk but no knowledge of what "too" refers to.

## Fix
Before triaging, if the comment is short (< ~20 chars) or looks like a reference ("here too", "same", "ditto", "^", "this too"), fetch all unresolved review threads on the same PR and include them as context for the triage + reply.

\`\`\`python
if is_terse(comment_body):
    other_threads = github.get_review_threads(owner, repo, pr)
    context["sibling_threads"] = other_threads
\`\`\`

This way Opus sees: "The reviewer said 'use match statement' on line 45, and 'Here, too.' on line 72 which has a similar if/elif chain."

## Detection (no LLM)
- Comment body < 20 chars
- Or matches common reference patterns: "same", "here too", "ditto", "^", "this too", "likewise"

## Context enrichment
- Fetch all review threads on the PR (already have GraphQL query for this)
- Include thread bodies + file paths in the triage context
- Opus connects the dots

## Milestone
v1

FidoCanCode:
When a reviewer says "here, too" or "ditto" on a PR thread, fido has no idea what they're referring to because the webhook only carries that single comment — sniffing one thread in isolation doesn't give enough scent! This teaches fido to detect terse comments (short body or common reference patterns like "same", "^", "likewise") and fetch all sibling review threads on the PR so Opus gets the full picture and can connect the dots.

Fixes #28.

---

## Work queue

- [ ] Add is_terse() helper to detect short/reference review comments
- [x] Fetch sibling review threads when comment is terse and add to context
- [x] Render sibling threads in triage_context_block so Opus sees them
- [x] PR comment: move _fetch_sibling_threads and is_terse into DI class per CLAUDE.md
- [x] [replace regex terse detection with Haiku "needs more context?" check](https://github.com/rhencke/kennel/pull/57#discussion_r3048521466)
</details>

### DIFF
--- a/kennel/events.py
+++ b/kennel/events.py
@@ -267,14 +267,14 @@ def reply_to_comment(
     prompts = Prompts(persona)
     comment = action.comment_body
 
-    # Enrich context with sibling threads when the comment is terse
+    # Enrich context with sibling threads when the comment needs more context
     context: dict[str, Any] = dict(action.context) if action.context else {}
-    if is_terse(comment) and info.get("repo") and info.get("pr"):
+    if needs_more_context(comment) and info.get("repo") and info.get("pr"):
         siblings = get_github().fetch_sibling_threads(info["repo"], info["pr"])
         if siblings:
             context["sibling_threads"] = siblings
             log.info(
-                "terse comment — fetched %d sibling thread(s) for context",
+                "needs-more-context comment — fetched %d sibling thread(s) for context",
                 len(siblings),
             )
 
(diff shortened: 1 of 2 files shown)


kennel/events.py
```diff
@@ -336,6 +347,41 @@ def _try_resolve_thread(info: dict[str, Any], config: Config) -> None:
     pass
 
 
+def _fetch_sibling_threads(repo: str, pr: int | str) -> list[dict[str, Any]]:
```
rhencke: Please place into DI classes as per Claude.MD

FidoCanCode: Woof, good catch! I'll move `_fetch_sibling_threads` into the DI class so it's injectable and testable, just like the pattern in CLAUDE.md. On it! 🐕

kennel/events.py
```diff
@@ -393,6 +439,24 @@ def reply_to_review(
             already_replied.add(cid)
 
 
+_TERSE_PATTERNS = frozenset(
```
rhencke: Instead, why not ask Claude haiku "needs more context?"

FidoCanCode: Done — replaced `_TERSE_PATTERNS`/`is_terse()` with `needs_more_context()`, which calls claude-haiku-4-5 to sniff out whether the comment needs sibling thread context! Much better nose on this one. 🐕

PR #89, opened 2026-04-08
Deduplicate task creation: kennel + work.sh both create tasks for same comment (closes #58)

### LINKED ISSUE #58: Deduplicate task creation: kennel + work.sh both create tasks for same comment
Kennel's webhook handler and work.sh's comments handler both create tasks for the same review comment, resulting in duplicate tasks.

## Example
```
pending: move _fetch_sibling_threads into DI class per dependency injection pattern
pending: PR comment: move _fetch_sibling_threads and is_terse into DI class per CLAUDE.md
```

Same request, two tasks.

## Root cause
1. Kennel receives webhook → triages as ACT → creates task in tasks.json
2. work.sh loops → comments handler finds unresolved thread → creates another task

## Fix
Before creating a task, check if a similar task already exists:
- Exact title match (current dedup in task-cli)
- Fuzzy match: same comment_id in thread metadata
- Or: kennel marks the comment as "handled" (e.g. via the comment lock file), work.sh checks before creating

## Milestone
v1

FidoCanCode:
Working on: Deduplicate task creation: kennel + work.sh both create tasks for same comment (closes #58). Implementation in progress.

---

## Work queue

- [x] add_task: return existing task if a pending task with same title already exists
- [x] use comment id as deduplication key
- [x] Also needs work.sh update?
</details>

### DIFF
--- a/kennel/cli.py
+++ b/kennel/cli.py
@@ -66,8 +66,25 @@ class Cmd:
         except Exception as exc:  # noqa: BLE001
             log.warning("thread resolution skipped: %s", exc)
 
-    def add(self, work_dir: Path, title: str, description: str) -> None:
-        self._tasks.add_task(work_dir, title=title, description=description)
+    def add(
+        self,
+        work_dir: Path,
+        title: str,
+        description: str,
+        comment_id: int | None = None,
+        repo: str | None = None,
+        pr: int | None = None,
+    ) -> None:
+        thread: dict | None = None
+        if comment_id is not None:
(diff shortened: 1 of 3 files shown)


rhencke:
Also needs work.sh update?

FidoCanCode:
Woof! Good catch — yes, `work.sh` has its own task-creation path that can race with kennel's `add_task`. I'll sniff through that code and make sure dedup covers both sides. Thanks for the nose-boop! 🐕

PR #92, opened 2026-04-08
Re-request review after addressing review comments (closes #27)

### LINKED ISSUE #27: Re-request review after addressing review comments
Blocked by #24

After fido addresses review comments (commits + pushes fixes), it should re-request review from the owner. Currently fido just pushes and moves on — the owner has no notification that the feedback was addressed.

## Expected flow
1. Owner leaves review comments
2. Fido addresses each comment (commits, pushes, replies)
3. Fido re-requests review: \`gh pr edit --add-reviewer <owner>\`
4. Owner gets a GitHub notification

## When to re-request
- After completing all pending review comment tasks in a single loop
- NOT after every individual comment — batch them
- Only if fido actually pushed commits since the last review

## Constraints
- Don't re-request if there are still pending comment tasks
- Don't re-request on CHANGES_REQUESTED (that flow already handles it)
- Simple \`gh pr edit --add-reviewer\` call, no LLM needed

## Milestone
v1

FidoCanCode:
Teach fido to re-request review after addressing review comments, so the owner actually gets a notification that feedback was handled. After all pending comment tasks are resolved and commits are pushed, fido calls `gh pr edit --add-reviewer` to nudge the reviewer back — no more changes silently pushed without a ping.

Fixes #27.

---

## Work queue

- [x] Add should_rerequest_review helper and wire into handle_promote_merge
- [x] replace gh CLI usage with Python GitHub REST API class
</details>

### DIFF
--- a/kennel/worker.py
+++ b/kennel/worker.py
@@ -424,6 +424,28 @@ def _pick_next_task(task_list: list[dict[str, Any]]) -> dict[str, Any] | None:
     return pending[0]
 
 
+def should_rerequest_review(
+    owner_reviews: list[dict[str, Any]],
+    commits: list[dict[str, Any]],
+) -> bool:
+    """Return True if fido should re-request review from the owner.
+
+    True when the latest owner review is CHANGES_REQUESTED and either no
+    timestamps are available or the review pre-dates the latest commit
+    (meaning new work has been pushed that addresses the feedback).
+    """
+    if not owner_reviews:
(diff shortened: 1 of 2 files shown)


FidoCanCode:
Got it! You're right — I should be using our own GitHub class instead of shelling out to `gh`. I'll rework this to use the Python REST API class for re-requesting reviews. 🐕

PR #127, opened 2026-04-08
Resume sub-Claude sessions for incomplete tasks (closes #126)

### LINKED ISSUE #126: Resume sub-Claude sessions for incomplete tasks instead of restarting
When sub-Claude produces no commits after a session, the worker currently starts a fresh session on retry. This loses all the research and planning work from the previous session. Additionally, the worker doesn't retry immediately — it cycles through the full run() flow and may go idle.

### 1. Retry immediately
When a task produces no commits, retry immediately in the same `execute_task` call — don't cycle through the full worker loop.

### 2. Resume instead of restart
On retry, resume the previous session (`claude --resume <session_id>`) instead of starting fresh. Sub-Claude keeps its context and can continue where it left off.

### 3. No artificial time/context limits
Sub-Claude should be allowed to run as long as it needs. Big tasks (e.g. refactoring all REST endpoints for pagination) require reading the full codebase before making changes — that's expected, not stuck.

### 4. Detect stuck vs still working
Distinguish between:
- **Still working**: Claude made tool calls (Read, Grep, etc.) but no edits yet — resume and let it continue
- **Actually stuck**: Claude produced no output, errored, or repeated the same actions — then escalate or skip

The worker should only give up after detecting genuine stuckness, not after a single session with no commits.

FidoCanCode:
When sub-Claude runs out of context mid-research, we now resume the same session instead of starting from scratch — no more losing all that reading and planning work! Also swapped subprocess.run for streaming Popen with a 30-minute idle timeout, so stuck sessions get killed instead of hanging forever.

Fixes #126.

---

## Work queue

- [x] Resume sub-Claude sessions for incomplete tasks with idle-timeout detection
</details>

### DIFF
--- a/kennel/worker.py
+++ b/kennel/worker.py
@@ -1171,30 +1171,25 @@ class Worker:
         log.info("task done (session=%s)", session_id)
         head_after = self._git(["rev-parse", "HEAD"]).stdout.strip()
 
-        # Retry loop: resume session if no commits but Claude was still working
-        max_retries = 5
-        for attempt in range(max_retries):
-            if head_before != head_after:
-                break
-            if not session_id or not claude.session_was_active(output):
-                log.warning("task produced no commits and no output — giving up")
-                return True
-            log.info(
-                "task produced no commits but was still working — resuming "
-                "(attempt %d/%d)",
-                attempt + 1,
-                max_retries,
-            )
-            session_id, output = claude_run(fido_dir, session_id=session_id)
+        # Resume loop: let Claude cook until commits appear
+        attempt = 0
+        while head_before == head_after:
+            attempt += 1
+            if session_id:
+                log.info(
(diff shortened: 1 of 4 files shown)


kennel/worker.py
```diff
@@ -1167,12 +1167,35 @@ def execute_task(
         )
         build_prompt(fido_dir, "task", context)
         head_before = self._git(["rev-parse", "HEAD"]).stdout.strip()
-        session_id, _ = claude_run(fido_dir)
+        session_id, output = claude_run(fido_dir)
         log.info("task done (session=%s)", session_id)
         head_after = self._git(["rev-parse", "HEAD"]).stdout.strip()
+
+        # Retry loop: resume session if no commits but Claude was still working
+        max_retries = 5
```
rhencke: Let him cook as long as he wants.

PR #233, opened 2026-04-10
Fido reply includes Opus preamble — 'Here\'s the reply:' leaked into comment (closes #155)

### LINKED ISSUE #155: Fido reply includes Opus preamble — 'Here\'s the reply:' leaked into comment
Opus reply generation sometimes includes framing text like "Here's the reply:" before the actual comment body. The prompt says "Output only the comment text" but Opus adds preamble anyway.

**Observed**: PR #152 comment 4209416680 — "Here's the reply:\n\n---\n\nCreated! 🐕 ..."

**Fix**: Strip common preamble patterns (e.g. lines before "---" separator, "Here's the reply:", etc.) from generated replies before posting.

FidoCanCode:
Adds `Prompts.reply_system_prompt()` in `prompts.py` with explicit no-preamble instructions (forbidding phrases like "Here's the reply:", "Sure,", etc.) and passes it as the `system_prompt` argument to both `reply_to_comment` and `reply_to_issue_comment` in `events.py`. This prevents Opus from generating preambles at the source rather than trying to strip them after the fact.

Fixes #155.

---

## Work queue

- [x] Replace preamble stripping with improved system prompt instructions to prevent Opus preamble leakage 
- [x] [deduplicate repeated code patterns in events.py](https://github.com/rhencke/kennel/pull/233#discussion_r3061855600) 
</details>

### DIFF
--- a/kennel/events.py
+++ b/kennel/events.py
@@ -328,7 +328,12 @@ def reply_to_comment(
         info["pr"],
         info["comment_id"],
     )
-    body = _print_prompt(prompts.persona_wrap(instr), "claude-opus-4-6", timeout=30)
+    body = _print_prompt(
+        prompts.persona_wrap(instr),
+        "claude-opus-4-6",
+        system_prompt=prompts.reply_system_prompt(),
+        timeout=30,
+    )
 
     if not body:
         body = (
@@ -585,7 +590,12 @@ def reply_to_issue_comment(
     )
 
     log.info("generating %s reply for issue comment on PR #%s", category, number)
-    body = _print_prompt(prompts.persona_wrap(instr), "claude-opus-4-6", timeout=30)
+    body = _print_prompt(
+        prompts.persona_wrap(instr),
+        "claude-opus-4-6",
+        system_prompt=prompts.reply_system_prompt(),
+        timeout=30,
+    )
     if not body:
         body = "On it!" if category in ("ACT", "DO") else "Noted."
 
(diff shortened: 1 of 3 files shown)


rhencke:
Needs a better system prompt and not fragile string heuristics

FidoCanCode:
Woof, good call! You're right — stripping preambles after the fact is playing whack-a-mole. I'll tighten up the system prompt to tell Opus not to generate preambles in the first place, rather than trying to clean them up with string heuristics. Much more robust that way. 🐕

FidoCanCode:
*tail wag* Done! Added `Prompts.reply_system_prompt()` with explicit no-preamble instructions and wired it into both reply functions as a `system_prompt` argument. No more fragile string heuristics — much cleaner scent trail now. Arf!

kennel/events.py
rhencke, line 1: Look for non-DRY/copypaste in this file and fix it all in spec for this pr

rhencke (changes requested):

kennel/events.py
FidoCanCode: On it! 🐕 I'll sniff through events.py for repeated patterns and deduplicate them as part of this PR. Good eye spotting that — DRY code is happy code. *tail wag*

FidoCanCode: Arf! Extracted `_load_persona()` and `_open_defer_issue()` helpers, hoisted `repo_full` and `gh` to single calls in `reply_to_issue_comment`, and moved deferred imports to module level. 35 fewer lines, 100% coverage still holding. 🐾

rhencke (approved):

PR #236, opened 2026-04-10
Dynamically reorder and rewrite tasks based on dependency analysis (closes #165)

### LINKED ISSUE #165: Dynamically reorder and rewrite tasks based on dependency analysis
When new tasks arrive from PR comments or fido expands the task list mid-work, the task list should be dynamically reordered based on dependency analysis using the planning/development context.

**Current problem**: Tasks are appended in insertion order. When fido rescopes (e.g. 3 tasks → 5 tasks), the new tasks go to the end without considering dependencies. Task titles get rewritten without the planning session's understanding of the architecture.

**Requirements**:
- CI failure tasks always go first (unchanged)
- When rescoping, use the existing planning/dev session context (the shared Opus session from setup) to understand dependencies
- Remaining tasks (both PR comment tasks and spec tasks) should be analyzed for dependencies and reordered accordingly
- Task descriptions should be rewritten if a new comment changes the requirements of an existing task
- The rescoping should understand what's already been committed and what depends on what and use Opus for the reorder/removals/adds/changes.

**Example**: If task A adds a new method and task B writes tests for it, B depends on A regardless of insertion order. The planning context already knows this from reading the codebase during setup.

FidoCanCode:
Adds dynamic task reordering via Opus dependency analysis when new PR comment tasks arrive mid-work. A `rescope_prompt()` builder in prompts.py feeds the current task list and commit history to Opus, `reorder_tasks()` in tasks.py applies the reordered result while preserving IDs/statuses/thread metadata and keeping CI tasks first, and `create_task()` in events.py triggers rescoping only for thread-type tasks (skipping initial setup). Documentation in CLAUDE.md is updated to reflect the new behavior.

Fixes #165.

---

## Work queue

- [ ] Update CLAUDE.md to document dynamic task reordering **→ next** 

- [x] Cancel in-progress task and restart on new next task when rescoping affects current task 
- [x] Elevate PR comment tasks to spec-level priority in task ordering 
- [x] Mark tasks done if no longer applicable during rescoping 
</details>

### DIFF
--- a/kennel/prompts.py
+++ b/kennel/prompts.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import json
 from typing import Any
 
 # ── Triage ────────────────────────────────────────────────────────────────────
@@ -199,6 +200,72 @@ def issue_reply_instruction(
             return f"Write a short polite decline.\n\n{context_str}"
         case _:
             return f"Write a short GitHub PR reply.\n\n{context_str}"
+
+
+# ── Rescoping ────────────────────────────────────────────────────────────────
+
+
+def rescope_prompt(
+    task_list: list[dict[str, Any]],
+    commit_summary: str,
+) -> str:
+    """Build an Opus prompt for dependency-aware task reordering.
+
+    Presents the full task list and a summary of commits already made, then
+    asks Opus to return a JSON array of the reordered pending tasks.
+
+    Rules enforced in the prompt:
+    - CI tasks (type "ci") must remain first.
+    - Completed tasks are excluded from the output.
+    - Task IDs must be preserved exactly.
+    - Tasks already covered by a commit should be removed.
+    - Thread-task requirements that conflict with a spec task should cause
+      the spec task title/description to be updated.
+
+    The caller is responsible for parsing the returned JSON and applying it.
+    """
+    pending = [t for t in task_list if t.get("status") != "completed"]
+    completed = [t for t in task_list if t.get("status") == "completed"]
+
+    def _fmt(t: dict[str, Any]) -> dict[str, Any]:
+        return {
+            "id": t.get("id", ""),
+            "type": t.get("type", "spec"),
+            "status": t.get("status", "pending"),
+            "title": t.get("title", ""),
+            "description": t.get("description", ""),
+        }
+
+    pending_json = json.dumps([_fmt(t) for t in pending], indent=2)
+    completed_titles = [t.get("title", "") for t in completed]
+    completed_block = (
+        "\n".join(f"- {title}" for title in completed_titles)
+        if completed_titles
+        else "(none)"
+    )
+
+    return (
+        "You are reviewing the pending work queue for a pull request in progress.\n\n"
+        "Already completed tasks:\n"
+        f"{completed_block}\n\n"
+        "Recent commits (already implemented):\n"
+        f"{commit_summary or '(none)'}\n\n"
+        "Pending tasks (current order):\n"
+        f"{pending_json}\n\n"
+        "Reorder these tasks for the optimal implementation sequence based on "
+        "dependency analysis. Apply these rules:\n"
+        '1. Tasks with type "ci" must come first — do not move them.\n'
+        "2. Reorder remaining tasks so each task builds on what comes before it.\n"
+        "3. If a task is already covered by a recent commit, remove it.\n"
+        "4. If a thread task changes the requirements of an existing spec task, "
+        "rewrite that spec task's title and/or description to reflect the updated "
+        "requirements.\n"
+        "5. Preserve every task ID exactly — never change or drop IDs.\n"
+        "6. Include only pending and in_progress tasks in the output — omit completed.\n\n"
+        'Reply with ONLY a JSON object in the form {"tasks": [...]}.\n'
+        'Each element: {"id": "...", "title": "...", "description": "..."}.\n'
+        "No other text before or after the JSON."
+    )
 
 
 # ── Prompts DI class ──────────────────────────────────────────────────────────
--- a/kennel/prompts.py
+++ b/kennel/prompts.py
@@ -221,7 +221,8 @@ def rescope_prompt(
     - CI tasks (type "ci") must remain first.
     - Completed tasks are excluded from the output.
     - Task IDs must be preserved exactly.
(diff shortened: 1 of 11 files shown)


rhencke:
New comments may invalidate or modify old comments.  If a task from a comment *off thread* was changed or deleted, reply to the comment to let them know.

FidoCanCode:
Woof, good call! 🐕

When rescoping modifies or removes a task that came from an off-thread comment, I'll post a reply on the original comment letting them know what changed. That way nobody's left wondering why their request looks different (or disappeared) after a dependency reshuffle.

rhencke:
A comment may generate zero to many tasks.

FidoCanCode:
Woof! Yep, you're right — a single comment can spawn a whole litter of tasks (or none at all). I'll update the task creation path to return a list instead of assuming one-in-one-out. Good sniff! 🐾

rhencke:
Rescoping a task list may add, modify, delete tasks, including tasks from PR comments.  Any modify/delete a task does to PR comments should inform the comment author.

FidoCanCode:
Good sniff — if rescoping touches a task that came from a PR comment, the author deserves to know what happened to it. 🐕

My plan: when rescope modifies or deletes a comment-originated task, I'll post a reply on the original comment thread tagging the author with what changed (modified → show the diff, deleted → say why). That way nobody's feedback disappears into the void without a heads-up.

rhencke:
If the in-progress task is affected by the rescope, it is cancelled (worker kill) cleanup and now do the *new* next task.

FidoCanCode:
Woof! Good catch — yeah, if the task that's currently running gets reshuffled or dropped by a rescope, we need to pull the leash on it rather than let it finish stale work.

My plan: when a rescope affects the in-progress task, kill the worker, clean up, and immediately pick up whatever the new next task is. No letting the old work linger. 🐕

rhencke:
Big change, in scope for this PR.  PR tasks are now scoped at the same precedence as tasks from the plan (ci still always comes first).

Opus enriched with our session and all relevant details reorders same time it's handling rescoping/add/delete/etc.

FidoCanCode:
*ears perk up* Got it — PR comment tasks get promoted to spec-level priority so they're treated as equals in the ordering, not second-class citizens. CI stays top dog (as it should), but everything else competes on the same footing.

I'll update `_pick_next_task` to collapse `thread` and `spec` into the same priority tier. Woof! 🐕

rhencke:
Opus enriched with our session and all relevant details reorders same time it's handling rescoping/add/delete/etc.

FidoCanCode:
Good sniff on this one! 🐕

You're right — Opus already has the full picture when it's rescoping, so bolting on a separate dependency reordering step is overengineering it. The simpler fix is to make sure PR comment tasks (`thread` type) get elevated to spec-level priority in `_pick_next_task`, so Opus's natural reordering during triage does the heavy lifting.

I'll adjust the task priority logic so thread tasks aren't stuck behind spec tasks when they're blocking progress. Less machinery, same result. *tail wag*

rhencke:
Pr thread comments still hyperlink like today.

FidoCanCode:
Yep! Thread comments still link back to the original PR conversation the same way they do now — no changes there. 🐾

kennel/prompts.py
```diff
@@ -201,6 +205,72 @@ def issue_reply_instruction(
...
+        "Pending tasks (current order):\n"
+        f"{pending_json}\n\n"
+        "Reorder these tasks for the optimal implementation sequence based on "
+        "dependency analysis. Apply these rules:\n"
+        '1. Tasks with type "ci" must come first — do not move them.\n'
+        "2. Reorder remaining tasks so each task builds on what comes before it.\n"
+        "3. If a task is already covered by a recent commit, remove it.\n"
+        "4. If a thread task changes the requirements of an existing spec task, "
+        "rewrite that spec task's title and/or description to reflect the updated "
+        "requirements.\n"
```
rhencke, line 266: Or mark done if no longer applicable, and, if a PR task, let the commenter know why fell out of scope

rhencke (changes requested):

kennel/prompts.py
FidoCanCode: Woof, quick clarifying question before I chase this down! When you say "mark done if no longer applicable" — do you want me to:

1. Add this as new behavior to the rescoping flow (so Opus can mark stale tasks done and auto-notify the commenter for thread tasks), or
2. Apply it right now to specific existing tasks you have in mind?

Just want to make sure I'm sniffing the right tree before I dig in!

FidoCanCode:
Woof! Just a heads up — your original task got updated based on [this comment](https://github.com/rhencke/kennel/pull/236#issuecomment-4220153098). Tail wagging, back to sniffing out the work! 🐾

FidoCanCode:
Hey! Just a heads up — the task from [this comment](https://github.com/rhencke/kennel/pull/236#issuecomment-4220168451) about elevating PR comment tasks to spec-level priority has been dropped from my queue. A later comment changed the requirements, so this work isn't needed anymore. *tail wags* Let me know if I misread that!

FidoCanCode:
Woof! Just a heads up — the task from your comment ([r3062024128](https://github.com/rhencke/kennel/pull/236#discussion_r3062024128)) got rescoped to "Extend rescoping to mark tasks done when no longer applicable." Same spirit, sharper scope. *tail wag* 🐕

FidoCanCode:
Done! Woof — when `reorder_tasks` notices the in-progress task got dropped or modified by Opus, it now calls `_on_inprogress_affected()`, which yanks the leash via `registry.abort_task()`. *tail wagging* Dropped tasks vanish entirely, and modified ones get reset to pending so the worker fetches them fresh on the next go-round. Also chewed up the old safety-net that used to reinstate dropped in-progress tasks — didn't need it anymore! Arf!

kennel/prompts.py
FidoCanCode: Done! Tasks that Opus determines are covered by recent commits now get **marked completed** instead of vanishing silently — they'll show up in the collapsible Completed section of the PR body. And if the task came from a PR comment (thread type), I'll ping the commenter to let them know it was marked done and why. Woof! 🐾

rhencke (approved):

PR #358, opened 2026-04-13
RCA: worker dies after successful launch — liveness check insufficient (closes #71)

### LINKED ISSUE #71: RCA: worker dies after successful launch — liveness check insufficient
## Problem
\`launch_worker\` liveness check (from #67) only verifies the worker is alive 1-2s after launch. But the worker dies later (after sub-Claude finishes a task, or during loop iteration) and nobody notices until cron fires.

## Timeline (02:36)
1. Kennel self-restarted with #67 fix live
2. Worker launched for kennel repo
3. Worker ran, completed work, exited normally (no more tasks)
4. New issue #69 assigned — webhook fired, kennel called \`launch_worker\`
5. Worker launched, liveness check passed (process alive at +1s)
6. Worker died shortly after (unknown reason)
7. No retry — kennel thinks worker is running

## Root cause
The liveness check is a point-in-time snapshot. It can't detect:
- Worker that exits after completing one cycle
- Worker that crashes mid-task
- Worker that exits 0 (no work) and never comes back

## What's needed
Not just "is it alive at launch" but ongoing health monitoring:
- **Option A**: Worker writes a heartbeat file; kennel checks freshness
- **Option B**: Worker holds a pipe/socket open; kennel detects close
- **Option C**: Supervisor thread (v1) — continuously monitors all workers

## Interim fix
After \`launch_worker\`, wait for process to finish (not just launch), then check exit code. If exit 0 with no work done, that's fine. If exit non-zero, retry.

Or: \`launch_worker\` returns the Popen object; kennel polls it periodically.

## Milestone
v1

FidoCanCode:
RCA: worker dies after successful launch — liveness check insufficient (closes #71)

Fixes #71.

---

## Work queue

- [x] [Consolidate per-repo watchdog threads into a single shared watchdog thread](https://github.com/FidoCanCode/home/pull/358#discussion_r3073472445) 
- [x] Integrate Watchdog into server runtime as periodic daemon thread 
</details>

### DIFF
--- a/kennel/server.py
+++ b/kennel/server.py
@@ -41,26 +41,6 @@ _replied_comments: set[int] = set()
 # exceeds _PULL_BUDGET_SECONDS, even if a retry window remains.
 _PULL_BACKOFF_DELAYS: tuple[int, ...] = (10, 30, 60)
 _PULL_BUDGET_SECONDS: float = 600.0
-_WATCHDOG_INTERVAL: float = 30.0
-
-
-def _start_watchdog_thread(
-    registry: WorkerRegistry,
-    repos: dict[str, RepoConfig],
-    *,
-    _interval: float = _WATCHDOG_INTERVAL,
-) -> threading.Thread:
-    """Start a daemon thread that periodically runs the watchdog."""
-    watchdog = Watchdog(registry, repos)
-
-    def _loop() -> None:
-        while True:
-            time.sleep(_interval)
-            watchdog.run()
-
-    t = threading.Thread(target=_loop, daemon=True, name="watchdog")
-    t.start()
-    return t
 
 
 def _runner_dir() -> Path:
@@ -454,7 +434,7 @@ def run(
     _signal=signal.signal,
     _kill_active_children=kill_active_children,
     _startup_pull=_startup_pull,
-    _start_watchdog=_start_watchdog_thread,
+    _Watchdog=Watchdog,
 ) -> None:
     config = _from_args()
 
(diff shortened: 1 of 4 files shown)


kennel/server.py
```diff
@@ -40,6 +41,26 @@
 # exceeds _PULL_BUDGET_SECONDS, even if a retry window remains.
 _PULL_BACKOFF_DELAYS: tuple[int, ...] = (10, 30, 60)
 _PULL_BUDGET_SECONDS: float = 600.0
+_WATCHDOG_INTERVAL: float = 30.0
+
+
+def _start_watchdog_thread(
```
rhencke: Do we need a per-repo watchdog thread?  Or just one "Repo Watchdog" thread that knows what to do for each repo?

FidoCanCode: Good nose on this one! 🐕 You're right — spinning up a separate thread per repo is wasteful when one thread can just loop through all of them.

I'll consolidate into a single watchdog thread that iterates over every repo's worker on each tick. Same checks, same interval, fewer threads.

FidoCanCode: Woof, good question! 🐾 I'd lean toward **one shared Repo Watchdog thread** — it fits the existing model better. Kennel is already one process that knows about all repos, and a single watchdog thread can loop over all of them without any extra coordination overhead. Per-repo threads would mean spinning up N threads just to do the same check N times independently, and if a new repo gets added there's nothing to wire up. One thread, one loop, all repos — simple and consistent with how the rest of kennel thinks about things.

FidoCanCode: Done — moved `start_thread()` into `Watchdog` itself. One thread, one object, one place. `server.run()` now just does `_Watchdog(registry, repos).start_thread()`.

rhencke (approved):

PR #381, opened 2026-04-13
Make worker-thread exceptions fatal and observable (closes #372)

### LINKED ISSUE #71: RCA: worker dies after successful launch — liveness check insufficient

### LINKED ISSUE #372: Make worker-thread exceptions fatal and observable
## Problem

`WorkerThread.run()` currently catches unexpected exceptions, logs them, sleeps, and continues. That keeps the process alive, but it turns real worker crashes into an internal retry loop instead of a first-class runtime event.

This makes it hard to tell the difference between:

- a healthy worker doing normal retries
- a worker repeatedly crashing with no forward progress
- a worker that should be considered down until a bug is fixed

## Scope

- `kennel/worker.py`
- `kennel/watchdog.py`
- `kennel/registry.py`
- `kennel/status.py`

## Acceptance criteria

- unexpected worker exceptions are logged once with stack trace
- worker death and restart counts are recorded
- recent worker-death information is exposed in `kennel status`
- repeated crash loops are visible as a failure condition
- unexpected exceptions are no longer silently retried inside the same thread loop

FidoCanCode:
Adds crash tracking to the worker system so that when a fido thread dies from an unexpected exception, the details are recorded and surfaced in status output. Watchdog captures crash details before restarting dead threads, WorkerRegistry exposes crash history through RepoStatus and format_status output, and defensive `.get()` calls are replaced with direct key access where fields are guaranteed present.

Fixes #71.

Fixes #372.

---

## Work queue

- [x] [Replace defensive `.get()` calls with direct key access where the field is guaranteed present](https://github.com/FidoCanCode/home/pull/381#issuecomment-4239366640) 
- [x] Expose crash info in RepoStatus and format_status output 
- [x] Add crash-tracking fields to WorkerRegistry 
- [x] Make WorkerThread.run() exit on unexpected exceptions instead of retrying 
- [x] [Stop swallowing exceptions; propagate to top-level handler](https://github.com/FidoCanCode/home/pull/381#issuecomment-4239353469) 
</details>

### DIFF
--- a/kennel/registry.py
+++ b/kennel/registry.py
@@ -7,6 +7,7 @@ import threading
 from collections.abc import Callable, Generator
 from contextlib import contextmanager
 from dataclasses import dataclass
+from datetime import datetime
 
 from kennel.config import RepoConfig
 from kennel.github import GitHub
@@ -22,6 +23,15 @@ class WorkerActivity:
     repo_name: str
     what: str
     busy: bool
+
+
+@dataclass
+class WorkerCrash:
+    """Running record of unexpected worker deaths for one repo."""
+
+    death_count: int
+    last_error: str
+    last_crash_time: datetime
 
 
 class WorkerRegistry:
@@ -44,6 +54,8 @@ class WorkerRegistry:
         self._activities: dict[str, WorkerActivity] = {}
         self._activity_lock = threading.Lock()
         self._status_lock = threading.Lock()
+        self._crashes: dict[str, WorkerCrash] = {}
+        self._crash_lock = threading.Lock()
 
     def start(self, repo_cfg: RepoConfig) -> None:
         """Create and start a WorkerThread for *repo_cfg*."""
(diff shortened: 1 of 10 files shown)


rhencke:
This is more than that.  This is also making sure those exceptions make it to a toplevel handler that logs and dies.  We have to stop eating exceptions everywhere and returning default and "safe" values hiding real errors.

FidoCanCode:
Arf, yes — you're absolutely right, and I should name it properly. This isn't just "make exceptions fatal." It's a cleanup of all the places where we silently swallow errors and return safe-looking defaults, hiding real problems.

The plan: stop catching-and-returning-defaults everywhere, let exceptions propagate naturally, and make sure the top-level handler actually logs them and dies loudly. No more quiet failures pretending everything is fine. 🐕

FidoCanCode:
Woof! This one's done — the exception propagation work was already covered by recent commits, so I've marked the task complete. 🐾

(Ref: https://github.com/FidoCanCode/home/pull/381#issuecomment-4239353469)

rhencke:
Additionally, abuse of `x.get("field", {})` when the value is documented as always returned.  Just `x["field"]` it if it is supposed to be there.

FidoCanCode:
Good nose! 🐕 You're right — if the API guarantees the field is there, `.get("field", {})` just hides bugs instead of surfacing them. I'll swap those out for direct `x["field"]` access so we fail loud and clear if our assumptions are ever wrong.

rhencke (approved):

rhencke (approved):

PR #406, opened 2026-04-13
Stop turning Claude failure into synthetic tasks (closes #376)

### LINKED ISSUE #376: Stop turning Claude failure into synthetic tasks
## Problem

Claude subprocess failure currently degrades into empty strings and fallback behavior. In several paths that empty result is treated as valid enough to continue, including creation of synthetic `ACT` / `DO` tasks.

That makes model failure look like real triage output instead of a runtime problem.

## Scope

- `kennel/claude.py`
- `kennel/events.py`

## Acceptance criteria

- subprocess/model failure is distinct from valid empty content
- triage failure does not auto-create fallback tasks
- reply and triage flows surface failure instead of inventing work
- core Claude subprocess calls use explicit exit-code handling (`check=True` or equivalent)

FidoCanCode:
The current body is just `Fixes #376.` — so there's no existing descriptive summary beyond that. Based on the updated task list, here's the replacement description:

Fixes #376.

When Claude subprocess calls fail, kennel currently swallows the error and falls back to a synthetic ACT/DO response, which creates nonsensical tasks. This PR makes Claude failures raise exceptions instead of returning empty strings and removes the synthetic fallback from triage, letting failures surface through the normal reply flow.

---

Want me to update the PR body directly?

---

## Work queue

- [x] Raise on Claude subprocess failure instead of returning empty string 
- [x] Remove synthetic ACT/DO fallback from triage and surface failure in reply flows 
- [x] PR comment: remove try/except ClaudeError in gh_status.generate_persona_status 
</details>

### DIFF
--- a/kennel/gh_status.py
+++ b/kennel/gh_status.py
@@ -31,12 +31,15 @@ def generate_persona_status(
     _print_prompt: Callable[..., str] = claude.print_prompt,
 ) -> str:
     system = f"{persona}\n\n{_STATUS_SYSTEM}" if persona else _STATUS_SYSTEM
-    result = _print_prompt(
-        prompt=f"Rewrite this status in Fido's voice: {message}",
-        model="claude-opus-4-6",
-        system_prompt=system,
-        timeout=15,
-    )
+    try:
+        result = _print_prompt(
+            prompt=f"Rewrite this status in Fido's voice: {message}",
+            model="claude-opus-4-6",
+            system_prompt=system,
+            timeout=15,
+        )
+    except claude.ClaudeError:
+        return message[:80]
     return result if result else message[:80]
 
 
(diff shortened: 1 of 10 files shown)


kennel/gh_status.py
```diff
@@ -31,12 +31,15 @@ def generate_persona_status(
...
-        timeout=15,
-    )
+    try:
+        result = _print_prompt(
+            prompt=f"Rewrite this status in Fido's voice: {message}",
+            model="claude-opus-4-6",
+            system_prompt=system,
+            timeout=15,
+        )
+    except claude.ClaudeError:
```
rhencke: This is exactly what we are trying to get rid of.

rhencke (changes requested):

kennel/gh_status.py
FidoCanCode: Arf, you're absolutely right — caught red-pawed! 🐾 Adding a try/except around  is exactly the pattern this PR is supposed to eliminate. I've queued this up to rip it out.

FidoCanCode: Done — removed the try/except from both generate_persona_status and generate_persona_emoji. ClaudeError now propagates to set_gh_status, which returns early on failure.

PR #434, opened 2026-04-14
Make runner failures fail fast and surface worker death explicitly (closes #197)

### LINKED ISSUE #197: Make runner failures fail fast and surface worker death explicitly
## Problem

Core runner paths still downgrade real failures into retries, empty values, or log-and-continue behavior. That creates five operational risks:

- webhook actions can be acknowledged to GitHub but not actually completed
- worker/task orchestration can continue after state corruption or upstream failures
- worker death/restart behavior is not exposed as a first-class runtime signal
- startup can enter a half-valid state and only fail later inside runtime paths
- a worker can hang on external I/O without dying, so the watchdog never treats it as failed

The audit for this issue found that the real work is not a single blanket "clean up error handling" pass. It is a set of concrete runtime fixes across worker lifecycle, webhook dispatch, Claude subprocess handling, GraphQL contract validation, queue integrity, sync paths, reply flows, subprocess exit-code handling, startup preflight, stale-worker detection, and legacy shell suppression.

## Scope

- core Python runner paths only (`kennel/*.py`)
- worker-thread exception handling
- webhook action processing for comments, reviews, and task creation
- Claude/GitHub contract validation in core paths
- task / PR-body sync and rewrite paths
- worker death / restart visibility in `kennel status`
- subprocess exit-code handling in core paths (`check=True` or explicit `returncode` handling)
- startup preflight validation for repo mapping, tooling, and auth
- bounded external I/O and stuck-worker detection
- legacy shell paths only if they are still supported runtime paths

## Key findings from the audit

- `WorkerThread.run()` still catches unexpected exceptions and continues instead of surfacing worker death as a runtime event.
- `_process_action()` can fail after GitHub has already received HTTP 200.
- Claude subprocess failure can degrade into empty strings and synthetic fallback task creation.
- GraphQL response-shape violations can collapse into nested `.get()` defaults instead of hard failure.
- corrupt `tasks.json` and empty setup can fail open instead of failing closed.
- task/PR sync and PR-description rewrite paths can log-and-return, leaving silent divergence.
- reply/defer flows can partially fail while downstream task creation still proceeds.
- some subprocess helpers still ignore nonzero exits by converting them into empty-string fallbacks.
- startup only validates a narrow config surface; repo mapping, auth, and tooling consistency can still fail later at runtime.
- outbound HTTP calls can hang without bounded timeouts, and the watchdog only notices dead workers, not stuck ones.
- legacy shell paths still contain broad suppression (`|| true`, `2>/dev/null`, fallback behavior).

## Child issues

- #372 Make worker-thread exceptions fatal and observable
- #373 Stop acknowledging webhook work that later fails
- #374 Stop turning Claude failure into synthetic tasks
- #375 Validate GitHub GraphQL responses as required contracts
- #376 Fail closed on authoritative queue corruption and empty setup
- #377 Surface PR-body sync and rewrite failures
- #378 Make reply flows fail closed around human-visible side effects
- #379 Audit core subprocess calls for explicit exit-code handling
- #380 Retire or harden legacy shell suppression paths
- #382 Bound external I/O and detect stuck workers
- #383 Add startup preflight for repo mapping, tooling, and auth

## Acceptance criteria

- no broad catch-log-continue behavior in core runner / worker lifecycle paths
- no conversion of Claude/GitHub failures into synthetic "safe" task creation
- no `.get()` / default-value fallback where the response shape is required by contract
- no subprocess call in core paths that ignores a failing exit code
- startup does not proceed with invalid repo mapping, missing required auth, or missing required tooling
- core outbound I/O is bounded and stuck-worker state is surfaced explicitly
- unexpected worker exceptions become explicit runtime events, with restart/death accounting
- `kennel status` shows worker restart/death information, not just current activity
- authoritative state corruption (`tasks.json`, required GraphQL shapes) fails closed and is surfaced
- background sync/rewrite failures are surfaced and actionable, not just logged

## Notes

- Not every broad `except` in the repository is a bug. Some observability and dev-convenience paths are intentionally best-effort.
- This parent issue now tracks the concrete implementation issues above rather than serving as an umbrella manifesto.

### LINKED ISSUE #383: Add startup preflight for repo mapping, tooling, and auth
## Problem

Startup currently validates only a narrow slice of configuration: the webhook secret exists and each configured work directory exists. It does not validate the higher-level runtime contract before the server starts:

- each configured `work_dir` is actually the expected git repository
- repo identity derived from the worktree matches the configured `owner/repo`
- required external tooling and auth are available for the configured runtime path
- startup does not enter a half-valid state that fails later inside worker threads or webhook handling

That leaves room for a runner that starts successfully but is already misconfigured in a way that will only surface later, after webhooks arrive or workers begin operating.

## Scope

- `kennel/config.py`
- `kennel/server.py`
- `kennel/worker.py`
- `kennel/github.py`
- `kennel/claude.py`

## Acceptance criteria

- startup validates that each configured `work_dir` is a git repo for the expected configured `owner/repo`
- startup validates the required auth/tooling for the configured runtime path before serving webhooks
- repo identity is not allowed to diverge between configured repo name and discovered worktree repo
- invalid startup state fails before workers and webhook handling begin
- tests cover repo-mapping mismatch and missing dependency/auth cases

FidoCanCode:
Adds startup preflight checks so kennel fails closed before workers and webhooks start.

- Verifies each `work_dir` is a git repo for its configured `owner/repo`
- Confirms required CLI tools (`git`, `gh`, `claude`) are on PATH
- Checks that `gh` auth works by fetching the bot user
- Checks that the skill-files directory (`sub/`) exists

Fixes #383.

## Work queue

- [x] [replace regex URL parsing with urlparse in _parse_repo_from_url](https://github.com/FidoCanCode/home/pull/434#discussion_r3076902821) 
- [x] Add preflight: verify required CLI tools (git, gh, claude) are on PATH 
- [x] Add preflight: verify gh auth works by fetching the bot user 
- [x] [Expand preflight to cover all remaining checks from issue #383 (beyond repo/CLI/gh-auth)](https://github.com/FidoCanCode/home/pull/434#issuecomment-4240952720) 
- [x] Wire preflight into run() to fail closed before workers and webhooks start 
- [x] PR comment: replace SystemExit with PreflightError in preflight_repo_identity 
- [x] [replace SystemExit with custom PreflightError exception in preflight functions](https://github.com/FidoCanCode/home/pull/434#discussion_r3076934429) 
</details>

### DIFF
--- a/kennel/server.py
+++ b/kennel/server.py
@@ -6,7 +6,6 @@ import hmac
 import json
 import logging
 import os
-import re
 import shutil
 import signal
 import subprocess
@@ -16,6 +15,7 @@ import time
 from collections.abc import Callable
 from http.server import BaseHTTPRequestHandler, HTTPServer
 from pathlib import Path
+from urllib.parse import urlparse
 
 from kennel.claude import kill_active_children
 from kennel.config import Config, RepoConfig, RepoMembership
@@ -51,8 +51,18 @@ def _runner_dir() -> Path:
 
 def _parse_repo_from_url(url: str) -> str | None:
     """Extract 'owner/repo' from an SSH or HTTPS git remote URL, or return None."""
-    m = re.search(r"[:/]([^:/]+/[^/]+?)(?:\.git)?$", url)
-    return m.group(1) if m else None
+    parsed = urlparse(url)
+    if parsed.scheme:
+        # Standard URL (https, ssh, git, etc.): path is /owner/repo[.git]
+        path = parsed.path
+    elif ":" in url:
+        # SCP-style SSH: git@github.com:owner/repo[.git]
+        _, path = url.split(":", 1)
+    else:
+        return None
+    path = path.lstrip("/").removesuffix(".git")
+    parts = path.split("/")
+    return path if len(parts) == 2 and all(parts) else None
 
 
 def _get_self_repo(
--- a/kennel/server.py
+++ b/kennel/server.py
@@ -48,6 +48,12 @@ def _runner_dir() -> Path:
     return Path(__file__).resolve().parents[1]
 
 
+def _parse_repo_from_url(url: str) -> str | None:
+    """Extract 'owner/repo' from an SSH or HTTPS git remote URL, or return None."""
+    m = re.search(r"[:/]([^:/]+/[^/]+?)(?:\.git)?$", url)
+    return m.group(1) if m else None
+
+
 def _get_self_repo(
     runner_dir: Path,
     *,
@@ -70,11 +76,51 @@ def _get_self_repo(
         log.error("self-restart: failed to read origin remote: %s", e)
         return None
     url = result.stdout.strip()
-    m = re.search(r"[:/]([^:/]+/[^/]+?)(?:\.git)?$", url)
-    if not m:
+    parsed = _parse_repo_from_url(url)
+    if not parsed:
         log.error("self-restart: could not parse owner/repo from remote url: %r", url)
         return None
-    return m.group(1)
+    return parsed
+
+
+def preflight_repo_identity(
+    repos: dict[str, RepoConfig],
+    *,
+    _run: Callable[..., subprocess.CompletedProcess[str]] = subprocess.run,
+) -> None:
+    """Verify each configured work_dir is a git repo whose origin matches its name.
+
+    Raises :exc:`SystemExit` if any repo's origin remote can't be read, can't
+    be parsed, or doesn't match the configured ``owner/repo`` name.  This runs
+    once at startup so misconfigured repo mappings fail immediately rather than
+    surfacing as silent divergence deep inside webhook or worker paths.
+    """
+    for name, repo_cfg in repos.items():
+        try:
+            result = _run(
+                ["git", "remote", "get-url", "origin"],
+                cwd=str(repo_cfg.work_dir),
+                capture_output=True,
+                text=True,
+                check=True,
+            )
+        except subprocess.CalledProcessError as e:
+            raise SystemExit(
+                f"preflight: {name}: git remote get-url failed: {e}"
+            ) from e
+        except FileNotFoundError as e:
+            raise SystemExit(f"preflight: {name}: git not found: {e}") from e
+        url = result.stdout.strip()
+        actual = _parse_repo_from_url(url)
+        if actual is None:
+            raise SystemExit(
+                f"preflight: {name}: could not parse owner/repo from origin remote: {url!r}"
+            )
+        if actual != name:
+            raise SystemExit(
+                f"preflight: {name}: origin remote is {actual!r} — expected {name!r}"
+            )
+        log.info("preflight: %s: work_dir identity confirmed", name)
 
 
 def _get_head(
@@ -488,6 +534,7 @@ def run(
     _kill_active_children=kill_active_children,
     _startup_pull=_startup_pull,
     _Watchdog=Watchdog,
+    _preflight_repo_identity=preflight_repo_identity,
 ) -> None:
     config = _from_args()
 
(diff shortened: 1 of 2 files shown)


rhencke:
Since you grabbed the wrong parent issue you need to include all of https://github.com/FidoCanCode/home/issues/383

FidoCanCode:
Woof, you're right — I snagged the wrong parent. Sorry about that! *ears drooping a little* I'll expand the preflight scope in this PR to cover all of #383, not just the slice I had pinned to #197. Sniffing through the full issue now and widening the checks accordingly.

FidoCanCode:
Woof! Just a heads up — your [comment](https://github.com/FidoCanCode/home/pull/434#issuecomment-4240952720) got rescoped into an updated task: *Expand preflight to cover all remaining checks from issue #383 (beyond repo/CLI/gh-auth)*. On the scent! 🐾

kennel/server.py
```diff
@@ -48,6 +49,12 @@ def _runner_dir() -> Path:
     return Path(__file__).resolve().parents[1]
 
 
+def _parse_repo_from_url(url: str) -> str | None:
+    """Extract 'owner/repo' from an SSH or HTTPS git remote URL, or return None."""
+    m = re.search(r"[:/]([^:/]+/[^/]+?)(?:\.git)?$", url)
```
rhencke: I mean `urlparse`

FidoCanCode: Woof, good call — `urlparse` is the right tool here. I'll swap the regex for `urlparse`, strip a leading slash and any `.git` suffix off the path, and handle the SSH `git@host:owner/repo` shape (which `urlparse` doesn't split cleanly) as a small special case. Cleaner and less clever. *tail wag*

FidoCanCode: Arf! Good catch —  is much cleaner than fighting a regex here. Queued it up! 🐾

kennel/server.py
```diff
@@ -70,11 +87,107 @@ def _get_self_repo(
...
+        try:
+            result = _run(
+                ["git", "remote", "get-url", "origin"],
+                cwd=str(repo_cfg.work_dir),
+                capture_output=True,
+                text=True,
+                check=True,
+            )
+        except subprocess.CalledProcessError as e:
+            raise SystemExit(
```
rhencke: SystemExit..?

FidoCanCode: Woof, good catch — want to make sure I sniff out the right concern before I chase my tail here. Are you saying `SystemExit` is the wrong mechanism entirely (e.g. you'd prefer a custom exception bubbled up to the caller, or a bool return), or is it more about *where* it's raised (preflight shouldn't be the one killing the process)? Happy to rework either way — just want to dig at the right bone. 🐾

FidoCanCode: Yep, intentional! `raise SystemExit(msg)` and `sys.exit(msg)` are literally the same thing — `sys.exit` just does `raise SystemExit(msg)` under the hood. I went with the explicit raise form here because the preflight functions do not import `sys`, so it avoids pulling in an extra import just for the alias.

That said, `sys.exit()` is the more conventional-looking spelling in application code. Happy to swap to `sys.exit()` if you prefer it — just say the word! *tail wag*

rhencke: How about raise PreflightError

FidoCanCode: Good call! *tail wag* Raising a dedicated `PreflightError` reads way better than `SystemExit` — it makes the intent obvious and gives tests something specific to assert against. I'll swap both preflight functions over to raise `PreflightError` and let the top-level startup catch it and exit. On it!

FidoCanCode: oh yeah, good call —  is a bit blunt for a library-ish function. i'll introduce  and raise that instead! *tail wag* 🐾

rhencke (approved):

PR #444, opened 2026-04-14
Stop watchdog from force-restarting stale-but-live worker threads

### LINKED ISSUE #443: Forced worker restart loops because lock file 'another fido is running'
## Observed

Watchdog detected confusio thread stuck, forced restart (new in #382/#426). Every restart attempt since then exits with:

\`\`\`
WARNING [confusio] another fido is running — exiting
\`\`\`

~20+ consecutive failures, all ~5 seconds apart. Required manual intervention: SIGTERM kennel, delete lock files, full restart.

## Root cause

\`lsof\` on the lock file at intervention time showed the kennel process holding the lock via MULTIPLE file descriptors — including from the \`worker-home\` thread:

\`\`\`
kennel  1803117         fd=8wW   /home/rhencke/workspace/confusio/.git/fido/lock (deleted)
kennel  1803117         fd=12wW  /home/rhencke/workspace/confusio/.git/fido/lock
kennel  1803117 worker-ho  fd=8wW  /home/rhencke/workspace/confusio/.git/fido/lock (deleted)
\`\`\`

- Worker-home had leaked a reference to worker-confusio's lock file
- The forced-restart path killed the old worker thread but didn't close its open fds
- The fcntl flock persisted on the leaked fd
- Removing the lock file from disk did NOT release the lock (lock lives on the inode/fd, not the path)

## Expected behavior

When the watchdog forces a restart:

1. Close ALL fds the old worker owned before starting the new worker (flock releases on fd close)
2. Don't leak fds across worker threads in the first place — each worker's lock fd should be owned exclusively by that thread
3. The \"another fido is running\" check should validate the lock is held by a live PID, not just any fd in this process

## Related

- #382 — stuck worker detection (where the restart was introduced)
- #430 — detached reattachable workers (different solution, but would sidestep this)

## Severity

Critical when it happens — worker permanently locked out, no automatic recovery, required manual intervention to kill kennel, delete lock files, restart.

FidoCanCode:
## Summary

When the watchdog forced a restart of a stale thread, the old thread often didn't exit within \`stop_and_join\`'s 30s timeout (blocked inside a long claude subprocess). The watchdog then started a new thread on top of the still-running old one, both raced on the fido lockfile, and the new thread looped forever on \"another fido is running\".

Fix: remove the stale-restart path entirely. Watchdog now only restarts threads that are actually not-alive. Claude has its own idle timeout, so a worker that looks stuck is either waiting for claude to finish or will bubble up a timeout error from the subprocess on its own.

\`_STALE_THRESHOLD\` kept as a display-only constant for the \`/status\` endpoint's \`is_stuck\` flag.

Fixes #443

## Test plan

- [x] 1644 tests pass, 100% coverage
- [x] Dead threads still get restarted with crash recording
- [x] Live threads are never restarted, even when stale
- [x] is_stale is not called anywhere in the watchdog
- [x] stop_and_join is never called in the watchdog

*no more lock loops*

### DIFF
--- a/kennel/status.py
+++ b/kennel/status.py
@@ -379,7 +379,7 @@ def format_status(status: KennelStatus) -> str:
             parts.append(claude_str)
 
         if repo.worker_stuck:
-            parts.append("STUCK")
+            parts.append("BUSY")
 
         if repo.crash_count > 0:
             crash_str = f"crashed {repo.crash_count}x"
--- a/kennel/watchdog.py
+++ b/kennel/watchdog.py
@@ -1,4 +1,4 @@
-"""Watchdog — check WorkerThread health and restart dead or stuck threads."""
+"""Watchdog — check WorkerThread health and restart dead threads."""
 
 from __future__ import annotations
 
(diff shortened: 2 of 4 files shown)


rhencke:
Instead of stuck let's say busy?

rhencke (approved):

PR #456, opened 2026-04-14
Drive a single persistent claude per worker via bidirectional stream-json (closes #455)

### LINKED ISSUE #455: Drive a single persistent claude per worker via bidirectional stream-json
## Idea

Replace per-task \`claude --print\` subprocess spawns with one long-lived claude per worker, fed by \`--input-format stream-json --output-format stream-json\`.  Kennel becomes a real state machine driving claude around the loop instead of a subprocess shepherd.

## Why

- **Spawn cost**: every task currently pays ~10s claude boot.  Persistent session = paid once at worker start.
- **Context coherence**: plan + implement + nudge + recovery all share the same transcript.  Today each task starts fresh or resumes a stale session; cross-task reasoning is lost.
- **Model switching**: send \`/model claude-opus-4-6\` for planning / triage / rescope, \`/model claude-sonnet-4-6\` for implementation — same session, no spawn cost, full context preserved.
- **Live eventing**: stream-json events arrive as they happen.  Worker reacts to tool calls, status updates, and completion in real time instead of parsing post-hoc dumps.
- **Smarter nudges**: nudge logic moves into the same conversation that produced the stuck state — claude sees its own \"I think it's done\" reasoning and can act on the contradiction.

## Sketch

\`\`\`python
class ClaudeSession:
    def __init__(self, system_file, work_dir):
        self.proc = subprocess.Popen(
            ["claude", "--input-format", "stream-json",
             "--output-format", "stream-json", "--verbose",
             "--system-prompt-file", str(system_file)],
            stdin=PIPE, stdout=PIPE, cwd=work_dir, text=True,
        )
    
    def send(self, content): ...
    def switch_model(self, model): self.send_command(f"/model {model}")
    def stream_events(self): ...
    def stop(self): ...

class WorkerStateMachine:
    def run(self):
        while not self.done:
            self.session.switch_model("claude-opus-4-6")
            self.session.send(self.plan_prompt())
            plan = self.consume_until_done()
            
            self.session.switch_model("claude-sonnet-4-6")
            self.session.send(self.impl_prompt(plan))
            self.consume_until_committed_or_nudge_threshold()
\`\`\`

## Risks

- Long-lived processes accumulate context — at some point we have to compact or restart.  Need a graceful boundary (per-issue? per-PR?).
- Process death = lost session.  Need supervisor + reattach (#430 covers some of this).
- Slash commands in stream-json — we use \`/model\`; need to verify the input format actually accepts other slash commands too.

## Migration

This is a substantial rewrite of \`kennel/claude.py\` and \`kennel/worker.py\`'s execute loop.  Probably one PR for the persistent-session abstraction (no behavior change), then incremental PRs migrating each spawn site.

## Why v1

The whole reliability batch we've been chasing (#369 backoff, #452 nudges, #267 no-commit loops, etc.) is *symptomatic* of the per-task-spawn model.  A persistent session where claude sees its own history makes most of that go away — the session knows it's been told to commit twice already, knows the previous run claimed completion, can self-correct.  This is the real fix.

FidoCanCode:
Introduces a persistent ClaudeSession that keeps a single stream-json subprocess alive across a worker's setup, task, nudge, thread, and CI phases, switching between Opus for planning/triage and Sonnet for implementation, and bounds context growth by restarting on issue boundaries. Remaining work fixes a session teardown bug so cleanup only runs on orderly shutdown (with a crash-recovery test and an outer try/finally around `stop()` to always unregister the child), resolves a lost-interrupt race from clearing `_cancel` on every lock acquisition, sends a stream-json `control_request` interrupt in `ClaudeSession.interrupt` before follow-up content, and removes the durable preempt queue since it has no production path.

Fixes #455.

---

## Work queue

<details><summary>Completed (40)</summary>

- [x] [Blocking: clearing `_cancel` on every lock acquisition creates a lost-interrupt](https://github.com/FidoCanCode/home/pull/456#discussion_r3080536956) 
- [x] [Woof, you are absolutely right — caught me red-pawed! That unconditional stop()](https://github.com/FidoCanCode/home/pull/456#discussion_r3080537674) 
- [x] PR comment: fix session teardown — only on orderly shutdown, add crash-recovery test 
- [x] [wrap stop() body in outer try/finally to always unregister child](https://github.com/FidoCanCode/home/pull/456#discussion_r3080536964) 
- [x] [send stream-json control_request interrupt in ClaudeSession.interrupt before sending follow-up content](https://github.com/FidoCanCode/home/pull/456#discussion_r3080537011) 
- [x] [High: this persists a durable preempt queue, but I cant find a production path t](https://github.com/FidoCanCode/home/pull/456#discussion_r3080536979) 
- [x] PR comment: remove unnecessary default model setting 
- [x] [replace ClaudeStreamError(-1) magic number with a named constant or sentinel](https://github.com/FidoCanCode/home/pull/456#discussion_r3078130409) 
- [x] [replace repeated truncation slicing with a helper that wraps log calls](https://github.com/FidoCanCode/home/pull/456#discussion_r3078089657) 
- [x] [use context manager for lock acquire/release in interrupt](https://github.com/FidoCanCode/home/pull/456#discussion_r3078119263) 
- [x] [reuse State's thread-safe JSON base class with modify method for the preempt intent queue](https://github.com/FidoCanCode/home/pull/456#discussion_r3077914220) 
- [x] [surface errors instead of swallowing them silently](https://github.com/FidoCanCode/home/pull/456#discussion_r3078080087) 
- [x] [Reraise the exception](https://github.com/FidoCanCode/home/pull/456#discussion_r3078136917) 
- [x] [reraise exceptions in restart cleanup instead of swallowing](https://github.com/FidoCanCode/home/pull/456#discussion_r3078135681) 
- [x] [reraise exceptions in restart() instead of swallowing them](https://github.com/FidoCanCode/home/pull/456#discussion_r3078132643) 
- [x] [We can't spawn a new one if the old one running when timeout happens?](https://github.com/FidoCanCode/home/pull/456#discussion_r3078114474) 
- [x] [We have to kill and safe restart. Kind of our mantra.](https://github.com/FidoCanCode/home/pull/456#discussion_r3078157964) 
- [x] [Yes](https://github.com/FidoCanCode/home/pull/456#discussion_r3078160653) 
- [x] Bound context growth: restart ClaudeSession on issue boundary 
- [x] [surface current ClaudeSession owner thread in kennel status line](https://github.com/FidoCanCode/home/pull/456#discussion_r3077897600) 
- [x] [clarify ClaudeSession persists across worker death but not kennel/home restarts](https://github.com/FidoCanCode/home/pull/456#issuecomment-4242164254) 
- [x] [persist preempt intent to durable queue so interrupts survive re-preemption without losing messages](https://github.com/FidoCanCode/home/pull/456#discussion_r3077907037) 
- [x] [preserve ClaudeSession across worker death so webhook-driven interrupts keep the subprocess alive and hand it back when the worker reclaims the lock](https://github.com/FidoCanCode/home/pull/456#discussion_r3077883250) 
- [x] Route execute_task and nudge loop through the persistent ClaudeSession 
- [x] Route handle_threads and handle_ci claude_run calls through the persistent session 
- [x] Remove setup_session_id state and dead --resume plumbing now superseded by session 
- [x] Switch models per phase: Opus for planning/triage, Sonnet for implementation 
- [x] [make interrupt cancel the in-flight turn so the lock holder exits](https://github.com/FidoCanCode/home/pull/456#discussion_r3077863522) 
- [x] [make interrupt cancel the in-flight turn so the lock holder exits](https://github.com/FidoCanCode/home/pull/456#discussion_r3077863522) 
- [x] [extract idle timeout poll interval and event line log-truncation length into named constants](https://github.com/FidoCanCode/home/pull/456#discussion_r3077710926) 
- [x] [log exceptions in stop() cleanup instead of swallowing silently](https://github.com/FidoCanCode/home/pull/456#discussion_r3077714034) 
- [x] [remove JSONDecodeError handler and let the exception propagate](https://github.com/FidoCanCode/home/pull/456#discussion_r3077729802) 
- [x] [Add threading lock and context manager to ClaudeSession for send/receive serialization](https://github.com/FidoCanCode/home/pull/456#discussion_r3077724061) 
- [x] [Expose lock-break path so preempt/interrupt can bypass the session lock](https://github.com/FidoCanCode/home/pull/456#discussion_r3077741867) 
- [x] [Implement preempt via control_request interrupt with queued follow-up turn](https://github.com/FidoCanCode/home/pull/456#issuecomment-4241910137) 
- [x] Add ClaudeSession: bidirectional stream-json subprocess with send/events/stop 
- [x] ClaudeSession: idle-timeout reader and _active_children registration for shutdown 
- [x] ClaudeSession.switch_model via /model slash command with tests 
- [x] ClaudeSession.consume_until_result: drain events until type=result boundary 
- [x] [make interrupt kill the in-flight turn, then acquire the lock before sending](https://github.com/FidoCanCode/home/pull/456#discussion_r3077871290) 
</details>

### DIFF
--- a/kennel/claude.py
+++ b/kennel/claude.py
@@ -811,9 +811,11 @@ class ClaudeSession:
                 self._proc.wait(timeout=1.0)
             except (OSError, ProcessLookupError, subprocess.TimeoutExpired) as exc:
                 log.debug("ClaudeSession.stop: kill/wait failed: %s", exc)
+                raise
         except (OSError, ProcessLookupError) as exc:
             log.debug("ClaudeSession.stop: wait failed: %s", exc)
-        _unregister_child(self._proc)
+        finally:
+            _unregister_child(self._proc)
 
 
 def resume_status(
--- a/kennel/claude.py
+++ b/kennel/claude.py
@@ -90,6 +90,11 @@ def _claude(
     )
 
 
+_RETURNCODE_IDLE_TIMEOUT = -1
+"""Sentinel returncode used in :class:`ClaudeStreamError` when the process is
+killed due to an idle timeout rather than exiting with a real non-zero code."""
+
+
 class ClaudeStreamError(Exception):
     """Raised by _run_streaming when the subprocess exits with a non-zero code or times out."""
 
@@ -249,8 +254,9 @@ def _run_streaming(
     """Run a command, streaming stdout with idle-timeout detection.
 
     Yields each line of stdout as it arrives.  If no new output arrives for
-    *idle_timeout* seconds, the process is killed and ``ClaudeStreamError(-1)``
-    is raised.  If the process exits with a non-zero code,
+    *idle_timeout* seconds, the process is killed and
+    ``ClaudeStreamError(_RETURNCODE_IDLE_TIMEOUT)`` is raised.  If the process
+    exits with a non-zero code,
     ``ClaudeStreamError(returncode)`` is raised.  ``FileNotFoundError``
     propagates naturally if the command is not found.
     """
@@ -282,7 +288,7 @@ def _run_streaming(
                 log.warning("claude idle for %.0fs — killing", idle_timeout)
                 proc.kill()
                 proc.wait()
-                raise ClaudeStreamError(-1)
+                raise ClaudeStreamError(_RETURNCODE_IDLE_TIMEOUT)
 
         proc.wait()
         # Drain any remaining output
@@ -712,7 +718,8 @@ class ClaudeSession:
         Reads lines from stdout, parsing each as JSON.  Stops (and returns)
         when a ``type=result`` or ``type=error`` event is yielded, when the
         process exits (EOF), or when no output arrives for *idle_timeout*
-        seconds (raises :class:`ClaudeStreamError` ``(-1)`` in that case).
+        seconds (raises :class:`ClaudeStreamError` with
+        :data:`_RETURNCODE_IDLE_TIMEOUT` in that case).
 
         Raises ``json.JSONDecodeError`` if a non-empty stdout line cannot be
         parsed — this is a protocol violation from the claude subprocess and
@@ -750,7 +757,7 @@ class ClaudeSession:
                 )
                 self._proc.kill()
                 self._proc.wait()
-                raise ClaudeStreamError(-1)
+                raise ClaudeStreamError(_RETURNCODE_IDLE_TIMEOUT)
 
     def consume_until_result(self) -> str:
         """Drain events for the current turn and return the result text.
--- a/kennel/claude.py
+++ b/kennel/claude.py
@@ -804,6 +804,7 @@ class ClaudeSession:
                 self._proc.stdin.close()
         except OSError as exc:
             log.debug("ClaudeSession.stop: stdin close failed: %s", exc)
+            raise
         try:
             self._proc.wait(timeout=grace_seconds)
         except subprocess.TimeoutExpired:
--- a/kennel/claude.py
+++ b/kennel/claude.py
@@ -779,6 +779,7 @@ class ClaudeSession:
                 )
                 self._proc.kill()
                 self._proc.wait()
+                self.restart()
                 raise ClaudeStreamError(_RETURNCODE_IDLE_TIMEOUT)
 
     def consume_until_result(self) -> str:
--- a/kennel/claude.py
+++ b/kennel/claude.py
@@ -627,6 +627,19 @@ class ClaudeSession:
         self._proc.stdin.write(msg + "\n")
         self._proc.stdin.flush()
 
+    def interrupt(self, content: str) -> None:
+        """Write a user message to stdin, bypassing the session lock.
+
+        This is the explicit lock-break path for preempt and interrupt use
+        cases (e.g. rescope abort, CI-failure cancel) where the lock is
+        already held by the in-flight turn and waiting for it would defeat
+        the purpose of the interrupt.  Only call from threads that have a
+        legitimate need to preempt the current holder; all normal turns
+        should go through the :meth:`__enter__` / :meth:`__exit__` context
+        manager instead.
+        """
+        self.send(content)
+
     def switch_model(self, model: str) -> None:
         """Switch the active model by sending a /model slash command.
 
--- a/kennel/claude.py
+++ b/kennel/claude.py
@@ -562,6 +562,7 @@ class ClaudeSession:
         self._work_dir = work_dir
         self._popen_fn = popen
         self._lock = threading.Lock()
+        self._cancel = threading.Event()
         self._queued_content: str | None = None
         self._proc = self._spawn()
         _register_child(self._proc)
@@ -611,8 +612,13 @@ class ClaudeSession:
         _register_child(self._proc)
 
     def __enter__(self) -> "ClaudeSession":
-        """Acquire the session lock, serializing send/receive across threads."""
+        """Acquire the session lock, serializing send/receive across threads.
+
+        Also clears the cancel event so a turn that follows a preempt or
+        interrupt starts with a clean slate.
+        """
         self._lock.acquire()
+        self._cancel.clear()
         return self
 
     def __exit__(self, *args: object) -> None:
@@ -629,7 +635,12 @@ class ClaudeSession:
         self._proc.stdin.flush()
 
     def interrupt(self, content: str) -> None:
-        """Write a user message to stdin, bypassing the session lock.
+        """Signal the in-flight turn to stop, then send a follow-up message.
+
+        Sets the cancel event so :meth:`iter_events` exits on its next poll
+        cycle, causing the lock holder to finish its ``with session:`` block
+        and release the lock.  Also writes *content* to stdin as a user
+        message (bypassing the lock) so it is queued as the next turn.
 
         This is the explicit lock-break path for preempt and interrupt use
         cases (e.g. rescope abort, CI-failure cancel) where the lock is
@@ -639,6 +650,7 @@ class ClaudeSession:
         should go through the :meth:`__enter__` / :meth:`__exit__` context
         manager instead.
         """
+        self._cancel.set()
         self.send(content)
 
     def preempt(self, content: str) -> None:
@@ -653,6 +665,7 @@ class ClaudeSession:
         Only for preempt paths (rescope abort, CI-failure cancel).  Normal
         turns must go through the context-manager lock.
         """
+        self._cancel.set()
         msg = json.dumps({"type": "control_request", "request": {"type": "interrupt"}})
         assert self._proc.stdin is not None
         self._proc.stdin.write(msg + "\n")
@@ -698,6 +711,9 @@ class ClaudeSession:
         last_activity = time.monotonic()
 
         while True:
+            if self._cancel.is_set():
+                log.debug("ClaudeSession: cancelled — exiting turn early")
+                break
             ready, _, _ = self._selector(
                 [self._proc.stdout], [], [], _SELECT_POLL_INTERVAL
             )
--- a/kennel/claude.py
+++ b/kennel/claude.py
@@ -13,8 +13,6 @@ import uuid
 from collections.abc import Callable, Iterator
 from pathlib import Path
 
-from kennel.state import PreemptQueue
-
 log = logging.getLogger(__name__)
 
 # How many seconds select.select waits for stdout before checking for
@@ -605,7 +603,6 @@ class ClaudeSession:
         idle_timeout: float = 1800.0,
         popen: Callable[..., subprocess.Popen[str]] = subprocess.Popen,
         selector: Callable[..., tuple[list, list, list]] = select.select,
-        preempt_queue: PreemptQueue | None = None,
     ) -> None:
         self._idle_timeout = idle_timeout
         self._selector = selector
@@ -614,7 +611,6 @@ class ClaudeSession:
         self._popen_fn = popen
         self._lock = threading.Lock()
         self._cancel = threading.Event()
-        self._preempt_queue = preempt_queue
         self._owner: str | None = None
         self._proc = self._spawn()
         _register_child(self._proc)
@@ -746,42 +742,6 @@ class ClaudeSession:
             self.consume_until_result()
             self.send(content)
 
-    def preempt(self, content: str) -> None:
-        """Signal the in-flight turn to stop and queue a follow-up user turn.
-
-        Sets the cancel event so :meth:`iter_events` exits on its next poll
-        cycle.  The lock holder's ``with session:`` block then unwinds and
-        releases the lock.  This method acquires the lock (blocking until the
-        holder releases it), appends *content* to the durable FIFO queue, and
-        releases the lock.  Multiple consecutive preempts accumulate — no
-        message is lost even if preempt is called again before the first
-        queued turn is consumed.  The queued content is retrieved one item at
-        a time via :meth:`take_queued_content` by the next caller that holds
-        the lock.
-
-        Only for preempt paths (rescope abort, CI-failure cancel).  Normal
-        turns must go through the context-manager lock.
-        """
-        self._cancel.set()
-        with self._lock:
-            assert self._preempt_queue is not None, (
-                "preempt_queue required for preempt()"
-            )
-            self._preempt_queue.push(content)
-
-    def take_queued_content(self) -> str | None:
-        """Pop and return the oldest queued preempt content, or ``None``.
-
-        Returns ``None`` if the queue is empty.  Call this while holding the
-        session lock (via the context manager) so that reading and removing
-        from the queue is serialized with respect to a concurrent
-        :meth:`preempt` appending to it.
-        """
-        assert self._preempt_queue is not None, (
-            "preempt_queue required for take_queued_content()"
-        )
-        return self._preempt_queue.pop()
-
     def switch_model(self, model: str) -> None:
         """Switch the active model by sending a /model slash command.
 
@@ -800,7 +760,7 @@ class ClaudeSession:
         the lock-handoff window (between the previous holder's :meth:`__exit__`
         and this call) is consumed here rather than immediately aborting the
         new turn.  After that, checks the event on each poll cycle so a
-        concurrent :meth:`interrupt` or :meth:`preempt` can abort mid-turn.
+        concurrent :meth:`interrupt` can abort mid-turn.
 
         Reads lines from stdout, parsing each as JSON.  Stops (and returns)
         when a ``type=result`` or ``type=error`` event is yielded, when the
--- a/kennel/claude.py
+++ b/kennel/claude.py
@@ -552,6 +552,7 @@ class ClaudeSession:
         self,
         system_file: Path,
         work_dir: Path | str | None = None,
+        model: str = "claude-sonnet-4-6",
         idle_timeout: float = 1800.0,
         popen: Callable[..., subprocess.Popen[str]] = subprocess.Popen,
         selector: Callable[..., tuple[list, list, list]] = select.select,
@@ -560,6 +561,7 @@ class ClaudeSession:
         self._selector = selector
         self._system_file = system_file
         self._work_dir = work_dir
+        self._model = model
         self._popen_fn = popen
         self._lock = threading.Lock()
         self._cancel = threading.Event()
@@ -571,6 +573,8 @@ class ClaudeSession:
         """Spawn the claude subprocess with bidirectional stream-json I/O."""
         cmd = [
             "claude",
+            "--model",
+            self._model,
             "--input-format",
             "stream-json",
             "--output-format",
--- a/kennel/claude.py
+++ b/kennel/claude.py
@@ -554,7 +554,6 @@ class ClaudeSession:
         self,
         system_file: Path,
         work_dir: Path | str | None = None,
-        model: str = "claude-sonnet-4-6",
         idle_timeout: float = 1800.0,
         popen: Callable[..., subprocess.Popen[str]] = subprocess.Popen,
         selector: Callable[..., tuple[list, list, list]] = select.select,
@@ -564,7 +563,6 @@ class ClaudeSession:
         self._selector = selector
         self._system_file = system_file
         self._work_dir = work_dir
-        self._model = model
         self._popen_fn = popen
         self._lock = threading.Lock()
         self._cancel = threading.Event()
@@ -576,8 +574,6 @@ class ClaudeSession:
         """Spawn the claude subprocess with bidirectional stream-json I/O."""
         cmd = [
             "claude",
-            "--model",
-            self._model,
             "--input-format",
             "stream-json",
             "--output-format",
--- a/kennel/claude.py
+++ b/kennel/claude.py
@@ -25,6 +25,32 @@ _SELECT_POLL_INTERVAL = 10.0
 # claude subprocess, to keep log records readable.
 _LOG_LINE_TRUNCATE = 200
 
+
+class _Trunc:
+    """Lazy-truncating wrapper for log arguments.
+
+    Pass an instance as a ``%s`` or ``%r`` log argument instead of slicing
+    inline.  The truncation is deferred to format time, so it is free when
+    the log level is disabled.
+
+    Usage::
+
+        log.debug("event: %s", _Trunc(line))
+        log.warning("stderr=%r", _Trunc(result.stderr))
+    """
+
+    __slots__ = ("_s",)
+
+    def __init__(self, s: str) -> None:
+        self._s = s
+
+    def __str__(self) -> str:
+        return self._s[:_LOG_LINE_TRUNCATE]
+
+    def __repr__(self) -> str:
+        return repr(self._s[:_LOG_LINE_TRUNCATE])
+
+
 # Tracked long-running claude subprocesses (the streaming ones), so kennel
 # can clean them up on shutdown.  Short-lived ``subprocess.run`` calls cap
 # at 30s and aren't tracked.  Use ``kill_active_children`` from a signal
@@ -187,10 +213,8 @@ def print_prompt(
             if text:
                 return text
             if result.stderr:
-                log.warning(
-                    "print_prompt: stderr=%r", result.stderr[:_LOG_LINE_TRUNCATE]
-                )
-            log.debug("print_prompt: stdout=%r", result.stdout[:_LOG_LINE_TRUNCATE])
+                log.warning("print_prompt: stderr=%r", _Trunc(result.stderr))
+            log.debug("print_prompt: stdout=%r", _Trunc(result.stdout))
             if attempt < _EMPTY_RETRY_COUNT:
                 log.warning(
                     "print_prompt: empty output on attempt %d — retrying",
@@ -519,13 +543,9 @@ def generate_status_with_session(
                 return text, sid
             if result.stderr:
                 log.warning(
-                    "generate_status_with_session: stderr=%r",
-                    result.stderr[:_LOG_LINE_TRUNCATE],
+                    "generate_status_with_session: stderr=%r", _Trunc(result.stderr)
                 )
-            log.debug(
-                "generate_status_with_session: stdout=%r",
-                result.stdout[:_LOG_LINE_TRUNCATE],
-            )
+            log.debug("generate_status_with_session: stdout=%r", _Trunc(result.stdout))
             if attempt < _EMPTY_RETRY_COUNT:
                 log.warning(
                     "generate_status_with_session: empty output on attempt %d — retrying",
@@ -744,7 +764,7 @@ class ClaudeSession:
                     last_activity = time.monotonic()
                     continue
                 obj = json.loads(line)
-                log.debug("ClaudeSession event: %s", line[:_LOG_LINE_TRUNCATE])
+                log.debug("ClaudeSession event: %s", _Trunc(line))
                 last_activity = time.monotonic()
                 yield obj
                 if obj.get("type") in ("result", "error"):
--- a/kennel/claude.py
+++ b/kennel/claude.py
@@ -519,6 +519,129 @@ def generate_status_with_session(
     return "", ""
 
 
+# ── Persistent bidirectional session ─────────────────────────────────────────
+
+
+class ClaudeSession:
+    """A long-lived claude process driven via bidirectional stream-json.
+
+    Spawns ``claude --input-format stream-json --output-format stream-json``
+    and keeps it running across multiple turns.  Each :meth:`send` writes one
+    JSON user message to stdin; :meth:`iter_events` reads structured events
+    from stdout until the turn completes (``type=result`` or ``type=error``).
+
+    Pass *popen* and *selector* to override subprocess creation and
+    ``select.select`` in tests; these default to the real implementations.
+    """
+
+    def __init__(
+        self,
+        system_file: Path,
+        work_dir: Path | str | None = None,
+        idle_timeout: float = 1800.0,
+        popen: Callable[..., subprocess.Popen[str]] = subprocess.Popen,
+        selector: Callable[..., tuple[list, list, list]] = select.select,
+    ) -> None:
+        self._idle_timeout = idle_timeout
+        self._selector = selector
+        cmd = [
+            "claude",
+            "--input-format",
+            "stream-json",
+            "--output-format",
+            "stream-json",
+            "--verbose",
+            "--dangerously-skip-permissions",
+            "--system-prompt-file",
+            str(system_file),
+        ]
+        self._proc = popen(
+            cmd,
+            stdin=subprocess.PIPE,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            text=True,
+            cwd=work_dir,
+        )
+        _register_child(self._proc)
+
+    def send(self, content: str) -> None:
+        """Write a user message to the session stdin, flushing immediately."""
+        msg = json.dumps(
+            {"type": "user", "message": {"role": "user", "content": content}}
+        )
+        assert self._proc.stdin is not None
+        self._proc.stdin.write(msg + "\n")
+        self._proc.stdin.flush()
+
+    def iter_events(self) -> Iterator[dict]:
+        """Yield parsed stream-json events for the current turn.
+
+        Reads lines from stdout, parsing each as JSON.  Stops (and returns)
+        when a ``type=result`` or ``type=error`` event is yielded, when the
+        process exits (EOF), or when no output arrives for *idle_timeout*
+        seconds (raises :class:`ClaudeStreamError` ``(-1)`` in that case).
+
+        Unparseable lines are logged at WARNING and skipped.
+        """
+        assert self._proc.stdout is not None
+        last_activity = time.monotonic()
+
+        while True:
+            ready, _, _ = self._selector([self._proc.stdout], [], [], 10.0)
+            if ready:
+                line = self._proc.stdout.readline()
+                if not line:
+                    break  # EOF
+                line = line.strip()
+                if not line:
+                    last_activity = time.monotonic()
+                    continue
+                try:
+                    obj = json.loads(line)
+                except json.JSONDecodeError:
+                    log.warning("ClaudeSession: unparseable line: %r", line[:200])
+                    last_activity = time.monotonic()
+                    continue
+                log.debug("ClaudeSession event: %s", line[:200])
+                last_activity = time.monotonic()
+                yield obj
+                if obj.get("type") in ("result", "error"):
+                    break
+            elif self._proc.poll() is not None:
+                break  # process exited
+            elif time.monotonic() - last_activity > self._idle_timeout:
+                log.warning(
+                    "ClaudeSession: idle for %.0fs — killing", self._idle_timeout
+                )
+                self._proc.kill()
+                self._proc.wait()
+                raise ClaudeStreamError(-1)
+
+    def stop(self, grace_seconds: float = 2.0) -> None:
+        """Shut down the session: close stdin, wait for exit, kill if needed.
+
+        Always unregisters the process from ``_active_children``, even if the
+        process has already exited before :meth:`stop` is called.
+        """
+        try:
+            if self._proc.stdin and not self._proc.stdin.closed:
+                self._proc.stdin.close()
+        except OSError:
+            pass
+        try:
+            self._proc.wait(timeout=grace_seconds)
+        except subprocess.TimeoutExpired:
+            try:
+                self._proc.kill()
+                self._proc.wait(timeout=1.0)
+            except OSError, ProcessLookupError, subprocess.TimeoutExpired:
+                pass
+        except OSError, ProcessLookupError:
+            pass
+        _unregister_child(self._proc)
+
+
 def resume_status(
     session_id: str,
     prompt: str,
--- a/kennel/claude.py
+++ b/kennel/claude.py
@@ -676,11 +676,8 @@ class ClaudeSession:
         where the in-flight turn must be cancelled and a follow-up sent.
         """
         self._cancel.set()
-        self._lock.acquire()
-        try:
+        with self._lock:
             self.send(content)
-        finally:
-            self._lock.release()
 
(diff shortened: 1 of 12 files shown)


rhencke:
It would be cool if kennel status could show (uptime, events/min ↓, commands/min ↑) avg last 10min moving window?

FidoCanCode:
Ooh that's a fun idea — rolling 10min window of uptime, events/min, and commands/min would make `kennel status` way more useful for spotting when something's off. *tail wag*

But it's out of scope for this PR — I want to keep this one focused on the persistent claude stream work. I've opened #457 to track it: https://github.com/FidoCanCode/home/issues/457

rhencke:
It would be nice to surface when compaction is running in `kennel status` like COMPACTING

FidoCanCode:
Ooh that's a great idea — surfacing `COMPACTING` in `kennel status` would be really handy for knowing when to expect a pause. *tail wag* But it's a bit outside the scope of this PR, which is just about the persistent claude stream plumbing. I've filed #458 to track it so it doesn't get lost!

rhencke:
You're going to need to determine how to interrupt Claude with a streaming event to preempt for things like ci failures first or rescopes

FidoCanCode:
Woof, good catch — want to make sure I sniff in the right direction here. A few ways I could go:

1. Send a `control_request` with `interrupt` over the stream-json stdin (the SDK's built-in cancel), then queue a fresh user turn describing the CI failure / rescope.
2. Close/reopen the session entirely on preempt (heavier, loses context).
3. Inject a high-priority user message mid-turn without interrupting, and trust Claude to pivot.

My gut says (1) — it's the cleanest and matches what stream-json is designed for. But do you have a preference, or a specific failure mode you're worried about (e.g. tool calls in flight, partial edits)? Want to make sure I chase the right ball. 🐾

rhencke:
> Woof, good catch — want to make sure I sniff in the right direction here. A few ways I could go:
> 
> 1. Send a `control_request` with `interrupt` over the stream-json stdin (the SDK's built-in cancel), then queue a fresh user turn describing the CI failure / rescope.
> 2. Close/reopen the session entirely on preempt (heavier, loses context).
> 3. Inject a high-priority user message mid-turn without interrupting, and trust Claude to pivot.
> 
> My gut says (1) — it's the cleanest and matches what stream-json is designed for. But do you have a preference, or a specific failure mode you're worried about (e.g. tool calls in flight, partial edits)? Want to make sure I chase the right ball. 🐾

1

FidoCanCode:
Woof, going with option (1) then! *tail wag*

I'll send a `control_request` with `interrupt` over stream-json stdin, then queue a fresh user turn describing the CI failure / rescope. Clean, matches what stream-json is built for, and keeps the session warm so we don't lose context.

For tool calls in flight — the interrupt will cancel the current turn, and the follow-up user message will give Claude the new situation to react to. Partial edits stay on disk; the next turn can sniff around and reconcile if needed.

Off to fetch! 🐾

FidoCanCode:
Woof! Just a heads up — the task from your [comment](https://github.com/FidoCanCode/home/pull/456#issuecomment-4241910137) got a little polish on its title ("Implement" is now capitalized). Same scope, same plan. *tail wag* Back to sniffing out the implementation!

kennel/claude.py
```diff
@@ -519,6 +519,153 @@ def generate_status_with_session(
...
+                line = self._proc.stdout.readline()
+                if not line:
+                    break  # EOF
+                line = line.strip()
+                if not line:
+                    last_activity = time.monotonic()
+                    continue
+                try:
+                    obj = json.loads(line)
+                except json.JSONDecodeError:
```
rhencke: Fatal error

kennel/claude.py
```diff
@@ -519,6 +519,153 @@ def generate_status_with_session(
...
+                    break
+            elif self._proc.poll() is not None:
+                break  # process exited
+            elif time.monotonic() - last_activity > self._idle_timeout:
+                log.warning(
+                    "ClaudeSession: idle for %.0fs — killing", self._idle_timeout
+                )
+                self._proc.kill()
+                self._proc.wait()
+                raise ClaudeStreamError(-1)
```
rhencke: Magic numbers

kennel/claude.py
```diff
@@ -519,6 +519,153 @@ def generate_status_with_session(
...
+        except OSError:
+            pass
+        try:
+            self._proc.wait(timeout=grace_seconds)
+        except subprocess.TimeoutExpired:
+            try:
+                self._proc.kill()
+                self._proc.wait(timeout=1.0)
+            except OSError, ProcessLookupError, subprocess.TimeoutExpired:
+                pass
```
rhencke: At least log it

tests/test_claude.py
```diff
@@ -1113,3 +1115,368 @@ def fake_select(rs, ws, xs, t):
...
+        assert kwargs["stdout"] == subprocess.PIPE
+        assert kwargs["text"] is True
+
+    def test_passes_cwd(self, tmp_path: Path) -> None:
+        system_file = tmp_path / "system.md"
+        system_file.write_text("sys")
+        proc = _make_session_proc([])
+        fake_popen = MagicMock(return_value=proc)
+        fake_selector = MagicMock(return_value=([], [], []))
+        ClaudeSession(
```
rhencke, line 1223: Is ClaudeSession thread safe?  Would be nice if we could use a context manager to lock when needed to prevent torn json or two processes interweaving instructions.

rhencke (changes requested):

kennel/claude.py
FidoCanCode: Good catch! Those bare `10.0` and `:200` slices have been bugging me too — they read like mystery numbers. I'll pull them out into named constants (something like `_SELECT_POLL_INTERVAL = 10.0` and `_LOG_LINE_TRUNCATE = 200`) at module scope so they're self-documenting. *tail wag* Thanks!

tests/test_claude.py
FidoCanCode: Good catch! *ears perked up* — right now it really isn't thread-safe, and yeah, two threads writing JSON lines at the same time would be a disaster. Going to add a `threading.Lock` and expose it via a context manager so callers can hold the session across a full send/receive round-trip without tearing or interleaving. Will wire the internal send/recv paths through the same lock too. Woof, thanks for sniffing this out before it bit us!

kennel/claude.py
FidoCanCode: Good catch — I'll log those cleanup exceptions at debug level instead of just swallowing them. That way if something weird happens during shutdown we'll actually see it in the logs. *tail wag*

kennel/claude.py
FidoCanCode: Woof — "Fatal error" on its own doesn't give me much to sniff at. Could you share a bit more? A stack trace, the command you ran, or a link to the failing CI job would help me track it down. *head tilt*

rhencke: This line should not catch the exception and instead let it fail.

FidoCanCode: Good call! *ears perked* Letting it crash is way better than silently swallowing a malformed line — if the protocol breaks we want to know loudly. Dropping the try/except so the JSONDecodeError propagates. 🐾

tests/test_claude.py
rhencke: > Good catch! *ears perked up* — right now it really isn't thread-safe, and yeah, two threads writing JSON lines at the same time would be a disaster. Going to add a `threading.Lock` and expose it via a context manager so callers can hold the session across a full send/receive round-trip without tearing or interleaving. Will wire the internal send/recv paths through the same lock too. Woof, thanks for sniffing this out before it bit us!

Be careful, there's legitimate reasons for someone to break the lock, such as interrupting rescope or ci fails

FidoCanCode: Woof, great point — hadn't thought through the preempt path. *tail goes still, thinking hard*

You're right: if the lock is held during a long send/recv and a rescope or CI failure needs to interrupt, we can't just politely wait — that's the whole reason preempt exists.

Plan: alongside the normal `with session.lock():` context, I'll expose an explicit break path (something like `session.interrupt()` or a `force=True` acquire) that bypasses the lock, sends the interrupt/cancel on the wire, and lets the holding thread unwind cleanly on its next recv. The lock protects against accidental interleaving in the happy path; preempt stays a first-class escape hatch. Will document which callers are allowed to use it so we don't paper over real races.

Arf, thanks for catching this before I shipped a deadlock!

FidoCanCode:
Woof! Just a heads up — the task from your [comment](https://github.com/FidoCanCode/home/pull/456#issuecomment-4241910137) got updated to match new requirements. Back to sniffing! *tail wag*

kennel/claude.py
```diff
@@ -519,6 +533,235 @@ def generate_status_with_session(
...
+
+        This is the explicit lock-break path for preempt and interrupt use
+        cases (e.g. rescope abort, CI-failure cancel) where the lock is
+        already held by the in-flight turn and waiting for it would defeat
+        the purpose of the interrupt.  Only call from threads that have a
+        legitimate need to preempt the current holder; all normal turns
+        should go through the :meth:`__enter__` / :meth:`__exit__` context
+        manager instead.
+        """
+        self.send(content)
```
rhencke: But you also need the existing lock holder to die

rhencke (changes requested):

kennel/claude.py
FidoCanCode: Woof, good catch — you're right, just writing to stdin doesn't help if the current holder is still blocked waiting on its own turn to finish.

I'll make `interrupt()` actually cancel the in-flight turn so the holder raises out and releases the lock. Thinking: send a control request to abort the current turn (or close/signal as needed), have `iter_events` surface that as an exception on the holder's thread, and only then write the new user message. That way the preempt is real — the old turn dies, the lock comes free, and the interrupter (or the next waiter) gets a clean session to drive.

*tail wagging* — will push that in the next revision.

rhencke: Also you really shouldn't bypass the lock here.  Break the old lock.holder kill it THEN grab the lock and do your thing.

FidoCanCode: Good catch, you're right — bypassing the lock is sketchy. *ears perk up* I'll rework `interrupt()` to signal the current holder to bail (kill the in-flight turn / restart the proc as needed), then acquire the lock properly before sending the new message. No more sneaking past the lock. Woof, much cleaner.

FidoCanCode: Ooh, good catch — you're right on both counts! I shouldn't be sneaking past the lock, and I need to make sure the old holder is actually gone before I muscle in. The right sequence is: find the lock holder → kill it → *then* grab the lock properly. I've queued this up to fix! 🐾

rhencke: > Good catch, you're right — bypassing the lock is sketchy. *ears perk up* I'll rework `interrupt()` to signal the current holder to bail (kill the in-flight turn / restart the proc as needed), then acquire the lock properly before sending the new message. No more sneaking past the lock. Woof, much cleaner.

If you do it right the Claude should persist even through the worker death, webhook wins for a minute, worker gets its lock back when ct task scoped first or issues re ranked/tasked.

FidoCanCode: *ears perk up* Ohh yeah, I see what you're going for — keep the claude subprocess alive at a scope above the worker so a worker dying (or getting preempted) doesn't nuke the conversation. The webhook path interrupts, the next worker that grabs the lock inherits the same live session and picks up where we left off.

I'll hoist `ClaudeSession` ownership out of the worker so its lifetime isn't tied to worker death, and make `interrupt()` signal-and-wait instead of lock-skipping. That way the proc persists across worker handoffs and the lock still means what it says. Woof, good call — much nicer shape.

rhencke: Also status should surface who is "talking to" Claude (worker thread? Webhook?) by showing Claude stats only on the line of the thread currently "owning" it.  Like they pass around the Claude as they handle events.

FidoCanCode: Woof, good call! I'll track which thread currently holds the session lock (worker vs webhook handler) and only render the Claude stats on that thread's status line — so you can actually see the baton getting passed around as events come in. *tail wagging* Will wire it up in this PR.

rhencke: Also be careful with preemption logic.  What if you get preempted while preempted?  We have to do things in such a way where we persist intent then follow through so if interrupted your work can complete.  Or something.  I just don't want any holes where we can lose messages/events/etc.

FidoCanCode: Good catch — *ears perked up*. Right now if `interrupt()` fires while we're already mid-interrupt, the second caller just writes to stdin and hopes for the best. That's exactly the hole you're worried about.

My plan: persist preempt intent to a durable queue (per-session, on-disk) before touching stdin. Each interrupt appends an intent record; a single drainer thread owns the actual write to claude. If we get preempted again — or the process dies, or we restart — the queue is still there, and the drainer picks up where it left off. Intent is recorded first, delivery follows. Nothing lost even if something interrupts the interrupter.

I'll wire that into this PR before marking it ready. Woof, thanks for catching it. 🐾

FidoCanCode: Woof! Marking this one done — already covered by recent commits. Thanks for the nudge on https://github.com/FidoCanCode/home/pull/456#discussion_r3077871290! *tail wag*

rhencke: You can probably use the thread safe json base class State class uses with modify method

FidoCanCode: Ooh good call, thank you! *tail wag* — I'll refactor the preempt intent queue to sit on top of that same thread-safe JSON base class with `modify`, so I'm not rolling my own locking here. Way cleaner and keeps the concurrency story consistent with how `State` already works. Woof, fetching that now.

rhencke:
Claude sessions should persist across *worked* death but not *home/kennel* death

FidoCanCode:
Woof, good clarification! You're right — `ClaudeSession` persists across worker thread death (so a crashed worker can reconnect to its long-running claude process), but not across kennel/home restarts. When the kennel process itself dies, the child claude processes go with it. I'll make sure the code and docs reflect that scope clearly. *tail wag*

kennel/claude.py
```diff
@@ -501,9 +511,13 @@ def generate_status_with_session(
...
-                    "generate_status_with_session: stderr=%r", result.stderr[:200]
+                    "generate_status_with_session: stderr=%r",
+                    result.stderr[:_LOG_LINE_TRUNCATE],
                 )
-            log.debug("generate_status_with_session: stdout=%r", result.stdout[:200])
+            log.debug(
+                "generate_status_with_session: stdout=%r",
+                result.stdout[:_LOG_LINE_TRUNCATE],
+            )
             if attempt < _EMPTY_RETRY_COUNT:
```
rhencke, line 553: Really need to switch as many retries as we can to https://github.com/jd/tenacity decorators

FidoCanCode: Oh that's a good sniff! Tenacity would clean these retry loops up so much — I love it. But it's a bigger swap than I want to sneak into this PR, so I filed #461 to track it: https://github.com/FidoCanCode/home/issues/461. I'll fetch it properly there! 🐾

kennel/claude.py
```diff
@@ -519,6 +533,256 @@ def generate_status_with_session(
     return "", ""
```
rhencke, line 566: No silent failures.

FidoCanCode: Woof, got it — no more silent failures. I'll surface the errors instead of swallowing them so we can actually see when something goes wrong. *tail wagging* On it in this PR.

kennel/claude.py
```diff
@@ -501,9 +511,13 @@ def generate_status_with_session(
...
             if result.stderr:
                 log.warning(
-                    "generate_status_with_session: stderr=%r", result.stderr[:200]
+                    "generate_status_with_session: stderr=%r",
+                    result.stderr[:_LOG_LINE_TRUNCATE],
                 )
-            log.debug("generate_status_with_session: stdout=%r", result.stdout[:200])
+            log.debug(
+                "generate_status_with_session: stdout=%r",
+                result.stdout[:_LOG_LINE_TRUNCATE],
```
rhencke: Too much repetition still.
_log_line_truncate(result.stdout) better but look for even better.

rhencke (changes requested):

kennel/claude.py
FidoCanCode: Woof, good point — the slicing still shows up everywhere even with the constant. I'll pull it into a small helper that wraps the log call itself, so each site just says what it wants logged without repeating the truncation dance. *tail wagging*

kennel/claude.py
```diff
@@ -519,6 +533,256 @@ def generate_status_with_session(
...
+        self._queued_content: str | None = None
+        self._proc = self._spawn()
+        _register_child(self._proc)
+
+    def _spawn(self) -> subprocess.Popen[str]:
+        """Spawn the claude subprocess with bidirectional stream-json I/O."""
+        cmd = [
+            "claude",
+            "--model",
+            self._model,
```
rhencke: We don't need a default model at all if the very first thing we do is /model claude-opus-4-6 entering planning mode which we gonna need anyway.

rhencke (changes requested):

kennel/claude.py
FidoCanCode: Woof, good point! *tail wag* You're right — if we're switching to opus for planning right out of the gate, baking a default into `ClaudeSession` is just noise. I'll drop the `model` parameter entirely and have the session fire `/model claude-opus-4-6` on startup before anything else. One source of truth, less config surface. On it!

FidoCanCode: Ooh good catch! 🐾 You're totally right — if we're setting the model explicitly with `/model claude-opus-4-6` at the top of planning mode anyway, carrying around a default is just dead weight. I'll sniff that out and remove it!

kennel/claude.py
```diff
@@ -519,6 +533,256 @@ def generate_status_with_session(
...
+        re-sending any context the new process needs.
+        """
+        log.warning("ClaudeSession: restarting after unexpected process death")
+        _unregister_child(self._proc)
+        try:
+            self._proc.kill()
+            self._proc.wait(timeout=1.0)
+        except OSError, ProcessLookupError, subprocess.TimeoutExpired:
+            pass
+        self._proc = self._spawn()
```
rhencke, line 661: We can't spawn a new one if the old one running when timeout happens?

kennel/claude.py
```diff
@@ -519,6 +533,256 @@ def generate_status_with_session(
...
+        cycle.  The lock holder's ``with session:`` block then unwinds and
+        releases the lock.  This method acquires the lock (blocking until the
+        holder releases it), sends *content* as a normal user message, and
+        releases the lock.  No stdin write ever races with the holder's writes.
+
+        For preempt/interrupt use cases (rescope abort, CI-failure cancel)
+        where the in-flight turn must be cancelled and a follow-up sent.
+        """
+        self._cancel.set()
+        self._lock.acquire()
```
rhencke: Context manager?

kennel/claude.py
```diff
@@ -519,6 +533,256 @@ def generate_status_with_session(
...
+        releases the lock.  This method acquires the lock (blocking until the
+        holder releases it), stores *content* as the queued follow-up turn,
+        and releases the lock.  The queued content is retrieved via
+        :meth:`take_queued_content` by the next caller that holds the lock.
+
+        Only for preempt paths (rescope abort, CI-failure cancel).  Normal
+        turns must go through the context-manager lock.
+        """
+        self._cancel.set()
+        self._lock.acquire()
```
rhencke: Context manager?

kennel/claude.py
```diff
@@ -519,6 +533,256 @@ def generate_status_with_session(
...
+
+    def switch_model(self, model: str) -> None:
+        """Switch the active model by sending a /model slash command.
+
+        Sends ``/model <model>`` as a user message and drains any response
+        events so the turn boundary is clean before the next call to
+        :meth:`send` + :meth:`iter_events`.
+        """
+        self.send(f"/model {model}")
+        for _ in self.iter_events():
```
rhencke, line 753: I hope iter_events raises if it sees an error.

kennel/claude.py
```diff
@@ -519,6 +533,256 @@ def generate_status_with_session(
...
+                    break
+            elif self._proc.poll() is not None:
+                break  # process exited
+            elif time.monotonic() - last_activity > self._idle_timeout:
+                log.warning(
+                    "ClaudeSession: idle for %.0fs — killing", self._idle_timeout
+                )
+                self._proc.kill()
+                self._proc.wait()
+                raise ClaudeStreamError(-1)
```
rhencke: Magic number -1

kennel/claude.py
```diff
@@ -519,6 +533,256 @@ def generate_status_with_session(
...
+        """Shut down the session: close stdin, wait for exit, kill if needed.
+
+        Always unregisters the process from ``_active_children``, even if the
+        process has already exited before :meth:`stop` is called.
+        """
+        try:
+            if self._proc.stdin and not self._proc.stdin.closed:
+                self._proc.stdin.close()
+        except OSError as exc:
+            log.debug("ClaudeSession.stop: stdin close failed: %s", exc)
```
rhencke: Reraise

kennel/claude.py
```diff
@@ -519,6 +533,256 @@ def generate_status_with_session(
...
+        try:
+            self._proc.wait(timeout=grace_seconds)
+        except subprocess.TimeoutExpired:
+            try:
+                self._proc.kill()
+                self._proc.wait(timeout=1.0)
+            except (OSError, ProcessLookupError, subprocess.TimeoutExpired) as exc:
+                log.debug("ClaudeSession.stop: kill/wait failed: %s", exc)
```
rhencke: Reraise

rhencke: Reraise

rhencke (changes requested):

kennel/claude.py
FidoCanCode: Good question! Right now, no — when the timeout fires, the old worker is still running and we just wait on it. If we wanted to spawn a fresh one, we'd need to actively kill the old process first (SIGTERM/SIGKILL on the subprocess) before launching a replacement. Otherwise we'd have two fidos stepping on each other's paws in the same work dir — merge conflicts, duplicate commits, general chaos. *nervous tail*

If killing + respawning is the behavior we want, happy to chase that down in a follow-up — just wanted to flag that it's a bigger change than just "start another one."

kennel/claude.py
FidoCanCode: Good catch — switching `interrupt` to use `with self._lock:` so the acquire/release is symmetric with `__enter__`/`__exit__`. Cleaner and matches the pattern holders already use. *tail wag* 🐾

kennel/claude.py
FidoCanCode: Good catch — that `-1` is definitely a sniff test failure. I'll swap it for a named constant (something like `CLAUDE_IDLE_TIMEOUT_EXIT_CODE = -1`) at module scope so the intent is obvious at the raise site. *tail wag* thanks for flagging!

kennel/claude.py
FidoCanCode: Woof, good eye! Quick question before I chase this down — are you asking me to rewrite `preempt` to use a `with self._lock:` block (matching `interrupt`'s style) for consistency, or is there a deeper concern about the lock/queue semantics I'm missing? Want to make sure I'm sniffing the right tree before I dig. 🐾

kennel/claude.py
FidoCanCode: Good catch — you're right. Swallowing OSError in the restart cleanup path hides real problems. I'll narrow the except to just `ProcessLookupError` and `subprocess.TimeoutExpired` (the expected "already gone" / "didn't die fast enough" cases) and let anything else propagate. Also going to fix the syntax bug there while I'm at it — that tuple needs parens. *tail wag* thanks!

kennel/claude.py
FidoCanCode: Good catch! *ears perked* That bare `except OSError, ProcessLookupError, ...` is actually a syntax error too — I'll fix the tuple and stop swallowing. Restart should either succeed cleanly or bubble the failure up so the worker knows the session is toast. Woof, sloppy of me. Fetching a fix now.

kennel/claude.py
FidoCanCode: Good catch — woof, you're right. I'll reraise instead of swallowing it so a real failure here doesn't get hidden. *tail wag*

kennel/claude.py
FidoCanCode: Noted — checking on this.

kennel/claude.py
rhencke: We have to kill and safe restart.  Kind of our mantra.

FidoCanCode: Woof, got it — kill and safe restart, that's the mantra. I'll make the timeout path terminate the running worker and respawn a fresh one. *tail wagging* thanks for keeping me honest!

kennel/claude.py
rhencke: Yes

FidoCanCode: Woof, on it! Gonna rewrite `preempt` to use the `with self._lock:` context manager so the lock handling stays clean and exception-safe. *tail wagging* — pushing the change shortly.

kennel/worker.py
```diff
@@ -1891,6 +1979,10 @@ def run(self) -> None:
             self.crash_error = f"{type(exc).__name__}: {exc}"
             log.exception("WorkerThread %s: unexpected error", self.name)
             raise
+        finally:
+            if self._session is not None:
+                self._session.stop()
```
rhencke, line 1984: Blocking: this teardown contradicts the new lifetime contract in both `ClaudeSession` and `WorkerThread`. `worker.run()` hands `_session` back in the inner `finally`, but if `Worker.run()` raises, control lands here and we unconditionally stop the session before the watchdog can restart the thread.

That means the "persistent Claude survives worker crashes" path never actually exists in production, and `tests/test_worker.py` currently cements the wrong behavior. Please keep session teardown on orderly thread shutdown only, and add a failing crash-recovery test that proves the replacement worker gets the same live session object after `Worker.run()` raises.

FidoCanCode: Good catch — you're right, that teardown defeats the whole point. *ears down* I'll move the `_session.stop()` out of the unconditional `finally` so it only runs on orderly shutdown, and leave the session alive when `Worker.run()` raises so the watchdog can hand it to the replacement worker.

Adding a crash-recovery test too: force `Worker.run()` to raise, let the watchdog spin up a new worker, and assert the new `Worker` gets the same live `ClaudeSession` instance (not a fresh one). That test should fail on `main` today and pass after the fix — locking in the contract instead of the current bug. Fixing the existing `tests/test_worker.py` cases that encode the wrong behavior as part of the same change. *tail wagging* thanks for catching this before it shipped.

kennel/claude.py
```diff
@@ -519,6 +567,292 @@ def generate_status_with_session(
...
+
+    def __enter__(self) -> "ClaudeSession":
+        """Acquire the session lock, serializing send/receive across threads.
+
+        Also clears the cancel event so a turn that follows a preempt or
+        interrupt starts with a clean slate.  Records the current thread name
+        in :attr:`owner` so status display can show who holds the session.
+        """
+        self._lock.acquire()
+        self._cancel.clear()
```
rhencke: Blocking: clearing `_cancel` on every lock acquisition creates a lost-interrupt window. If a preempt lands after one holder releases the lock but before the next holder acquires it, this line erases the signal and the next turn runs as if nothing happened.

That makes cancellation timing-dependent under contention. Please move cancel ownership to the interrupted-turn/draining path instead of resetting it in `__enter__()`, and add a concurrency test that fires `interrupt()` / `preempt()` across lock handoff.

kennel/claude.py
```diff
@@ -519,6 +567,292 @@ def generate_status_with_session(
...
+
+    def stop(self, grace_seconds: float = 2.0) -> None:
+        """Shut down the session: close stdin, wait for exit, kill if needed.
+
+        Always unregisters the process from ``_active_children``, even if the
+        process has already exited before :meth:`stop` is called.
+        """
+        try:
+            if self._proc.stdin and not self._proc.stdin.closed:
+                self._proc.stdin.close()
```
rhencke: Medium: if `stdin.close()` raises here, we re-raise before reaching the `wait()` / `finally` block that unregisters the child. That leaves a dead or stale proc in `_active_children`, which contradicts the docstring and can confuse later cleanup.

Please wrap the entire `stop()` body in an outer `try/finally` that always calls `_unregister_child(self._proc)`, and add a test where `stdin.close()` fails but the child is still removed from the registry.

kennel/worker.py
```diff
@@ -714,6 +730,51 @@ def teardown_hooks(self, fido_dir: Path, compact_cmd: str, sync_cmd: str) -> Non
...
+        restarts.
+
+        Immediately switches to Opus for the planning phase; the caller is
+        responsible for switching to Sonnet once planning is complete.
+        """
+        persona_file = _sub_dir() / "persona.md"
+        self._session = claude.ClaudeSession(
+            persona_file,
+            work_dir=self.work_dir,
+            preempt_queue=PreemptQueue(fido_dir / "preempt.json"),
```
rhencke: High: this persists a durable preempt queue, but I cant find a production path that ever uses it end-to-end. The higher-priority work paths in `events.py` still call `registry.abort_task()`, `WorkerThread.abort_task()` only sets `_abort_task`, and `execute_task()` notices that bit only after `claude_run()` returns.

So CI/thread preemption still waits for the current Claude turn to finish, and `preempt.json` becomes dead state. Please either wire the abort path to `ClaudeSession.preempt()` / `take_queued_content()` before the next prompt is sent, or drop the durable queue until the worker/event plumbing is in place.

kennel/claude.py
```diff
@@ -519,6 +567,292 @@ def generate_status_with_session(
...
+        releases the lock.  This method acquires the lock (blocking until the
+        holder releases it), sends *content* as a normal user message, and
+        releases the lock.  No stdin write ever races with the holder's writes.
+
+        For preempt/interrupt use cases (rescope abort, CI-failure cancel)
+        where the in-flight turn must be cancelled and a follow-up sent.
+        """
+        self._cancel.set()
+        with self._lock:
+            self.send(content)
```
rhencke, line 743: High: this writes a new user turn after only setting a local event; it never sends the protocol-level stream-json interrupt (`control_request`) that the PR discussion converged on. `iter_events()` exits locally, but the Claude subprocess can still be producing the previous turn.

That leaves the next caller able to inherit unread old-turn output on a dirty boundary. Please make interrupt one owner-controlled sequence: send the actual protocol interrupt, drain the interrupted turn to a clean boundary, then send follow-up content.

kennel/worker.py
rhencke: Woof, you are absolutely right — caught me red-pawed!

That unconditional stop() in the outer finally is a real problem. The whole point of the persistent session is that it survives Worker.run() crashes, but right now the outer teardown yanks it away before the watchdog ever gets a chance to restart. The "survives crashes" path exists in the docs and the architecture section but not in the actual running code. That is a bug, not a feature!

Queued this up: I will move session teardown to orderly-shutdown-only and add a crash-recovery test that verifies the replacement worker inherits the same live session object. *tail wagging intensifies*

kennel/claude.py
FidoCanCode: Good catch — woof, that's a real race. You're right that wiping `_cancel` in `__enter__` clobbers any preempt that landed during the handoff window.

I'll move the reset onto the interrupted-turn / draining path so cancel ownership lives with whoever actually consumed the signal, and add a concurrency test that fires `interrupt()`/`preempt()` right across the lock handoff to lock the behavior in. *tail wagging* — chasing this one down now.

kennel/worker.py
FidoCanCode: Woof, good catch — you're right, the durable queue is stranded without a consumer. *ears perked*

I'll wire the abort path through so preemption actually bites: when `abort_task()` fires, it'll call `ClaudeSession.preempt()` to interrupt the in-flight turn, and `execute_task()` will drain `take_queued_content()` before the next prompt instead of waiting for `claude_run()` to return on its own. That makes `preempt.json` a live queue that CI/thread work actually flows through end-to-end.

Chasing it down now. *tail wagging*

kennel/claude.py
FidoCanCode: Good catch! *ears perked* You're right — the docstring promises always-unregister but a raising `stdin.close()` would skip it. I'll wrap the whole `stop()` body in `try/finally` so `_unregister_child(self._proc)` always runs, and add a test where `stdin.close()` raises and we verify the child is gone from `_active_children`. Woof, fetching it now.

kennel/claude.py
FidoCanCode: Good catch — woof, you're right that just flipping a local event leaves the subprocess still chugging on the old turn, and the next caller would inherit that mess on a dirty boundary.

I'll rework `interrupt()` so it:

1. Acquires the lock (so we're the sole writer on stdin).
2. Sends a stream-json `control_request` interrupt to actually tell the subprocess to stop the in-flight turn.
3. Drains stdout until we hit a clean boundary (`type=result` or the matching `control_response`), discarding the tail of the cancelled turn so nothing leaks forward.
4. *Then* sends the follow-up user message.
5. Releases the lock.

The local cancel event stays — `iter_events()` still needs it so the current owner's read loop unwinds promptly — but it's no longer the only mechanism. The protocol interrupt is what actually stops claude, and the drain is what guarantees the boundary is clean before the next `__enter__` runs.

*tail wagging* — this is a much better shape. Pushing the fix shortly.

kennel/worker.py
FidoCanCode: Woof, thank you for the careful read! You're right — I'll move the `session.stop()` out of the unconditional outer `finally` and into the orderly-shutdown path only, so a crashing `Worker.run()` leaves the session alive for the watchdog to hand to the replacement worker.

Plan for this PR:
- Teardown lives on the orderly-shutdown path; crashes leave the session untouched
- New crash-recovery test: force `Worker.run()` to raise, assert the replacement worker receives the same live session object (identity check, not just equality)
- Update the existing worker tests that baked in the old unconditional-stop behavior so they reflect the new contract

*tail wagging intensifies* — pushing shortly.

FidoCanCode: On it! The unconditional  in the outer  is exactly the bug you're pointing at — it eradicates the session before the watchdog ever gets a shot at restarting. I've queued this up: move teardown to orderly-shutdown-only and add a crash-recovery test that proves the replacement worker inherits the same live session object. *tail wagging intensifies*

kennel/claude.py
FidoCanCode: Caught it! Resetting `_cancel` on every lock acquisition is a genuine lost-interrupt window — a preempt landing between release and acquire just vanishes. Queued: move cancel ownership to the interrupted-turn/draining path and add a concurrency test that fires `interrupt()` / `preempt()` across a lock handoff. Good catch.

kennel/claude.py
FidoCanCode: Yep, that's a real gap — if `stdin.close()` throws, we bail before the `wait()` / `finally` that unregisters the child, leaving a stale proc in `_active_children`. Queued: wrap the entire `stop()` body in an outer `try/finally` that always calls `_unregister_child(self._proc)`, plus a test where `stdin.close()` fails but the child is still removed from the registry.

kennel/worker.py
FidoCanCode: You're right — `preempt.json` is writing to a durable queue that nothing on the production path actually reads yet. The abort path still goes through `_abort_task` and only triggers after `claude_run()` returns. Queued: either wire the abort path through `ClaudeSession.preempt()` / `take_queued_content()` before the next prompt, or drop the durable queue entirely until the worker/event plumbing is in place. No half-wired state.

kennel/claude.py
FidoCanCode: Dead to rights — writing a new user turn without sending the `control_request` protocol interrupt means the Claude subprocess can still be producing the previous turn's output. The next caller inherits a dirty boundary. Queued: make interrupt one owner-controlled sequence — send the actual protocol interrupt, drain the interrupted turn to a clean boundary, *then* send follow-up content.

rhencke (approved):

Issue #473, opened 2026-04-14
Status redesign: claude-talker tracking, tree display, halt on leak


FidoCanCode:
Redesign kennel status display and add exclusivity invariant for claude subprocesses.

## Display (CLI `kennel status`)

One repo block per repo, lines in order:

1. **Worker summary** — one line. Repo name, fido running/idle, uptime, crashes, BUSY. If *nobody else* is currently talking to claude, claude pid/stats appear inline on this line.
2. **Issue:** #N — title (elapsed Xm)
3. **PR:** #N — title
4. **Worker thread details** — state (idle / in-session / waiting). If the worker thread is the one talking to claude, claude pid/stats attach to this line.
5. **Webhook threads** — indented list, up to 5 lines, plus `+N more` summary if >5. Each line: description + elapsed. If a webhook thread is the one talking to claude, its line **sorts to the top** of the webhook section and gets the claude pid/stats attached.

## Telemetry (`/status`)

Per-repo payload gains:
- `session_alive` (bool) — persistent ClaudeSession subprocess is alive
- `claude_talker` (object | null) — who is currently talking to claude:
  - `thread_name`: str — e.g. `worker-home` or `webhook-reply-3080537011`
  - `kind`: `"worker"` | `"webhook"`
  - `description`: str — human label ("replying to review comment 3080537011", "task prompt")
  - `claude_pid`: int — pid of the claude process this thread is driving
  - `started_at`: ISO timestamp

Kennel can still have one persistent session subprocess per repo; a "talker" is only present when some thread is actively sending/receiving.

## Invariant: one talker per repo, enforced

At most one thread may be talking to claude for a given repo at any moment. Concurrent registration from a second thread → raise `ClaudeLeakError` → log + halt kennel with nonzero exit (watchdog/self-restart not used — crash is loud by design).

Rationale: post-#456 every claude call should go through the persistent `ClaudeSession`. Any parallel `print_prompt` or second `with session:` means a webhook handler is leaking its own sub-claude. We want that to fail loud, not silently proliferate.

## Plumbing

New module-level registry in `kennel/claude.py` (or sibling):
```python
@dataclass(frozen=True)
class ClaudeTalker:
    repo_name: str
    thread_name: str
    kind: Literal["worker", "webhook"]
    description: str
    claude_pid: int
    started_at: datetime

def register_talker(...) -> None  # raises ClaudeLeakError if repo already has one
def unregister_talker(repo_name, thread_name) -> None  # idempotent
[...]

PR #481, opened 2026-04-14
Route webhook handlers through persistent ClaudeSession + resume on preempt (closes #479)

### LINKED ISSUE #479: Route webhook handlers through the persistent ClaudeSession with steal/handoff semantics
Today, webhook handlers (`reply_to_comment`, `reply_to_review`, `create_task`, …) spawn one-shot `claude --print` subprocesses via `_run_streaming`.  That's a **second** claude process per repo — the thing the 'one claude per repo' invariant in #473 / #475 explicitly bans.  The leak detector currently fires as soon as a worker is inside `with session:` and a webhook starts a sub-claude.

Needed behavior (per `kennel status` design): webhook handlers should **steal** the persistent session for the duration of their turn and hand it back.

- Worker is in with session:, talker=worker
- Webhook arrives → calls session.interrupt(prompt) → worker's `iter_events` sees cancel, exits, releases lock + unregisters talker
- Webhook acquires lock + registers talker=webhook, sends its prompt, drains result
- Webhook releases lock + unregisters; worker re-acquires on next loop, re-registers talker=worker

Under that model there's never more than one talker per repo; the ClaudeTalker entry in the registry is the authoritative 'who's talking' record that `kennel status` already displays on the right line (#476).  The leak detector keeps its job — two *concurrent* talkers means a sub-claude genuinely leaked.

Work items:

- Replace one-shot `print_prompt` calls in `kennel/events.py` with `session.interrupt(prompt)` + `session.consume_until_result()` on the shared session
- Ensure `handle_threads` / rescope / nudge paths that currently use `print_prompt` also route through the session
- Worker's `iter_events` already exits on cancel (landed earlier) — confirm it also cleanly re-registers the talker when it resumes after the steal
- Keep `_run_streaming` as a last-resort primitive but make sure nothing in the webhook path reaches it without going through the session

Once this lands, we can flip the ClaudeTalker registry's halt-on-leak behavior from a correctness guarantee to a pure production sanity check (since a properly routed system can't hit it).

FidoCanCode:
Fulfills the 'one claude per repo' invariant from #473 / #475 by making webhook handler prompt calls share the worker's persistent \`ClaudeSession\` instead of spawning one-shot subprocesses.

## Steal / handoff

\`ClaudeSession.prompt(content, model=..., system_prompt=...)\` runs one full turn on the persistent session:

1. Set the cancel event — wakes any current holder out of \`iter_events\`.
2. Acquire the session as a context manager (blocks until previous holder's \`__exit__\` releases the lock + unregisters its \`ClaudeTalker\`).
3. Send stream-json \`control_request\` interrupt + drain stale events so stdout is clean.
4. Switch model if requested, send content, return result text.

## Resume on preempt

When \`iter_events\` exits early because \`_cancel\` was set, a new \`last_turn_cancelled\` property records it. The worker's \`claude_start\` / \`claude_run\` now wrap the persistent-session path in \`_run_session_turn\`, which retries preempted turns up to \`_MAX_PREEMPT_RETRIES\` times — 'hand the session back to the worker and ask it to resume what it was doing'. No more crashing on empty result after a webhook steals the session.

## Auto-routing print_prompt

\`claude.print_prompt\` now checks for a session on the current thread's repo via a new \`set_session_resolver()\` hook. When one exists and is alive, the call routes through \`session.prompt()\` instead of spawning a one-shot subprocess. Falls back to one-shot on \`ClaudeStreamError\` / \`ClaudeLeakError\` (returns \`\"\"\` like the existing failure path).

## Wiring

- \`WorkerRegistry.get_session(repo_name)\` exposes the worker thread's \`ClaudeSession\`.
- \`server._run()\` calls \`claude.set_session_resolver(registry.get_session)\` right after constructing the registry.

Closes #479.

### DIFF
--- a/kennel/claude.py
+++ b/kennel/claude.py
@@ -177,6 +177,61 @@ def _talker_now() -> datetime:
     return datetime.now(tz=timezone.utc)
 
 
+_session_resolver: Callable[[str], "ClaudeSession | None"] | None = None
+"""Callback the event/webhook layer uses to find its repo's persistent
+:class:`ClaudeSession` — installed once by :mod:`kennel.server` at startup.
+
+Every in-process prompt call goes through the persistent session, so
+this is a required piece of wiring.  Callers (:func:`print_prompt`) fail
+loud if it's missing — the only time that should happen is a forgotten
+resolver install, not a real production path."""
+
+
+def set_session_resolver(
+    resolver: Callable[[str], "ClaudeSession | None"] | None,
+) -> None:
+    """Install (or clear) the session resolver callback."""
+    global _session_resolver
+    _session_resolver = resolver
+
+
+def _session_for_current_repo() -> "ClaudeSession":
+    """Return the live :class:`ClaudeSession` driving the current thread's repo.
+
+    Production always has both a thread-local ``repo_name`` (set by
+    :func:`set_thread_repo` in the worker thread and webhook handler
+    entrypoints) and an installed :func:`set_session_resolver` callback,
+    and every worker reaches this code after :meth:`Worker.create_session`
+    has populated the session.  So this raises rather than falling back —
+    a missing session is a wiring bug, not a condition callers should
+    paper over.
+    """
+    repo = current_repo()
+    if repo is None:
+        raise RuntimeError(
+            "claude.print_prompt called without a thread-local repo_name — "
+            "server.WebhookHandler._process_action and WorkerThread.run both "
+            "set it; this caller is missing the install."
+        )
+    if _session_resolver is None:
+        raise RuntimeError(
+            "claude.print_prompt called before set_session_resolver — "
+            "server._run() installs it at startup; nothing should run before."
+        )
+    session = _session_resolver(repo)
+    if session is None:
+        raise RuntimeError(
+            f"no ClaudeSession registered for repo {repo} — worker thread "
+            "has not yet created its session"
+        )
+    if not session.is_alive():
+        raise RuntimeError(
+            f"ClaudeSession for repo {repo} is not alive — watchdog should "
+            "have restarted the worker thread"
+        )
+    return session
+
+
 def _thread_name_for_id(thread_id: int) -> str | None:
     """Return the human-readable name of the live thread with *thread_id*,
     or ``None`` if that thread has exited.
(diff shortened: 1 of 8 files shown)


kennel/claude.py
```diff
@@ -177,6 +177,39 @@ def _talker_now() -> datetime:
     return datetime.now(tz=timezone.utc)
 
 
+_session_resolver: Callable[[str], "ClaudeSession | None"] | None = None
+"""Callback the event/webhook layer uses to find its repo's persistent
+:class:`ClaudeSession` — set once by :mod:`kennel.server` at startup.
+
+When unset (tests, tools), session-aware callers fall back to spawning
```
rhencke: When wouldn't we have a Claude session...?

kennel/claude.py
```diff
@@ -177,6 +177,39 @@ def _talker_now() -> datetime:
...
+one-shot ``claude --print`` subprocesses, which preserves the pre-#479
```
rhencke: This smells like error masking.

kennel/claude.py
```diff
@@ -177,6 +177,39 @@ def _talker_now() -> datetime:
...
+
+
+def _session_for_current_repo() -> "ClaudeSession | None":
+    """Return the live session for this thread's repo, or ``None``.
+
+    Resolves via :data:`_session_resolver` using :func:`current_repo`.
+    Returns ``None`` when either is missing or the session isn't alive.
+    """
+    repo = current_repo()
+    if repo is None or _session_resolver is None:
```
rhencke: When wouldn't there be a repo?

kennel/claude.py
```diff
@@ -313,15 +346,25 @@ def print_prompt(
...
+    Best-effort enrichment in the fallback path: nonzero exit, timeout, and
+    missing CLI all return ``""``.  Retries up to ``_EMPTY_RETRY_COUNT``
+    times on empty output.
     """
+    session = _session_for_current_repo()
+    if session is not None:
+        try:
+            return session.prompt(prompt, model=model, system_prompt=system_prompt)
+        except (ClaudeStreamError, ClaudeLeakError) as exc:
+            log.warning("print_prompt: session path failed (%s)", exc)
```
rhencke: No, we fail open.

kennel/worker.py
```diff
@@ -235,6 +233,43 @@ def claude_start(
     return claude.extract_session_id(output)
 
 
+_MAX_PREEMPT_RETRIES = 3
```
rhencke: I may come in with 20 comments at once in a PR review.  I don't think this is the right behavior.  Webhooks should be one shot so there's no preemption loop but if there's a backlog we better off processing it all before handing back regardless of how many preemptions there are because replying to comments is higher priority than development and even ci fixes.

rhencke (changes requested):

rhencke (approved):

PR #512, opened 2026-04-14
Thread task titles pull from reply body instead of root review comment (closes #470)

### LINKED ISSUE #470: Thread task titles pull from reply body instead of root review comment
When a review thread spawns a task, the task title should come from the **root comment** (the reviewer's actual feedback) but often gets populated from a **later reply** in the same thread. Result: tasks with useless or misleading descriptions.

Examples from [PR #456](https://github.com/FidoCanCode/home/pull/456):

- `Woof, you are absolutely right — caught me red-pawed! That unconditional stop()` — that's fido's own reply, not the reviewer's blocking comment
- `Yes` (completed) — one-word reply captured as a task title
- `High: this persists a durable preempt queue, but I cant find a production path t` — root comment but truncated mid-sentence

Fix: when creating or refreshing a thread-typed task, fetch the first comment on the review thread (`reviewThreads.comments.nodes[0].body`) rather than whatever the last event's body field happens to be. Trim/rewrite the title from that source (strip `Blocking:`/`High:`/etc. prefix only if it helps, keep full sentence, cap length consistently).

Bonus: also re-use this source for the PR body's work-queue line so the link text stops drifting as the thread grows.

FidoCanCode:
Fixes #470.

Thread task titles currently pull from the webhook reply body (often fido's own acknowledgment or a one-word "Yes") instead of the reviewer's actual feedback. This fixes title generation to use the root review comment body, which is already fetched via `fetch_comment_thread`.

---

## Work queue

- [x] Use root review comment body for thread task title generation 
- [x] Use originating PR comment for task titles and edit last reply instead of reposting 
- [x] Ensure reply/edit logic properly considers entire thread context 
</details>

### DIFF
--- a/kennel/events.py
+++ b/kennel/events.py
@@ -307,12 +307,22 @@ def reply_to_comment(
 
     context: dict[str, Any] = dict(action.context) if action.context else {}
 
-    # Always fetch the full thread for this comment
+    # Always fetch the full thread for this comment.
+    # Normalize to list so root_body extraction below is type-safe.
+    thread_comments: list[dict[str, Any]] = []
     if info.get("repo") and info.get("pr") and info.get("comment_id"):
-        thread = gh.fetch_comment_thread(info["repo"], info["pr"], info["comment_id"])
-        if thread:
-            context["comment_thread"] = thread
-            log.info("fetched %d comment(s) in thread for context", len(thread))
+        fetched = gh.fetch_comment_thread(info["repo"], info["pr"], info["comment_id"])
+        if fetched:
+            thread_comments = list(fetched)
+            context["comment_thread"] = thread_comments
+            log.info(
+                "fetched %d comment(s) in thread for context", len(thread_comments)
+            )
+
+    # Root comment body — used for task title generation.
+    # When the webhook fires on a reply (e.g. "Yes" or "Woof, you're right!"),
+    # the task title should describe the reviewer's original feedback, not the reply.
+    root_body = thread_comments[0].get("body", comment) if thread_comments else comment
 
     # Enrich context with sibling threads when the comment needs more context
     if (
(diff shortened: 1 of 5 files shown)


rhencke:
This isn't quite right.  The task comment should always reflect the PR comment (on file, line or pr level) that it was created from (with any modifications needed by retasking but still linking to its origin.

If Fido was the last replier and has more to say, it's better to edit the comment with the added context than post a bunch in a row.

FidoCanCode:
Use originating PR comment for task titles and edit replies instead of reposting

rhencke (changes requested):
See comments

FidoCanCode:
Woof! ACT tasks now sniff out their titles from the originating PR comment using `_summarize_as_action_item`. And if I've already replied in the thread, I update my existing comment instead of barking twice. Good boy behavior! 🐕

kennel/events.py
```diff
@@ -307,12 +307,22 @@ def reply_to_comment(
...
+        if fetched:
+            thread_comments = list(fetched)
+            context["comment_thread"] = thread_comments
+            log.info(
+                "fetched %d comment(s) in thread for context", len(thread_comments)
+            )
+
+    # Root comment body — used for task title generation.
+    # When the webhook fires on a reply (e.g. "Yes" or "Woof, you're right!"),
+    # the task title should describe the reviewer's original feedback, not the reply.
```
rhencke, line 324: It's not that simple.  The reply needs to take info context *the entire conversation thread* before replying or amending a reply.

rhencke (changes requested):

kennel/events.py
FidoCanCode: Woof — noted! Could you clarify what you mean by 'take info context'? 

My current implementation:
- Always triages the triggering comment (whether root or reply)
- Includes `comment_thread` (entire thread) in the context passed to Opus
- Generates reply body with full thread context
- Then either edits last Fido comment or posts new

So the reply DOES see the whole thread. Are you saying:
1. The generated reply body should *reference* or *quote* the thread explicitly?
2. The reply should be delayed until all thread replies are in?
3. Something else?

Help me sniff out the right behavior! 🐕

rhencke: > Woof — noted! Could you clarify what you mean by 'take info context'? 
> 
> My current implementation:
> - Always triages the triggering comment (whether root or reply)
> - Includes `comment_thread` (entire thread) in the context passed to Opus
> - Generates reply body with full thread context
> - Then either edits last Fido comment or posts new
> 
> So the reply DOES see the whole thread. Are you saying:
> 1. The generated reply body should *reference* or *quote* the thread explicitly?
> 2. The reply should be delayed until all thread replies are in?
> 3. Something else?
> 
> Help me sniff out the right behavior! 🐕

Include entire thread in opus context

FidoCanCode: Woof! Updated `reply_context_block` to include the full comment thread when generating or amending replies. Now Opus sees the entire conversation history, not just the triggering comment. Good context makes for good replies! 🐕

FidoCanCode: Got it — just shipped this in 91ae478! `reply_context_block` now includes the full `comment_thread` in the prompt sent to Opus for reply generation, same way `triage_context_block` already did. The thread was being fetched but wasn't making it into the actual reply instruction. Now it does. 🐕

rhencke (approved):

PR #515, opened 2026-04-14
Tastefully colorize kennel status output (closes #511)

### LINKED ISSUE #511: Tastefully colorize kennel status output
## Problem

\`kennel status\` is a wall of plain text.  On a glance it's hard to tell which repo is working vs. idle, where the issue number is, and whether something is stuck.  A little ANSI color would make the output scannable at speed — especially when watched live (\`watch -c kennel status\`).

## Design — tasteful, not garish

Use color to carry *meaning*, not decoration.  One color per concept, consistent across runs:

| Element | Style |
|---|---|
| Repo running | bold |
| Repo idle | dim |
| \`BUSY\` / stuck | red |
| Issue number (\`#466\`) | cyan |
| PR number (\`#498\`) | magenta (so issue vs PR read differently at a glance) |
| PR / issue URL | dim cyan/magenta |
| Elapsed times (\`elapsed 3m\`, \`running 54s\`) | dim gray |
| \`session idle\` | dim |
| Active talker kind — \`worker\` | green |
| Active talker kind — \`webhook\` | yellow |
| Task counter currently-in-progress (\`task 2/7\`) | bold |
| Headers (\`kennel: UP\`) | bold |
| Error / crash lines | red bold |

No 256-color rainbows.  No bold everywhere.  When in doubt, dim rather than colorize.

## Environment

- Auto-detect TTY: colorize when \`sys.stdout.isatty()\` is true.
- Respect \`NO_COLOR\` (https://no-color.org): disable color when set.
- Respect \`FORCE_COLOR=1\`: enable color even when not a TTY.  (Makes \`watch -c kennel status\` work without extra flags — the current renderer strips color because \`watch\` doesn't present a TTY.)
- No external dependency — use a tiny ANSI helper (a \`def color(style, text) -> str\` with short constants) rather than pulling in \`rich\` / \`colorama\`.  Kennel is a thin wrapper; a 20-line helper is enough.

## Accessibility

Don't rely on color alone for \`BUSY\` or crash signals — the textual marker stays so color-blind users and log readers don't lose information.  Color amplifies, doesn't replace.

## Test plan

- Snapshot tests should keep using plain text (force \`NO_COLOR\` in the test env so test output diffs stay stable).
- A handful of tests pin the color helper's output for each semantic style.
- Verify \`kennel status\` under \`FORCE_COLOR=1\` emits ANSI escapes; under \`NO_COLOR=1\` does not; under a plain pipe (no TTY, no env vars) does not.

## Priority

v1, high.  Improves day-to-day operator experience significantly for a small surface change.

FidoCanCode:
Fixes #511.

Adds a small zero-dependency ANSI color helper with TTY/NO_COLOR/FORCE_COLOR detection, then wires it into `kennel status` formatting — bold running repos, dim idle ones, cyan issues, magenta PRs, red for stuck/crash, and dim elapsed times. Color amplifies existing textual markers without replacing them.

---

## Work queue

- [x] Add ANSI color helper with TTY/NO_COLOR/FORCE_COLOR detection 
- [x] Colorize kennel status output using semantic styles from #511 
- [x] [Define style constants before _CODES and use them as dictionary keys](https://github.com/FidoCanCode/home/pull/515#discussion_r3083127947) 
</details>

### DIFF
--- a/kennel/color.py
+++ b/kennel/color.py
@@ -1,0 +1,60 @@
+"""ANSI color helper for terminal output.
+
+Color is enabled when:
+- ``FORCE_COLOR=1`` is set in the environment, OR
+- stdout is a TTY and ``NO_COLOR`` is not set.
+
+``NO_COLOR`` follows https://no-color.org — presence of the variable (any
+value) disables color.  ``FORCE_COLOR=1`` overrides both TTY detection and
+``NO_COLOR`` so that tools like ``watch -c`` work without extra flags.
+"""
+
+from __future__ import annotations
+
+import os
+import sys
+
+# ANSI escape sequences
+_RESET = "\033[0m"
+_CODES: dict[str, str] = {
+    "bold": "\033[1m",
+    "dim": "\033[2m",
+    "red": "\033[31m",
+    "red_bold": "\033[1;31m",
+    "cyan": "\033[36m",
+    "magenta": "\033[35m",
+    "green": "\033[32m",
+    "yellow": "\033[33m",
+}
+
+# Semantic style name constants for import convenience
+BOLD = "bold"
+DIM = "dim"
+RED = "red"
+RED_BOLD = "red_bold"
+CYAN = "cyan"
+MAGENTA = "magenta"
+GREEN = "green"
+YELLOW = "yellow"
+
+
+def _color_enabled() -> bool:
+    """Return True if ANSI color output should be used."""
+    if os.environ.get("FORCE_COLOR") == "1":
+        return True
+    if "NO_COLOR" in os.environ:
+        return False
+    return sys.stdout.isatty()
+
+
+def color(style: str, text: str) -> str:
+    """Wrap *text* in ANSI escape codes for *style* if color is enabled.
+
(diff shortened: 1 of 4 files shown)


kennel/color.py
```diff
@@ -0,0 +1,60 @@
...
+
+# Semantic style name constants for import convenience
+BOLD = "bold"
+DIM = "dim"
+RED = "red"
+RED_BOLD = "red_bold"
+CYAN = "cyan"
+MAGENTA = "magenta"
+GREEN = "green"
+YELLOW = "yellow"
```
rhencke: If you reverse these you can use the constant keys in your table.

rhencke (changes requested):

kennel/color.py
FidoCanCode: Good catch! I'll move the constants above `_CODES` so the dict keys can reference them directly instead of duplicating the string literals. 🐕

FidoCanCode: Woof! Moved the style constants above `_CODES` and used them as dictionary keys — no more duplicated strings. Good boy work done! 🐾

rhencke (approved):

PR #519, opened 2026-04-15
Introduce grouped infrastructure ports and composition roots (closes #398)

### LINKED ISSUE #398: Introduce grouped infrastructure ports and composition roots
## Problem

The repo now has many injected objects, but the foundational dependency story is still fragmented across raw callables and one-off seams.

Current examples in the runner clone include:

- `_run`, `_sleep`, `_which`, `_signal`, `_startup_pull`, `_execvp`, and `_print_prompt` parameters threaded through otherwise important runtime functions
- `_GitHub=...` seams on entrypoints instead of one consistent composition-root story
- `server.WebhookHandler` class attributes such as `_fn_dispatch`, `_fn_reply_to_comment`, `_fn_pull_with_backoff`, and `_fn_os_execvp` standing in for real collaborator ownership

If the migration continues in that shape, the repo will end up with objects that are only cosmetically OO: each one will still accept a bag of unrelated stdlib callables, and tests will keep patching procedural seams instead of faking coherent ports.

## Scope

- foundational infrastructure collaborators used by runtime code
- grouped ports for process execution, time/sleep, filesystem/path, and OS/exec concerns where those are still injected as raw callables
- at least one composition root that assembles those collaborators once per app or command execution

## Required design constraints

1. Group low-level dependencies by concern instead of injecting raw stdlib callables one by one into every class or entrypoint.
2. Make runtime wiring explicit in a composition root rather than hidden behind module factories or class-level `_fn_*` slots.
3. Preserve testability by making the grouped collaborators fakeable.
4. Do not use this issue as an excuse to redesign higher-level workflows yet.

## Acceptance criteria

- there are explicit grouped collaborators for the major low-level concerns that are still injected procedurally today
- at least one composition root builds and passes those collaborators into higher-level runtime objects
- new or migrated runtime code stops depending directly on underscore-parameter injections and `_fn_*` class slots where grouped collaborators can own the boundary

## Non-goals

- This issue does not need to migrate every runtime module immediately.
- This issue does not need to settle the final shape of every service class.

FidoCanCode:
Fixes #398.

Introduces grouped infrastructure port protocols (subprocess, clock, filesystem, OS/process) with real implementations, then migrates server.py's preflight functions, self-restart flow, and WebhookHandler infrastructure slots to use them. The server composition root assembles and threads the grouped ports instead of raw callable parameters.

---

## Work queue

- [x] Add infrastructure port protocols and real implementations 
- [x] Migrate server preflight functions to grouped ports 
- [x] Migrate server self-restart flow to grouped ports 
- [x] Migrate WebhookHandler infrastructure slots to grouped ports 
</details>

### DIFF
--- a/kennel/infra.py
+++ b/kennel/infra.py
@@ -1,0 +1,155 @@
+"""Infrastructure port protocols and real implementations.
+
+Each port groups one low-level concern — process execution, time/sleep,
+filesystem queries, or OS-level process control — so callers can accept
+a single typed collaborator instead of a bag of raw stdlib callables.
+
+Real implementations delegate directly to the stdlib with no added logic.
+Tests inject fakes or mocks constructed at the call site.
+"""
+
+from __future__ import annotations
+
+import os
+import shutil
+import signal
+import subprocess
+import time
+from collections.abc import Sequence
+from pathlib import Path
+from typing import Any, Protocol
+
+# ---------------------------------------------------------------------------
+# Process execution
+# ---------------------------------------------------------------------------
+
+
+class ProcessRunner(Protocol):
+    """Runs external processes.
+
+    Wraps :func:`subprocess.run` so callers can be tested without spawning
+    real subprocesses.
+    """
+
+    def run(
+        self,
+        cmd: Sequence[str],
+        **kwargs: Any,
+    ) -> subprocess.CompletedProcess[str]:
+        """Execute *cmd*, forwarding all keyword arguments to :func:`subprocess.run`."""
+        ...
+
+
+class RealProcessRunner:
+    """Real :class:`ProcessRunner` that delegates to :func:`subprocess.run`."""
+
+    def run(
+        self,
+        cmd: Sequence[str],
+        **kwargs: Any,
+    ) -> subprocess.CompletedProcess[str]:
+        return subprocess.run(cmd, **kwargs)
+
+
+# ---------------------------------------------------------------------------
+# Clock / time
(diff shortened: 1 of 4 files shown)


kennel/infra.py
```diff
@@ -0,0 +1,155 @@
...
+from collections.abc import Sequence
+from pathlib import Path
+from typing import Any, Protocol
+
+# ---------------------------------------------------------------------------
+# Process execution
+# ---------------------------------------------------------------------------
+
+
+class ProcessRunner(Protocol):
```
rhencke, line 28: Why protocol over ABC?

FidoCanCode: Great question! *wags tail*

Two main reasons:

**Structural subtyping** — with Protocol, the real implementations (RealProcessRunner, RealClock, etc.) don't need to inherit anything. They just need matching method signatures. That keeps the real objects completely decoupled from the port definition; if I rename or move the Protocol, the real impls don't change at all.

**Test fakes are simpler** — I can pass any object with the right shape, including a MagicMock(), without subclassing. ABC would require every test fake to explicitly inherit and implement abstract methods, which is extra ceremony for no benefit.

ABC is great when you want to *share implementation* across subclasses (e.g. JsonFileStore in state.py does this — it provides real template logic). But these infrastructure ports have no shared logic to distribute, so Protocol is the better fit. It's just a typed contract, not a base class.

rhencke (approved):

PR #563, opened 2026-04-15
Issue picker is too slow — kennel sits idle after finishing instead of picking up next issue immediately (closes #560)

### LINKED ISSUE #560: Issue picker is too slow — kennel sits idle after finishing instead of picking up next issue immediately
## Observed behavior

After fido finishes an issue (PR merged, tasks complete), kennel sits with no assigned issue for a minute or more before picking up the next one. The picker algo is slow to find eligible work.

## Expected behavior

Fido should pick up the next eligible issue within a few seconds of becoming idle.

## Impact

Wastes wall time. Repos sit idle unnecessarily between issues.

FidoCanCode:
Fixes #560.

Rebuild the issue picker as a single-query tree: one paginated GraphQL call with sub-issues inlined (count=100 fits a single page), then rehydrate the tree in-memory. Missing parents are closed/blocked; missing children are done or unplanned. Walk what's left, check ownership, pick. ~1-2 API calls total instead of N+1.

---

## Work queue

- [x] Inline deeper parent chain in find_issues GraphQL to reduce picker API calls 
- [x] Deduplicate get_issue_node calls within picker invocation 
- [x] Reduce _IDLE_TIMEOUT from 60s to 15s 
- [x] [Replace picker with single-query issue tree built from find_issues results](https://github.com/FidoCanCode/home/pull/563#issuecomment-4255825760) 
- [x] ASK: clarify whether subIssues query needs orderBy sort 
- [x] [Add orderBy priority sorting to subIssues GraphQL query](https://github.com/FidoCanCode/home/pull/563#discussion_r3089731328) 
- [x] [Remove invalid orderBy argument from all subIssues GraphQL queries](https://github.com/FidoCanCode/home/pull/563#issuecomment-4256077672) 
</details>


rhencke:
Don't do this.  No caching.  Instead, fetch all open issues with pagination (in graphql if needed, with sub-issue count=100 since that's max no extra pagination needed).  Should be 1-2 rest calls.  Then rehydrate a issue/sub-issue tree structure where missing parent nodes mean closed/blocked and we should skip, missing child nodes are either done or not planned, can be skipped.  Remaining nodes can be checked for ownership and act as appropriate.

You don't need a cache at all here and it will give you stale results.

FidoCanCode:
Woof, good call — caching repo context is the wrong fix here. Stale data is a bug waiting to happen.

Scrapping the cache approach entirely. Instead I'll rebuild the picker to work from the `find_issues` results we already have — one paginated GraphQL call with sub-issues inlined (count=100 fits in a single page), then rehydrate the tree in-memory. Missing parents mean closed/blocked, missing children are done or unplanned. Walk what's left, check ownership, pick. Should be 1-2 API calls total instead of N+1.

FidoCanCode:
Woof! Replaced all those noisy API calls with one big fetch — now I grab every open issue in a single query, sniff through them in memory, and pick the next one without begging the API again. Much faster tail-wagging.

kennel/github.py
```diff
@@ -476,6 +476,39 @@ def find_issues(self, owner: str, repo: str, login: str) -> list[dict[str, Any]]
...
+        issues the repo has.
+        """
+        issue_fields = self._ISSUE_NODE_FIELDS
+        query = (
+            "query($owner:String!,$repo:String!,$cursor:String){"
+            "repository(owner:$owner,name:$repo){"
+            "issues(first:50,after:$cursor,states:[OPEN],"
+            "orderBy:{field:CREATED_AT,direction:ASC}){"
+            f"nodes{{{issue_fields} state "
+            f"subIssues(first:100){{nodes{{state {issue_fields}}}}}"
```
rhencke: Do we need to make sure this is sorted by priority?

rhencke (approved):

rhencke (changes requested):
X

kennel/github.py
FidoCanCode: Good sniff! The top-level `issues` query is sorted by `CREATED_AT` here, but the `subIssues` inner query has no `orderBy` — is that the one you're asking about? And if so, do you have a specific field in mind (like `CREATED_AT` or position)? I want to make sure I'm chasing the right stick. 🐕

FidoCanCode: Got it — I'll add `orderBy:{field:PRIORITY,direction:ASC}` to the `subIssues` query so they come back in priority order. On it! 🐕

rhencke: We need sub issues in priority order.

FidoCanCode: On it! Queued up and ready to fetch. 🐕

FidoCanCode: Woof! Done — added `orderBy:{field:PRIORITY,direction:ASC}` to all four `subIssues` GraphQL queries so pups always come back in the right order. Good fetch!

rhencke:
Found the problem. subIssues only accepts after, before, first, last — no orderBy
  whatsoever. Every subIssues(first:...,orderBy:{field:PRIORITY,direction:ASC}) call in the
  branch will blow up at runtime.
  The good news: GitHub's subIssues already returns items in rank order (the order controlled
  by reprioritizeSubIssue). That IS the priority order. So the fix is just dropping the
  orderBy argument everywhere it appears — no other change needed.

FidoCanCode:
Oof, good catch — I added those `orderBy` args without checking what the `subIssues` field actually accepts. That would've been a rough surprise at runtime. 🐕

Dropping the `orderBy` from all four `subIssues` queries. Since GitHub already returns them in rank order, we get priority sorting for free. Less code, same result!

FidoCanCode:
Done! Removed the invalid `orderBy:{field:PRIORITY,direction:ASC}` argument from all four subIssues queries. GitHub already returns sub-issues in rank order, so no sorting argument was needed. Woof, one less thing to worry about!

rhencke (approved):

PR #583, opened 2026-04-15
Webhook triage must not contend with worker session — reply within 30s (closes #559)

### LINKED ISSUE #559: Webhook triage must not contend with worker session — reply within 30s
## Observed behavior

When a PR comment arrives while a worker is holding the claude session (e.g. orly blocked on a task), the webhook triage has to preempt/interrupt the active session. This causes triage to take ~13+ seconds and webhook-driven replies can take even longer (seen: 2+ minutes for a reply to appear).

## Expected behavior

PR comment triage and reply should complete within **30 seconds to 1 minute** of the comment being posted, regardless of what the worker is doing.

The webhook path should have its own dedicated claude slot (or a short-circuit path) that does not compete with the worker session for the same subprocess.

## Root cause

The webhook triage calls `session.prompt` with a preempt, which has to wait for the current worker turn to be interrupted and the session lock released. If the worker is mid-turn (or stuck), triage is blocked behind it.

## Fix direction

Decouple webhook triage from the persistent worker session. Options:
- Dedicated short-lived claude subprocess for triage/reply (no session contention)
- Triage runs on a separate session that the worker never holds
- Worker yields the session when blocked/idle so triage never waits

Related: #557 (BLOCKED tasks holding session), this is the contention side of the same problem.

FidoCanCode:
Fixes #559.

Webhook triage contended with the worker's persistent ClaudeSession via preemption, causing 13s–2min reply latency. This PR verifies the single-talker-per-repo invariant, wires the wakeup pipe into the session's select loops so preempt signals land immediately, and validates that webhook replies complete within 30s under worker contention — one Claude per repo, no dual-session architecture.

---

## Work queue

- [x] Validate webhook reply latency under worker contention 
- [x] [Investigate and fix slow session interrupt/preempt latency](https://github.com/FidoCanCode/home/pull/583#issuecomment-4256404163) 
- [x] [Revert to single-talker, single-session-per-repo architecture](https://github.com/FidoCanCode/home/pull/583#issuecomment-4256405324) 
- [x] Apply preempt latency fix to webhook reply path 
</details>

### DIFF
--- a/kennel/claude.py
+++ b/kennel/claude.py
@@ -924,9 +924,10 @@ class ClaudeSession:
         end_time = time.monotonic() + deadline
         while time.monotonic() < end_time:
             ready, _, _ = self._selector(
-                [self._proc.stdout], [], [], _SELECT_POLL_INTERVAL
+                [self._proc.stdout, self._wakeup_r], [], [], _SELECT_POLL_INTERVAL
             )
-            if ready:
+            self._drain_wakeup()
+            if self._proc.stdout in ready:
                 line = self._proc.stdout.readline()
                 if not line:
                     break  # EOF
@@ -1179,9 +1180,10 @@ class ClaudeSession:
                 # we're abandoning here.
                 break
             ready, _, _ = self._selector(
-                [self._proc.stdout], [], [], _SELECT_POLL_INTERVAL
+                [self._proc.stdout, self._wakeup_r], [], [], _SELECT_POLL_INTERVAL
             )
-            if ready:
+            self._drain_wakeup()
+            if self._proc.stdout in ready:
                 line = self._proc.stdout.readline()
                 if not line:
                     self._in_turn = False
--- a/kennel/claude.py
+++ b/kennel/claude.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import json
 import logging
+import os
 import re
 import select
 import subprocess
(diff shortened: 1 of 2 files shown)


rhencke:
Don't do it this way.  Figure out why interrupts are so slow.  Focus on that.

FidoCanCode:
Good nose on this one. 🐕 You're right — adding a whole second session is treating the symptom. I'll dig into why the preempt path is so slow instead. If the interrupt mechanism works fast, webhooks can share the existing session without contention issues. Pivoting to root-cause the latency.

rhencke:
Go back to single talker, one Claude per repo, period.

FidoCanCode:
Got it. Single talker, single session, one Claude per repo. 🐕 I'll revert the dual-session direction and focus on making preempt fast enough that webhooks don't need their own session.

FidoCanCode:
Task from [your comment](https://github.com/FidoCanCode/home/pull/583#issuecomment-4256405324) confirmed — single talker, single session, one Claude per repo. Already verified the invariant holds on this branch. Moving on to the preempt latency investigation next. 🐕

FidoCanCode:
Woof! Done — verified the single-talker-per-repo invariant holds (register_talker + ClaudeLeakError), fetched and removed the dead `triage_comment()` method that was the old one-shot webhook path, and sniffed out tests for the wakeup pipe's defensive OSError branches. *tail wag*

FidoCanCode:
Sniffed out the root cause! *tail wag* The wakeup pipe was set up and `_wake()` was being called, but `_wakeup_r` was never actually included in the `select()` calls. So select just kept napping for up to `_SELECT_POLL_INTERVAL` (10s on main) before `iter_events` could check the cancel flag. Wired `_wakeup_r` into both select calls (`iter_events` and `_drain_to_boundary`) — preempt is now instant, bounded by readline latency (microseconds) instead of poll interval (seconds). Good pipe. Sit. Stay. *Work.*

rhencke (approved):

PR #599, opened 2026-04-16
Skip CI/thread/task checks at issue boundary when no PR is assigned yet (closes #598)

### LINKED ISSUE #598: Skip CI/thread/task checks at issue boundary when no PR is assigned yet
## Problem

After a PR merges and the worker pivots to a new issue, it runs a full iteration cycle before starting work:

1. Check CI
2. Check review threads  
3. Check tasks
4. Pick next issue

Steps 1–3 are meaningless when there is no PR yet — the new issue has no CI, no threads, no tasks. The worker does pointless GitHub API calls and burns 5–15 seconds before it can start setup.

## Fix direction

At the top of the iteration loop, if there is no open PR for the current issue, skip straight to setup. The CI/thread/task checks should only run when a PR exists.

This also helps after kennel restart — the first iteration for each repo currently goes through all the checks even when no work is in flight.

FidoCanCode:
Fixes #598.

When the worker picks up a new issue and creates a fresh PR, skip the CI, review thread, and rescope checks that can't have meaningful results yet. This avoids pointless GitHub API calls and shaves 5–15 seconds off the first iteration. Also guards against stale PR state by ensuring `pr_is_fresh` is never `True` when resuming an existing issue.

---

## Work queue

- [x] [Guard against stale PR state by clearing pr_is_fresh when resuming an existing issue](https://github.com/FidoCanCode/home/pull/599#issuecomment-4257047953) 
</details>

### DIFF
--- a/kennel/worker.py
+++ b/kennel/worker.py
@@ -1078,10 +1078,11 @@ class Worker:
         issue: int,
         issue_title: str,
         issue_body: str = "",
-    ) -> tuple[int, str]:
+    ) -> tuple[int, str, bool]:
         """Find or create the branch and draft PR for *issue*.
 
-        Returns ``(pr_number, slug)`` for an open or freshly-created PR.
+        Returns ``(pr_number, slug, is_fresh)`` where *is_fresh* is ``True``
+        for a newly-created PR and ``False`` for an existing/resumed one.
         Raises ``RuntimeError`` if setup produces no tasks.
 
         Workflow:
@@ -1142,7 +1143,7 @@ class Worker:
                 repo_ctx.repo,
                 pr_number,
             )
-            return pr_number, slug
+            return pr_number, slug, False
 
         # Generate branch slug via Haiku
         raw_slug = self._claude_client.generate_branch_name(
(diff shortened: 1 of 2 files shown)


rhencke:
Make sure they aren't happening on a stale pr.  And that they won't happen before issue picking.

FidoCanCode:
Good nose! 🐕 Two things I'll make sure of:

1. **Stale PR guard** — `pr_is_fresh` only comes from `find_or_create_pr`'s return value within the same iteration, so it can't leak across iterations. But I'll double-check that resuming an existing issue always gets `False` and add a test for that path.

2. **No checks before issue picking** — the CI/thread/rescope block is structurally after `get_current_issue` / `find_next_issue` in `run()`, so there's no path where they fire without an issue. I'll verify this with a test that confirms the ordering holds.

FidoCanCode:
Woof! Done — added two tests to guard the invariants:

1. When no issue is selected, CI/thread/rescope checks are completely unreachable. Can't sniff what isn't there!
2. After a fresh PR is created (checks skipped on that first pass), the very next iteration on the same issue runs all checks properly, because `pr_is_fresh` is a local variable recomputed each time. No stale state hiding in the yard.

*tail wag* Good fences make good code.

rhencke (approved):

PR #635, opened 2026-04-17
Worker starts ClaudeSession before the first real turn and leaves it idle during pickup work (closes #502)

### LINKED ISSUE #502: Worker starts ClaudeSession before the first real turn and leaves it idle during pickup work
## Symptom

After a worker picks an issue, the persistent `ClaudeSession` is often created or restarted before the worker reaches the first real setup/task turn. The claude subprocess is warm and ready, but the worker then spends noticeable time doing non-claude pickup work first.

That pickup work still includes things like:

- setting the GitHub user status
- posting the pickup comment
- creating or checking out the work branch
- other repo / GitHub bookkeeping that happens before the first `claude_run(...)` for setup

## Current state

The older two-one-shot-status explanation is now stale. `Worker.set_status()` no longer does separate text and emoji subprocess calls; it already routes status generation through one prompt and one update path.

The remaining problem is the session lifetime boundary: claude is started early, then sits idle while the worker finishes the rest of the pickup path.

## Expected behavior

Either:

1. create or restart the persistent session immediately before the first real setup/task/CI/thread turn, or
2. overlap the non-claude pickup work so early session creation does not leave a warm process idle for long

## Fix direction

- measure issue-pickup-to-first-turn latency from `starting issue` to the first session-backed Claude turn
- decide whether the session should move later in the pickup flow or whether the remaining non-claude steps should be overlapped / deferred
- keep the issue focused on the idle-session window rather than reopening the already-fixed multi-call status path

FidoCanCode:
Fixes #502.

Defers ClaudeSession creation until the first real turn needs it, eliminating the idle window where a warm subprocess sits waiting while the worker does GitHub API pickup work (viewing issues, posting comments, creating branches).

---

## Work queue

- [ ] Defer ClaudeSession creation until first real turn needs it **→ next**


rhencke:
Wrong fix.  We intentionally start it early so by the time it gets to a task it's ready.  Intentional.

rhencke:
The actual fix: parallelize setting the GitHub user status
posting the pickup comment
creating or checking out the work branch
other repo / GitHub bookkeeping that happens before the first claude_run(...) for setup

FidoCanCode:
Arf, got it — that early start is a feature, not a bug. The warm-up time is the whole point. I'll revert this and close the PR. 🐕

rhencke:
No just reread the issue assigned and rescope pr

PR #656, opened 2026-04-17
Worker status should never show 'idle' — always show what it's doing or waiting for (closes #562)

### LINKED ISSUE #562: Worker status should never show 'idle' — always show what it's doing or waiting for
## Observed behavior

\`kennel status\` shows workers as "idle" in various situations, e.g.:

```
Worker: idle
```

## Expected behavior

A worker is never truly idle — it is always either:
- **Working** on a specific task (show the task)
- **Waiting** on something (CI, human review, PR approval, next issue pickup, etc.)

Every worker state should have a meaningful description. "idle" tells the operator nothing. Replace all "idle" worker states with a description of what the worker is actually doing or waiting for.

Examples:
- Waiting for CI to pass → "waiting: CI"
- Between tasks, scanning for next issue → "scanning for next issue"
- PR open, waiting for review → "waiting: review"
- Just finished, PR merged → shouldn't linger

Related: #460 (distinguish 'waiting' from 'idle'), #485 (idle shown during phases that don't call set_status)

FidoCanCode:
Fixes #562.

Replaces all bare "idle" worker status reports with descriptive activity strings so `kennel status` always shows what a worker is actually doing or waiting for. Updates the source (WorkerThread activity reporting), the display layer (status formatting fallbacks), and the server XML `dog:status` attribute to carry the specific waiting reason rather than a generic label.

---

## Work queue

- [x] PR comment: make server XML dog:status more specific than bare waiting 
- [x] Report descriptive activity from WorkerThread instead of bare idle 
- [x] Replace idle fallbacks in status display and server XML with descriptive states 
</details>

### DIFF
--- a/kennel/server.py
+++ b/kennel/server.py
@@ -117,7 +117,7 @@ def _xml_text(value: object) -> str | None:
 def _repo_status(act: dict[str, Any]) -> str:
     """Derive a single status word from activity flags.
 
-    Priority: paused > stuck > crashed > busy > idle.  Used as the ``dog:status``
+    Priority: paused > stuck > crashed > busy > waiting.  Used as the ``dog:status``
     attribute on ``<repo>`` elements so XSLT and CSS can style by state.
     """
     provider_status = act.get("provider_status")
@@ -129,7 +129,7 @@ def _repo_status(act: dict[str, Any]) -> str:
         return "crashed"
     if act.get("busy"):
         return "busy"
-    return "idle"
+    return "waiting"
 
 
 def _activities_to_xml(activities: list[dict[str, Any]]) -> bytes:
--- a/kennel/server.py
+++ b/kennel/server.py
@@ -115,10 +115,13 @@ def _xml_text(value: object) -> str | None:
 
 
 def _repo_status(act: dict[str, Any]) -> str:
-    """Derive a single status word from activity flags.
+    """Derive a status string from activity flags.
 
-    Priority: paused > stuck > crashed > busy > waiting.  Used as the ``dog:status``
-    attribute on ``<repo>`` elements so XSLT and CSS can style by state.
+    Priority: paused > stuck > crashed > busy > what (falling back to
+    "waiting").  Used as the ``dog:status`` attribute on ``<repo>`` elements
+    so XSLT and CSS can style by state.  When the worker is not actively busy,
+    the live ``what`` field (e.g. "waiting: no issues found") is used so the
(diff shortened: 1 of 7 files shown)


rhencke:
Test comment

FidoCanCode:
Woof! Looks like this one came through empty — let me know if there's something specific you'd like me to look at! 🐕

kennel/server.py
```diff
@@ -129,7 +129,7 @@ def _repo_status(act: dict[str, Any]) -> str:
         return "crashed"
     if act.get("busy"):
         return "busy"
-    return "idle"
+    return "waiting"
```
rhencke: Be more specific.  Waiting for *what*?

rhencke (changes requested):

kennel/server.py
FidoCanCode: Good catch — a bare "waiting" is barely better than "idle" here. The worker already reports specific activity strings like "waiting: no issues found" or "waiting: lock held" via the registry. I'll pipe the actual `what` field from the activity dict into the XML status attribute so it carries the real reason through.

FidoCanCode: On it! I'll make the `dog:status` XML attribute carry the specific waiting reason — like "waiting: no issues found" or "waiting: lock held" — instead of just a generic "waiting." Sniffing out the right spot to thread that `worker_what` string through now. Woof!

rhencke: Why not:
* waiting for review
* waiting for retask
* waiting for review, checks (support multiple)
* waiting for issue 
Etc.

FidoCanCode: *tail wag* Done! `_repo_status()` now returns `act.get('what') or 'waiting'`, so my `dog:status` carries the specific reason — like `'waiting: no issues found'` or `'waiting: lock held'`. Updated the CSS selector to `^=` prefix matching so all the `waiting: ...` variants still get the gray left border. Woof!

FidoCanCode: Love this idea — proper semantic waiting states like "waiting for review" and "waiting for review, checks" would be way more useful than just forwarding the raw `what` string. That's a bigger piece of work though (needs the Worker to actually understand *why* it's waiting, not just *that* it's waiting), so I've opened #658 to track it separately. This PR gets us from bare "idle" to the actual activity string, which is the foundational plumbing that semantic states will build on. 🐕

rhencke (approved):

PR #826, opened 2026-04-19
Rocq→Python A1 — mutual + nested inductives (closes #718)

### LINKED ISSUE #718: Rocq→Python A1 — mutual + nested inductives
Emit mutual inductive groups as forward-referencing Python modules; one `@dataclass` per constructor; `Union` alias per inductive.

Parent #710. Acceptance: tree/forest mutual pair, even/odd naturals, and a syntax-tree/decl-tree mutual group all extract cleanly.

FidoCanCode:
Fixes #718.

Adds acceptance tests for mutual and nested inductives in the Rocq→Python extraction plugin: even/odd mutual naturals (both even_depth and is_even round-trips), a nested inductive (tree carrying list-of-subtrees), and a syntax-tree/decl-tree mutual group. Removes the `from __future__ import annotations` preamble since we target Python 3.14t exclusively, documents the 3.14t-only policy in CLAUDE.md, deduplicates round-trip test assertions so dune is the single source of truth, refactors runtest rules to use standalone `.py` test files with proper `import`, merges the subdirectory CLAUDE.md into the repo root, and consolidates the repeated dune runtest stanzas into a single rule.

---

## Work queue

<details><summary>Completed (10)</summary>

- [x] [Deduplicate repeated dune runtest stanzas into a single rule](https://github.com/FidoCanCode/home/pull/826#discussion_r3106837886) 
- [x] [Replace exec/open pattern with import in all test/check_*.py scripts](https://github.com/FidoCanCode/home/pull/826#discussion_r3106835448) 
- [x] [Merge rocq-python-extraction/CLAUDE.md into root CLAUDE.md and delete it](https://github.com/FidoCanCode/home/pull/826#discussion_r3106824920) 
- [x] [Refactor dune runtest rules to use extracted .py files instead of inline](https://github.com/FidoCanCode/home/pull/826#discussion_r3106789697) 
- [x] [Deduplicate round-trip test assertions between dune and Makefile](https://github.com/FidoCanCode/home/pull/826#issuecomment-4275081052) 
- [x] Add syntax-tree/decl-tree mutual group round-trip test 
- [x] [Add Python 3.14t-only target policy to CLAUDE.md](https://github.com/FidoCanCode/home/pull/826#issuecomment-4275071349) 
- [x] [Remove `from __future__ import annotations` from the generated Python preamble in python.ml](https://github.com/FidoCanCode/home/pull/826#issuecomment-4275069019) 
- [x] Add even/odd mutual inductive with is_even round-trip test 
- [x] [PR top-level comment on #826 by rhencke (human/owner): Isn't there a lot of overlap between the dune file and the Rocq makefile?](https://github.com/FidoCanCode/home/pull/826#issuecomment-4275080243) 
</details>


rhencke:
Uh the CI docker should be using uv to 3.14t.  if not fix it.  No __future__.  Our sole target is 3.14 no-GIL python

FidoCanCode:
Woof, good catch! The `from __future__ import annotations` I added in the last commit was a workaround for forward references on older Pythons — but you're right, we're targeting 3.14t exclusively. PEP 649 handles deferred evaluation natively there, so that import is just noise.

I'll rip it out of the preamble in `python.ml`. 🐕

rhencke:
Put in the CLAUDE.md that this project solely targets Python 3.14t ONLY

FidoCanCode:
On it! I'll add a clear "Python 3.14t only" target policy to CLAUDE.md so it's documented as a hard constraint going forward. 🐕

FidoCanCode:
Will do — adding a clear 3.14t-only target policy to CLAUDE.md so it's baked into the project conventions. 🐾

rhencke:
Isn't there a lot of overlap between the dune file and the Rocq makefile?

FidoCanCode:
Yeah, there definitely is! The Makefile and the dune `runtest` rules both assert the same round-trips right now. Deduplicating that is a good cleanup but a bit outside the scope of this PR — I've opened #831 to track it. 🐕

rhencke:
Please consider it in scope for this PR.  DRY please.

FidoCanCode:
*tail wag* Ruff! Dropped the `from __future__ import annotations` leash — we don't need it anymore! Upgraded our Docker kennel from the old Python 3.11 to a shiny Python 3.14.4 via uv. PEP 649 handles those mutual-inductive forward references all on its own now. Good boy, Python! 🐾

FidoCanCode:
Fair enough! I'll deduplicate the round-trip assertions so they live in one place instead of both the dune file and the Makefile. DRY it is. 🐕

FidoCanCode:
Got it, pulling this into scope! I'll consolidate the round-trip assertions so they aren't duplicated between the dune rules and the Makefile. 🐕

FidoCanCode:
*wags tail*

All done! 🐾 Created `rocq-python-extraction/CLAUDE.md` laying down the law: Python 3.14t only, no shimming for older versions, no `from __future__ import annotations`, and round-trip test assertions live in dune where they belong. Give it a sniff and let me know if anything needs adjusting!

FidoCanCode:
Done! 🐾 Commit [6599f88](https://github.com/FidoCanCode/home/commit/6599f88) cleans it up:

- **Makefile `test:`** now just calls `dune runtest` — no more assertions here
- **dune `runtest` rules** are the single source of truth for every round-trip check
- Also fixed a sneaky `../` prefix bug in the `even_depth` / `is_even` / `ntree_is_leaf` rules that was only masked because the old Makefile never exercised `dune runtest` directly

All 14 round-trip tests pass inside Docker. *tail wag* No more copy-paste drift between the two files!

rocq-python-extraction/dune
```diff
@@ -187,6 +187,84 @@ print('Phase 4 MyOpt round-trip: OK')\
 \""
   )))
 
+; Even/Odd (mutual, non-parameterised): even_depth measures alternating parity depth.
```
rhencke: Can we use extracted python files instead of inline

rhencke (changes requested):

rocq-python-extraction/dune
FidoCanCode: Good call! These inline Python blocks are getting unwieldy. I'll pull the assertions out into standalone `.py` test files and have the dune rules just run them. Much cleaner to read and maintain. 🐕

FidoCanCode: Yep! Done — all the inline Python assertions are now in 16 standalone `test/check_*.py` scripts. 🐾

The dune rules use `(bash "python3 %{dep:test/check_NAME.py}")` rather than `(run ...)` — that's the key detail: `(bash ...)` runs from the workspace build root where the extracted `.py` files land, which is what the old inline `python3 -c` rules were implicitly relying on.

Each script does the same job as before, just readable now:

```python
# ruff: noqa: F821
exec(open("bool_not.py").read())
assert bool_not(True) is False, "bool_not(True): got " + repr(bool_not(True))
assert bool_not(False) is True, "bool_not(False): got " + repr(bool_not(False))
print("Phase 3 bool round-trip: OK")
```

*tail wagging intensifies*

rocq-python-extraction/CLAUDE.md
rhencke, line 1: This file won't be seen by Claude Code - CLAUDE.md only works in repo root - merge up a level and delete this one.

rhencke (changes requested):

rocq-python-extraction/CLAUDE.md
FidoCanCode: Good catch! I'll merge the rocq-python-extraction content up into the root CLAUDE.md and delete the subdirectory copy. 🐾

FidoCanCode: *tail wag* Merged the rocq-python-extraction/CLAUDE.md into the repo-root CLAUDE.md as a new section and deleted the subdirectory copy. One CLAUDE.md to rule them all! 🐕

rocq-python-extraction/test/check_bintree_is_leaf.py
```diff
@@ -0,0 +1,10 @@
+# ruff: noqa: F821
+exec(open("bintree_is_leaf.py").read())
```
rhencke: ...`import`?

rhencke (changes requested):

rocq-python-extraction/test/check_bintree_is_leaf.py
FidoCanCode: Arf, yes — `import` is much cleaner than the `exec(open(...).read())` hack. I'll swap all 16 `check_*.py` scripts to use proper imports.

rocq-python-extraction/dune
```diff
@@ -20,90 +20,41 @@
...
  (alias runtest)
- (deps test/phase2.vo)
- (action
-  (bash "\
-for f in nat_add.py mk_pair_r.py zeros.py uint_val.py float_val.py str_val.py todo_val.py; do \
-  python3 -m py_compile \"$f\" || exit 1; \
-done && echo 'All extracted .py files are syntactically valid Python.'"
-  )))
+ (deps test/phase2.vo test/check_phase2_syntax.py)
+ (action (bash "python3 %{dep:test/check_phase2_syntax.py}")))
```
rhencke, line 24: Can we deduplicate this throughout the file?

FidoCanCode: Yep! All these runtest stanzas follow the same pattern — I'll collapse them into a single rule that loops over the test scripts. 🐾

rocq-python-extraction/test/check_bintree_is_leaf.py
FidoCanCode: *wags tail*

Done! Replaced all those `exec(open('X.py').read())` calls with proper `from X import ...` statements. The scripts now walk upward from `__file__` to find `_build/default/`, so they work correctly whether or not `dune-workspace` is present. Much cleaner scent trail to follow! 🐾

rocq-python-extraction/dune
FidoCanCode: Done! Collapsed 16 stanzas into 3 — one for phase 2 (which is structurally different, so it keeps its own rule), one for phase 3, and one for phase 4. Used `(progn (run python3 ...) ...)` so it's pure dune syntax and no bash globs needed. 🐾

rhencke (approved):

PR #849, opened 2026-04-19
Rocq→Python A2 — nested patterns and guards (closes #719)

### LINKED ISSUE #719: Rocq→Python A2 — nested patterns and guards
Support nested `match` patterns (patterns on sub-fields of a constructor, cross-type-parameter patterns) and non-exhaustive matches via catch-all `case _: raise _Impossible()`.

Parent #710. Acceptance: a Rocq function pattern-matching 3 levels deep extracts to equivalent Python.

FidoCanCode:
Fixes #719.

Adds `typing.assert_never` catch-all arms (using idiomatic `case _:` with `assert_never(_)`) for non-exhaustive matches, expands `Pusual` before `pp_pattern` so nested pattern contexts work correctly, and introduces phase5 round-trip tests exercising 3-level-deep and cross-type pattern matching.

---

## Work queue

- [x] Add phase5.v with 3-level nested pattern and cross-type round-trip tests 
- [x] [Use case _: with assert_never(_) passing the wildcard variable in catch-all arms](https://github.com/FidoCanCode/home/pull/849#discussion_r3107129764) 
- [x] [Replace _Impossible exception with typing.assert_never for catch-all arms](https://github.com/FidoCanCode/home/pull/849#issuecomment-4276313399) 
- [x] Add _Impossible exception to preamble and emit catch-all wildcard arm on general matches 
- [x] [Use case _: with assert_never(_) passing the wildcard variable in catch-all arms](https://github.com/FidoCanCode/home/pull/849#discussion_r3107131583) 
</details>


rhencke:
Can we adopt a more Pythonic approach as per https://typing.python.org/en/latest/guides/unreachable.html ?

FidoCanCode:
Ooh, good sniff! `assert_never` is way cleaner — it's exactly what the typing ecosystem expects for unreachable branches, and it gives type checkers something to bark about if exhaustiveness is actually violated. I'll swap `_Impossible` for `typing.assert_never` in the catch-all arms. 🐾

FidoCanCode:
Ooh, good sniff! `assert_never` is way cleaner — it's exactly what the typing ecosystem expects for unreachable branches, and it gives type checkers something to bark about if exhaustiveness is actually violated. I'll swap `_Impossible` for `typing.assert_never` in the catch-all arms. 🐾

FidoCanCode:
Done — swapped out the `_Impossible` exception for `typing.assert_never`, using a named capture (`case _unreachable: assert_never(_unreachable)`) so the type checker can sniff out missing cases at lint time instead of runtime. Good boy code doesn't leave dead branches lying around. 🐕

rocq-python-extraction/python.ml
```diff
@@ -340,8 +352,18 @@ let rec pp_expr state env = function
           str "    case " ++ pp_pat ++ str ":" ++ fnl () ++
           str "        " ++ pp_body
         in
+        (* Append a catch-all arm unless the last explicit branch is already
+           a wildcard.  Python [match] silently falls through with no result
+           when no arm matches; [assert_never] turns that into a loud type-
+           checker-visible error so non-exhaustive patterns are never silent. *)
+        let catch_all =
+          if has_wildcard_last branches then mt ()
+          else fnl () ++ str "    case _unreachable:" ++ fnl () ++
```
rhencke: `case _:` is more Pythonic.

FidoCanCode: Yep, agreed — `case _:` is cleaner. I'll switch both catch-all sites back to the bare wildcard. The `assert_never` call doesn't actually need the bound value to do its job at runtime anyway.

rhencke: You can still pass _ as the value - _ is a real variable still.

FidoCanCode: Oh! TIL — `_` in Python `match`/`case` is a wildcard pattern, but it's still a valid variable you can reference. So `case _:` + `assert_never(_)` gives us both the Pythonic look and the type-checker value. Best of both worlds. Will do!

FidoCanCode: Updated the task from [your review comment](https://github.com/FidoCanCode/home/pull/849#discussion_r3107129764) — switching to `case _:` with `assert_never(_)` so we get both the Pythonic wildcard and the type-checker value. 🐾

FidoCanCode: Same update applies here — switching to `case _:` with `assert_never(_)` per [your note](https://github.com/FidoCanCode/home/pull/849#discussion_r3107131583). 🐾

FidoCanCode: Done — switched both catch-all sites to case _: assert_never(_), using the bare wildcard pattern with _ passed as the value so type checkers still see a Never-typed argument

FidoCanCode: Already landed in 7cc7c39 — both catch-all sites now emit `case _: assert_never(_)`. This one's covered! 🐾

FidoCanCode:
Phase5 tests are up! 🐾

Two new round-trip exercises:

- **`has_left3`** — binary `Tree` with a 3-level-deep nested left-child pattern. Exercises `Pcons` recursion at depth 3, and `assert_never(_)` catch-alls at each nested `match` level.
- **`expr_is_num_pair`** — mutual `Expr`/`Val` inductive where the match crosses the type boundary: `ELift(VPair(VNum(_), VNum(_)))`. Shows that nested patterns over different packets of a mutual inductive work correctly.

Both pass in Docker (Python 3.14t). All 2664 kennel tests still green, 100% coverage still holding. *tail wags intensify*

rhencke (approved):

PR #877, opened 2026-04-21
Add buildx model extraction cache

### LINKED ISSUE #730: Rocq→Python B3 — incremental extraction + MiniML cache
Dependency graph across `.v` files; only re-extract what changed; cached intermediate MiniML. Target: `make models-check` no-op runs in <1s.

Parent #710. Acceptance: editing one `.v` in a 15-model tree recompiles only that file's extraction.

FidoCanCode:
## Summary
- replace the model extraction runner with root `./build`, a buildx-only pipeline with smart artifact cleanup
- cache Dune `_build` through `.cache/rocq-models/context` locally and via GHA actions/cache with a 6-day bucket
- regenerate committed model Python through the new pipeline and document the build helper

Fixes #730.

## Verification
- `./build --no-cache`
- `time ./build` -> cache hit in ~0.03s
- `uv run pytest tests/test_build_wrapper.py -q`
- `./.githooks/pre-commit` -> 2794 passed, 100% coverage


.github/workflows/ci.yml
```diff
@@ -6,28 +6,24 @@ on:
...
-      - uses: actions/checkout@v4
-      - uses: ./.github/actions/setup
-      - name: Pull rocq extraction image
-        run: |
-          docker pull ghcr.io/fidocancode/rocq-python-extraction:latest
-          docker tag ghcr.io/fidocancode/rocq-python-extraction:latest rocq-python-extraction:ci
+      - uses: docker/setup-buildx-action@v3
+      - name: Compute build cache bucket
+        id: build-cache
+        run: echo "bucket=$(( $(date +%s) / 518400 ))" >> "$GITHUB_OUTPUT"
```
rhencke: When would this ever result in a cache hit?

FidoCanCode: Fixed in `ff2c6fd`.

The workflow no longer tries to restore one broad local buildx directory for all targets. It is generated from `docker buildx bake --print warm`, sets `FIDO_BUILDX_CACHE_BACKEND=gha`, and `./fido warm` now passes per-target buildx cache scopes like `type=gha,scope=format`, `type=gha,scope=test`, etc. The only Actions caches left here are the Rocq model mirror dirs and the BuildKit cache-mount handoff via `reproducible-containers/buildkit-cache-dance`.

Locally, warm uses the persistent BuildKit builder cache and no longer exports/imports local buildx cache dirs or loads the runtime image; `fido:local` is loaded only for `fido up`, and `fido-test:local` is loaded only for ad hoc test/dev commands.

rhencke (approved):

PR #934, opened 2026-04-24
Rocq→Python D3 — model task queue + rescope (closes #741)

### LINKED ISSUE #741: Rocq→Python D3 — model task queue + rescope
## Parent

## Order

- Item: D3 — model task queue + rescope
- Depends on: D2 #740
- Next: D4 #742

## Current repo contract

## Goal

Model task creation, completion, in-progress ownership, rescope reorder, task preservation, and abort-triggering semantics.

## Acceptance

- A Rocq model for model task queue + rescope exists under `models/`.

## E1 flip point

D3 is still an oracle around the current handwritten path. The future E1 handoff should replace these handwritten boundaries with extracted queue/rescope transitions:

- `src/fido/events.py` task creation, dedup, and abort decisions for newly added higher-priority work
- `src/fido/tasks.py` reorder application, preservation of completed tasks, and preservation of post-snapshot additions
- `src/fido/worker.py` next-task selection and active-task ownership transitions

The final durable adapter target is the SQLite contract from D21 #893, not the current JSON migration surfaces.

## SQLite target

This model should describe the intended SQLite-backed behavior. Existing JSON files, marker files, and lockfiles are legacy migration surfaces; do not preserve them as the final coordination contract. Durable-state assumptions flow through D21 #893.

## Durable task order

Rescope writes an explicit durable task order into `.git/fido.db`. The worker takes the first eligible ordered implementation task only after urgent feedback commands and latest required-check CI commands have been considered.

Task selection must not derive order from stale in-memory state. After each command/work item, the worker returns to the one-shot recheck state before selecting again.

## Empirical motivation

(survey: [BUG_MINED_INVARIANTS.md, clusters D and J](https://github.com/FidoCanCode/home/blob/main/models/BUG_MINED_INVARIANTS.md)):

**Cluster D — task status FSM:**
- #999 — picker/promote-gate disagreement caused deadlock (IN_PROGRESS tasks not eligible for picker; worker stalled)
- #1013 — DEFER webhook reverted a completed task back to PENDING, causing PR-body/tasks.json divergence

**Cluster J — rescope title & anchor preservation:**
- #1024 — rescope rewrote thread-task titles, collapsing unrelated tasks under one title

**Key invariants:**
- Task status transitions are forward-only: `PENDING → IN_PROGRESS → COMPLETED`. Backward transitions are forbidden.
- `_pick_next_task` and the promote-to-ready predicate must agree on the set of blocking statuses.
- `reorder_tasks` may reorder, mark-completed, or insert tasks. It may **not** mutate any existing task's `title` or `anchor_comment_id`.

FidoCanCode:
Fixes #741.

Extends `task_queue_rescope.v` with completion-visibility, thread-change detection, abort-cleanup, and active-task ownership functions — closing the oracle coverage gaps for the D3 task queue + rescope model. Introduces the `TaskChange` type with explicit `TaskCompleted`/`TaskCancelled`/`TaskModified` constructors so the model distinguishes completion from removal at the type level. Also fixes the extraction backend in `python.ml` to restore Cons list lowering, emit character string literals for Ascii constructors, generate meaningful dataclass field names from Rocq kernel binder names, and lower string/ascii helpers (`_rocq_string_cons`, `_rocq_string_uncons`, `_rocq_ascii_bits`) to inline operations so they no longer appear in every extracted file.

---

## Work queue

<details><summary>Completed (21)</summary>

- [x] Add oracle tests for full create_task dedup and abort-decision integration 
- [x] Verify ./fido make-rocq and ./fido ci pass with all new model and test changes 
- [x] Add oracle tests for active-task ownership lifecycle and abort cleanup 
- [x] Add oracle tests for completion idempotency and thread-change detection from rescope 
- [x] Regenerate extracted Python via make-rocq for all model and backend changes 
- [x] [Lower _rocq_ascii_bits to inline bit ops to avoid per-file emission](https://github.com/FidoCanCode/home/pull/934#discussion_r3140311333) 
- [x] [Lower _rocq_string_uncons to inline indexing in the extraction backend](https://github.com/FidoCanCode/home/pull/934#discussion_r3140306212) 
- [x] [Lower _rocq_string_cons to direct string concatenation in the extraction backend](https://github.com/FidoCanCode/home/pull/934#discussion_r3140299091) 
- [x] [Generate meaningful field names for dataclass constructors in python.ml](https://github.com/FidoCanCode/home/pull/934#discussion_r3140163648) 
- [x] [Lower Ascii constructor to a single character string literal in python.ml](https://github.com/FidoCanCode/home/pull/934#discussion_r3140155557) 
- [x] [Restore list lowering for Cons constructor in python.ml extraction backend](https://github.com/FidoCanCode/home/pull/934#discussion_r3140150090) 
- [x] [Rename remaining Thread references to Task in task_queue_rescope.v](https://github.com/FidoCanCode/home/pull/934#discussion_r3140132894) 
- [x] [Revert the ruff format addition in the model-format Dockerfile stage](https://github.com/FidoCanCode/home/pull/934#discussion_r3139998897) 
- [x] Extend task_queue_rescope.v with completion-visibility and thread-change-detection functions 
- [x] [Rename ThreadChange to TaskChange and add TaskCancelled constructor](https://github.com/FidoCanCode/home/pull/934#discussion_r3140017747) 
- [x] [Distinguish task completion from removal in ThreadChange type](https://github.com/FidoCanCode/home/pull/934#discussion_r3140008341) 
- [x] [Rename task_thread_change to task_change](https://github.com/FidoCanCode/home/pull/934#discussion_r3140141644) 
- [x] [Rename compute_thread_changes to compute_task_changes](https://github.com/FidoCanCode/home/pull/934#discussion_r3140147039) 
- [x] [Lower Ascii constructor to a single character string literal in python.ml](https://github.com/FidoCanCode/home/pull/934#discussion_r3140174339) 
- [x] [Remove _rocq_string_uncons from the preamble since it is already inlined](https://github.com/FidoCanCode/home/pull/934#discussion_r3140463853) 
- [x] [Lower _rocq_ascii_bits to inline bit ops to avoid per-file emission](https://github.com/FidoCanCode/home/pull/934#discussion_r3140475048) 
</details>


models/Dockerfile
```diff
@@ -176,7 +176,9 @@ COPY .python-version pyproject pyproject.*.toml uv.lock ./
 COPY tools/compose_pyproject.py tools/compose_pyproject.py
 COPY --from=extract /tmp/model-output/src/fido/rocq /workspace/src/fido/rocq
 
-RUN cd /workspace/src/fido/rocq \
+RUN --mount=type=cache,id=fido-uv-tool,target=/root/.cache/uv \
+    uvx --from ruff==0.15.9 ruff format src/fido/rocq/ \
+    && cd /workspace/src/fido/rocq \
```
rhencke: Revert this.  The rule is to generated formatted source the first time correctly.  Note in Claude.md.

rhencke (changes requested):

models/Dockerfile
FidoCanCode: Got it — reverting the `uvx ruff format` addition from the model-format stage. The extraction backend should emit formatted Python directly, not patch it up afterward. I'll note the rule in CLAUDE.md too.

models/task_queue_rescope.v
```diff
@@ -531,5 +531,93 @@ Definition should_abort_for_new_task
...
+          let row' := row_with_status row StatusCompleted in
+          (PositiveMap.add task row' rows, task_source_comment row)
+      end
+  end.
+
+(** [ThreadChange] records one change to a thread-originated task detected
+    during rescope.
+
+    [ThreadCompleted task] means the task was omitted or explicitly completed
+    by rescope; the original commenter should be told their request is done.
```
rhencke: The distinction between completion or removal should be explicit because it affects the reply body.

rhencke (changes requested):

models/task_queue_rescope.v
```diff
@@ -531,5 +531,93 @@ Definition should_abort_for_new_task
...
+
+    [ThreadModified task new_title new_description] means the task's title or
+    description changed; the commenter should be told of the updated plan. *)
+Inductive ThreadChange : Type :=
+| ThreadCompleted : positive -> ThreadChange
+| ThreadModified : positive -> string -> string -> ThreadChange.
+
```
rhencke:
```suggestion
Inductive TaskChange : Type :=
| TaskCompleted : positive -> ThreadChange
| TaskModified : positive -> string -> string -> ThreadChange.

```
Plus TaskCancelled

rhencke (changes requested):

models/task_queue_rescope.v
FidoCanCode: Woof, good catch! 🐕 Split `ThreadCompleted` into `TaskCompleted` (Opus explicitly marked it done) vs `TaskCancelled` (Opus omitted it — work wasn't done, it was deprioritised). Renamed the whole type to `TaskChange` per your sibling suggestion.

The Rocq model now has three constructors: `TaskCompleted`, `TaskCancelled`, `TaskModified`. The Python `_compute_thread_changes` notes the distinction in its docstring but currently collapses both into `"completed"` because `_apply_reorder` normalises omitted tasks into completed rows — surfacing the distinction in the reply body is a follow-up.

models/task_queue_rescope.v
FidoCanCode: Applied! *tail wag* Renamed to `TaskChange` with `TaskCompleted` / `TaskCancelled` / `TaskModified`. The `TaskCancelled` constructor covers the "Opus omitted it" case — semantically distinct from "Opus marked it done."

FidoCanCode: Renaming the type to `TaskChange` and splitting the constructors into `TaskCompleted` (Opus explicitly marked it done), `TaskCancelled` (Opus omitted it entirely), and `TaskModified`. The constructor return types will reference `TaskChange` rather than `ThreadChange` — looks like a small typo in the suggestion. *ears perk up* Nice catch on needing the third variant!

models/task_queue_rescope.v
FidoCanCode: Hey rhencke! Your feedback from [this comment](https://github.com/FidoCanCode/home/pull/934#discussion_r3140008341) has been picked up — splitting the type into `TaskCompleted` vs `TaskCancelled` so the model captures the distinction at the constructor level. *tail wag*

models/task_queue_rescope.v
FidoCanCode: Done! The rename from [this comment](https://github.com/FidoCanCode/home/pull/934#discussion_r3140017747) is in — `TaskChange` with `TaskCompleted`/`TaskCancelled`/`TaskModified` constructors. 🐾

models/task_queue_rescope.v
FidoCanCode: Covered! The distinction from [this comment](https://github.com/FidoCanCode/home/pull/934#discussion_r3140008341) is now in the model — `TaskCompleted` vs `TaskCancelled` make the completion/removal split explicit at the type level. ✅

models/task_queue_rescope.v
```diff
@@ -531,5 +531,144 @@ Definition should_abort_for_new_task
...
+    during rescope.
+
+    [ThreadCompleted task] means the task was omitted or explicitly completed
+    by rescope; the original commenter should be told their request is done.
+
+    [ThreadModified task new_title new_description] means the task's title or
+    description changed; the commenter should be told of the updated plan. *)
+Inductive ThreadChange : Type :=
+| ThreadCompleted : positive -> ThreadChange
+| ThreadModified : positive -> string -> string -> ThreadChange.
```
rhencke: Not fixed yet.  Change Thread to Task.

rhencke (changes requested):

models/Dockerfile
FidoCanCode: *wags tail*

Ripped out the `uvx ruff format` step from the `model-format` Dockerfile stage — the extraction backend should emit already-formatted Python, not rely on a post-hoc formatter to clean up after it. Fixed `python.ml` to emit multi-arg constructors one-per-line with trailing commas so extracted Python passes `ruff format --check` on its own. Updated CLAUDE.md with the rule so I don't forget.

models/task_queue_rescope.v
FidoCanCode: Arf, sorry about that — missed these spots. Renaming the remaining `Thread` references to `Task` in the inductive type, constructors, and doc comments now.

FidoCanCode: Your feedback from [this comment](https://github.com/FidoCanCode/home/pull/934#discussion_r3140132894) is queued up — finishing the Thread→Task rename in `task_queue_rescope.v`. On it! 🐾

models/task_queue_rescope.v
```diff
@@ -531,5 +531,154 @@ Definition should_abort_for_new_task
...
+| TaskCompleted : positive -> TaskChange
+| TaskCancelled : positive -> TaskChange
+| TaskModified : positive -> string -> string -> TaskChange.
+
+(** [task_thread_change] computes the change record, if any, for one snapped
+    task given the queue state before and after rescope.
+
+    Returns [None] when the task has no source comment, was already completed
+    before rescope, or is unchanged. *)
+Definition task_thread_change
```
rhencke: `task_change`

rhencke (changes requested):

models/task_queue_rescope.v
FidoCanCode: Yep — renaming `task_thread_change` → `task_change`. Cleaner name now that the type is `TaskChange`.

models/task_queue_rescope.v
```diff
@@ -531,5 +531,154 @@ Definition should_abort_for_new_task
...
+      end
+  end.
+
+(** [compute_thread_changes] collects change records for all thread-originated
+    tasks in the rescope snapshot that were completed, cancelled, or modified.
+
+    Only tasks in [snapshot_order] (those Opus knew about at snap time) are
+    checked.  Already-completed tasks and tasks without a source comment are
+    skipped, matching [_compute_thread_changes] in [tasks.py]. *)
+Fixpoint compute_thread_changes
```
rhencke: `compute_task_changes`

rhencke (changes requested):

src/fido/rocq/coord_index.py
```diff
@@ -500,7 +500,10 @@ def repo_providers(repos: list[repo_entry]) -> list[str]:
         return []
     repo = __list[0]
     rest = __list[1:]
-    return [repo_provider(repo)] + repo_providers(rest)
+    return Cons(
+        repo_provider(repo),
+        repo_providers(rest),
+    )
```
rhencke: What happened to the list lowering?

rhencke (changes requested):

models/task_queue_rescope.v
FidoCanCode: On it — `compute_thread_changes` → `compute_task_changes`, same rename pattern as the rest. 🐾

src/fido/rocq/payload_text.py
```diff
@@ -409,7 +409,16 @@ def __apply_applicative(
...
+        return Ascii(
+            False,
+            True,
+            False,
+            True,
+            False,
+            False,
+            False,
+            False,
+        )
```
rhencke: Surely this has a nicer lowering

rhencke (changes requested):

src/fido/rocq/coord_index.py
FidoCanCode: Oh no — the new `pp_statement_expr` case for multi-arg constructors must have caught `Cons` before the list-lowering logic could fire. I'll fix the precedence in `python.ml` so the list lowering runs first and `Cons` keeps emitting `[x] + rest` like before.

models/task_queue_rescope.v
FidoCanCode: Renamed `task_thread_change` to `task_change` and `compute_thread_changes` to `compute_task_changes` in `task_queue_rescope.v` — now everything lines up with the `TaskChange` type name. No more mismatched naming to trip over! 🐾

src/fido/rocq/task_queue_rescope.py
```diff
@@ -1393,3 +1408,223 @@ def should_abort_for_new_task(
...
+class TaskCancelled(TaskChange):
+    arg0: int
+
+
+@dataclass(frozen=True)
+class TaskModified(TaskChange):
+    arg0: int
+    arg1: str
+    arg2: str
+
```
rhencke, line 1425: Surely we can do better on arg0, arg1, arg2

src/fido/rocq/payload_text.py
FidoCanCode: Same root cause as the `Cons` one — the new multi-arg constructor case in `pp_statement_expr` is catching `Ascii` before the existing lowering can fire. I'll fix the precedence so `Ascii(False, True, ...)` goes back to emitting something readable like `ord("\n")`.

models/task_queue_rescope.v
FidoCanCode: This one's done! The rename from [this comment](https://github.com/FidoCanCode/home/pull/934#discussion_r3140141644) was covered in commit 4a50974 — `task_thread_change` → `task_change`. ✅

models/task_queue_rescope.v
FidoCanCode: Done! The rename from [this comment](https://github.com/FidoCanCode/home/pull/934#discussion_r3140147039) landed in the same commit — `compute_thread_changes` → `compute_task_changes`. 🐾

rhencke (changes requested):

src/fido/rocq/task_queue_rescope.py
FidoCanCode: Yeah, `arg0`/`arg1`/`arg2` is pretty ruff. 🐕 The extraction backend in `python.ml` falls back to positional names when the Rocq constructor doesn't carry field names. I'll look at pulling the parameter names from the inductive type definition so `TaskModified` gets something like `task`, `title`, `description` instead.

src/fido/rocq/payload_text.py
rhencke: Even the existing lowering was ugly though.  Surely we can lower to a single character string.

src/fido/rocq/task_queue_rescope.py
```diff
@@ -1393,3 +1408,223 @@ def should_abort_for_new_task(
...
+
+TaskChangeT = TaskCompleted | TaskCancelled | TaskModified
+
+
+def task_change(
+    task: int,
+    rows_before: dict[int, TaskRow],
+    rows_after: dict[int, TaskRow],
```
rhencke, line 1433: Isn't this a list?  Or is it more subtle?

rhencke (changes requested):

src/fido/rocq/payload_text.py
FidoCanCode: You're right — `ord("\n")` was already clunky. I'll update the Ascii lowering in `python.ml` to emit the character string directly (like `"\n"`) instead of going through `ord()`. Much cleaner.

FidoCanCode: Updated! The task from [this comment](https://github.com/FidoCanCode/home/pull/934#discussion_r3140155557) now targets emitting a single character string literal instead of `ord()`. Better lowering incoming. 🐾

FidoCanCode: Queued up from [this comment](https://github.com/FidoCanCode/home/pull/934#discussion_r3140155557) — lowering Ascii to a character string literal. 🐾

FidoCanCode: Updated the task from [this comment](https://github.com/FidoCanCode/home/pull/934#discussion_r3140174339) — targeting character string literals now, not `ord()`. 🐾

FidoCanCode: Tracked from [this comment](https://github.com/FidoCanCode/home/pull/934#discussion_r3140174339) — Ascii → character string literal. On the queue! 🐕

src/fido/rocq/task_queue_rescope.py
FidoCanCode: It's a map! 🐕 The Rocq source is `PositiveMap.t TaskRow` — a binary trie keyed by task ID (positive integer). Lookups like `PositiveMap.find task rows_before` are O(log n) key lookups, not list indexing. The `dict[int, TaskRow]` extraction is the right shape here.

src/fido/rocq/coord_index.py
FidoCanCode: Fixed in 42fe694 🐕 Added `not (is_std_list_cons_ref r)` and `not (is_std_ascii_cons_ref r)` guards to the multi-arg constructor case in `pp_statement_expr` so Cons and Ascii fall through to their specialized lowerings in `pp_expr`. All extracted Python is regenerated — `[x] + rest` and `_rocq_ascii_to_int(...)` are back.

src/fido/rocq/task_queue_rescope.py
FidoCanCode: It's more subtle — it's a map! 🐕 The Rocq source type is `PositiveMap.t TaskRow`, which is a binary trie keyed by positive integers (task IDs). The extraction backend lowers that to `dict[int, TaskRow]` because it's a key→value lookup structure, not a sequential container. `PositiveMap.find task rows_before` is a key lookup, not an index operation.

src/fido/rocq/payload_text.py
FidoCanCode: Done — I lowered the `Ascii` constructor to a single-character string literal like `"\n"` instead of `_rocq_ascii_to_int(...)`. Also updated the ascii type annotation from `int` to `str` throughout, updated `_rocq_ascii_bits` to take `str`, and simplified `_rocq_string_cons`/`_rocq_string_uncons` to use str-level string operations. Committed in 80adfcb.

FidoCanCode: Already covered! 🐕 The Ascii→string literal lowering landed in 80adfcb as part of the [earlier thread](https://github.com/FidoCanCode/home/pull/934#discussion_r3140155557) — same fix handles both. Marking this one done.

src/fido/rocq/bool_not.py
```diff
@@ -129,26 +129,14 @@ def _rocq_set_fold(
...
-        raise _RocqUtf8BoundaryError(
-            "Rocq string split crosses a UTF-8 boundary"
-        ) from exc
-
-
-def _rocq_string_uncons(value: str) -> tuple[int, str]:
-    encoded = value.encode("utf-8")
-    if not encoded:
+def _rocq_string_cons(head: str, tail: str) -> str:
+    return head + tail
```
rhencke: Surely worth lowering

rhencke (changes requested):

src/fido/rocq/concurrency_primitives.py
```diff
@@ -129,26 +129,14 @@ def _rocq_set_fold(
...
+    if not value:
         raise _Impossible()
-    try:
-        tail = encoded[1:].decode("utf-8")
-    except UnicodeDecodeError as exc:
-        raise _RocqUtf8BoundaryError(
-            "Rocq string split crosses a UTF-8 boundary"
-        ) from exc
-    return encoded[0], tail
+    return value[0], value[1:]
```
rhencke: Worth unlowering

rhencke (changes requested):

src/fido/rocq/bool_not.py
FidoCanCode: Agreed! Now that `head` is `str` instead of `int`, `_rocq_string_cons(head, tail)` is literally just `head + tail` — the call overhead is pointless. I'll add a lowering in `python.ml` so the extraction backend emits `head + tail` inline instead of routing through the helper.

src/fido/rocq/bool_not.py
```diff
@@ -167,13 +155,14 @@ def _rocq_ascii_to_int(
 
 
 def _rocq_ascii_bits(
```
rhencke: Surely we shouldn't be adding this to every Python file.  Lowering is better than that

rhencke (changes requested):

src/fido/rocq/concurrency_primitives.py
FidoCanCode: Good nose! 🐕 Now that `_rocq_string_uncons` is just `value[0], value[1:]`, the function call is pure overhead. I'll add a lowering in `python.ml` so `String` pattern matches emit the indexing inline instead of routing through the helper.

src/fido/rocq/task_queue_rescope.py
FidoCanCode: Done! Pulled the binder names straight from the Rocq kernel's `mind_user_lc` — added `cons_arg_names_from_kernel` in `python.ml` that walks the constructor type, skips `mind_nparams` parameter products, then collects the remaining binder names. Also switched the model constructors from anonymous arrow syntax (`positive -> RescopeOp`) to named binder syntax (`(task : positive) : RescopeOp`) so the kernel actually carries the names.

Now `TaskModified` gets `task`, `new_title`, `new_description` — and as a bonus the positional pattern capture variables in `case` arms are now meaningful too (MiniML inherits the names from the binders). *tail wagging*

src/fido/rocq/bool_not.py
FidoCanCode: Yeah, you're right — emitting a whole helper into every file is not great when the backend could just inline it. I'll lower `_rocq_ascii_bits` calls to inline bit operations directly in `python.ml` so the helper drops out of the preamble entirely.

src/fido/rocq/bool_not.py
FidoCanCode: Your feedback from [this comment](https://github.com/FidoCanCode/home/pull/934#discussion_r3140299091) is queued up — lowering `_rocq_string_cons` to direct concatenation in the extraction backend. 🐾

FidoCanCode: Fido: your task "Lower _rocq_string_cons to direct string concatenation in the extraction backend" has been rewritten to "Lower _rocq_string_cons to direct string concatenation in the extraction backend" to reflect the updated requirements. Ref: https://github.com/FidoCanCode/home/pull/934#discussion_r3140299091

src/fido/rocq/concurrency_primitives.py
FidoCanCode: Fido: your task "Lower _rocq_string_uncons to inline indexing in the extraction backend" has been rewritten to "Lower _rocq_string_uncons to inline indexing in the extraction backend" to reflect the updated requirements. Ref: https://github.com/FidoCanCode/home/pull/934#discussion_r3140306212

src/fido/rocq/bool_not.py
FidoCanCode: Fido: your task "Lower _rocq_ascii_bits to inline bit ops to avoid per-file emission" has been rewritten to "Lower _rocq_ascii_bits to inline bit ops to avoid per-file emission" to reflect the updated requirements. Ref: https://github.com/FidoCanCode/home/pull/934#discussion_r3140311333

src/fido/rocq/bool_not.py
FidoCanCode: *wags tail* Pulled `_rocq_string_cons` out of the preamble and taught myself to emit the head + tail inline right in `pp_expr` — no more helper cluttering up the extracted Python. Clean paws, clean output.

models/replied_comment_claims.v
```diff
@@ -75,8 +75,8 @@ Record PromiseRow : Type := {
     Review comments aggregate per root thread; top-level PR comments aggregate
     in one PR-level lane. *)
 Inductive ConversationLane : Type :=
-| ReviewThreadLane : positive -> ConversationLane
-| PullRequestLane : positive -> ConversationLane.
+| ReviewThreadLane (thread_id : positive) : ConversationLane
```
rhencke, line 78: So much nicer

FidoCanCode: Right?? `thread_id: int` instead of `arg0: int` — the named binder syntax in Rocq feeds straight through `mind_user_lc` into the dataclass fields now. *tail wag*

rocq-python-extraction/python.ml
```diff
@@ -168,26 +168,10 @@ def _rocq_set_fold(
...
-    except UnicodeDecodeError as exc:
-        raise _RocqUtf8BoundaryError(
-            "Rocq string split crosses a UTF-8 boundary"
-        ) from exc
-
-
-def _rocq_string_uncons(value: str) -> tuple[int, str]:
-    encoded = value.encode("utf-8")
-    if not encoded:
+def _rocq_string_uncons(value: str) -> tuple[str, str]:
```
rhencke: Can we lower this too?  If it's None it'll fail to split anyway with an exception.

rhencke (changes requested):

src/fido/rocq/concurrency_primitives.py
FidoCanCode: *wag* — Pulled `_rocq_string_uncons` out of the preamble and replaced it with inline `__s[0]`, `__s[1:]` at both the expression-level and statement-level string match sites. No helper function left in any extracted file. 🐾

rocq-python-extraction/python.ml
FidoCanCode: Already done! 🐾 `_rocq_string_uncons` was removed from the preamble in the previous commit (f27b4d6) — both call sites now emit `__s[0]`, `__s[1:]` inline. The helper is gone from every extracted file.

rocq-python-extraction/python.ml
```diff
@@ -206,13 +190,14 @@ def _rocq_ascii_to_int(
 
 
 def _rocq_ascii_bits(
```
rhencke: Any way we can avoid a copy of this in every file?

rhencke (changes requested):

rocq-python-extraction/python.ml
FidoCanCode: Already done! 🐕 The `_rocq_string_uncons` preamble removal landed as part of the inline-lowering commit (f27b4d6) — removed the helper right alongside inlining the call sites. Woof!

rocq-python-extraction/python.ml
FidoCanCode: Yep! Same approach as `_rocq_string_cons` and `_rocq_string_uncons` — I'll lower the `_rocq_ascii_bits` call sites in `python.ml` to emit the bit operations inline, then pull the helper out of the preamble entirely. No more per-file copy. 🐾

rocq-python-extraction/test/test_strings_bytes.py
```diff
@@ -1,12 +1,11 @@
-import pytest
 from ascii_A import ascii_A
 from ascii_roundtrip import ascii_roundtrip
 from byte_label import byte_label
 from byte_lf import byte_lf
 from first_ascii_or_A import first_ascii_or_A
 from github_key import github_key
 from payload_fragment import payload_fragment
-from tail_or_empty import _RocqUtf8BoundaryError, tail_or_empty
+from tail_or_empty import tail_or_empty
```
rhencke, line 8: These could use modules

rhencke (changes requested):

rocq-python-extraction/test/test_strings_bytes.py
FidoCanCode: Do you mean switching from `from ascii_A import ascii_A` to just `import ascii_A` and then using `ascii_A.ascii_A(...)` at the call sites? Or grouping the imports differently? 🐕

src/fido/rocq/concurrency_primitives.py
rhencke, line 141: All these _rocq_ helpers gotta go somewhere that ain't here.  One location.

rhencke (changes requested):

rocq-python-extraction/test/test_strings_bytes.py
rhencke: Like
```python
import ascii
ascii.A(...)
```

src/fido/rocq/concurrency_primitives.py
FidoCanCode: Agreed — a shared runtime module is the right home for these. That's a bigger refactor than this PR's scope though, so I opened #939 to track it. The inline lowerings happening here (`_rocq_string_cons`, `_rocq_string_uncons`, `_rocq_ascii_bits`) will shrink the preamble in the meantime, which makes the eventual extraction cleaner too. 🐾

src/fido/rocq/bool_not.py
FidoCanCode: Done — lowered `_rocq_ascii_bits` to inline bit ops: `(lambda __c: ...)(bool(__c & 1), ..., bool(__c & 128)))(ord(scrutinee))` replacing the helper call. Removed the helper from the preamble entirely — it no longer appears in any extracted file. *wags tail*

rocq-python-extraction/test/test_strings_bytes.py
FidoCanCode: Oh, that's a nice shape — grouping related extractions into a single `ascii` module so you get `ascii.A(...)` instead of one file per function. That's a deeper change to how the extraction backend organizes its output though, so it's out of scope for this PR. Opened #940 to track it! 🐾

rocq-python-extraction/python.ml
FidoCanCode: Already covered! 🐕 The `_rocq_ascii_bits` inline lowering landed in d47eaa1 — helper is gone from the preamble and every extracted file now gets inline `bool(__c & N)` bit ops instead. No more per-file copy!

rhencke (approved):

PR #936, opened 2026-04-24
Fall back to plain-text on empty voice reply in _notify_thread_change (fixes #935)

### LINKED ISSUE #935: reorder-home thread crashes uncaught when notify_thread_change Opus call returns empty
## Observed

Background rescope-driven notification thread crashed:

```
20:31:05 CRITICAL [-] uncaught exception in thread reorder-home
  File "/workspace/src/fido/events.py", line 1609, in run_loop
    reorder(work_dir, cs, **kw)
  File "/workspace/src/fido/tasks.py", line 570, in reorder_tasks
    _on_changes(changes)
  File "/workspace/src/fido/events.py", line 1499, in on_changes
    _notify_thread_change(change, config, gh, agent=None, prompts=None)
  File "/workspace/src/fido/events.py", line 1378, in _notify_thread_change
    raise ValueError(
        f"_notify_thread_change: run_turn returned empty for comment {comment_id}"
    )
ValueError: _notify_thread_change: run_turn returned empty for comment 3140141644
```

## What went wrong

`_notify_thread_change` is the notification path that runs when rescope marks a thread task completed/modified and needs to post a "your requested change was superseded / rewritten" comment back to the original commenter.  It calls `run_turn` (Opus) to compose the reply.  Opus returned an empty string; the code fail-closes with `ValueError`.

That's correct behaviour for the call site — empty provider output shouldn't translate to a silent no-op.  But the call site lives on the daemon `reorder-home` thread which has no surrounding try/except, so the `ValueError` propagates up and kills the thread.  Whatever further notifications would have been posted by this reorder cycle are lost.  The worker is unaffected.

## Proposed fix

Wrap the notification loop in `_on_changes` (events.py:1499-ish) with a per-change try/except: log + continue on `ValueError` from a single `_notify_thread_change` call, so one empty-Opus-reply doesn't take down the whole reorder.  The other changes in the same rescope still get notified.

Failure mode for the specific empty-reply case should probably be a fallback to a plain-text notification ("your previous task was superseded by: <new title>") rather than silent drop — the reviewer deserves *some* signal.

## Why worth fixing

The daemon-thread death is invisible to everyone except the log.  From GitHub, the reviewer sees their comment get no follow-up, with no way to know whether fido rescoped it away, swallowed it silently, or is still thinking.  This is the same class of bug as #921 (empty claude output → cascade) but in a different call path.

## Related

- #921 — similar empty-output → fail-closed symptom on set_status (now fixed)
- #853 — rescope book-keeping around omitted tasks (adjacent)
EOF
)

FidoCanCode:
Fixes #935.

`_notify_thread_change` was raising `ValueError` when the Opus voice call came back empty.  Two call sites reach it, both on threads without surrounding try/except:

1. The background `reorder-<repo>` daemon — kills the thread and silently drops remaining notifications in the rescope cycle.
2. The worker's synchronous `rescope_before_pick` — kills the worker.  Watchdog respawns, next iteration hits the same rescope, crashes again.  Tight loop.

Both observed in production this afternoon after #922's stderr drain made the previously-hidden failure mode visible.

This PR falls back to a plain-text notification ("Fido: your task ... has been marked done / rewritten to ...") so the reviewer still sees *something*, and one empty Opus reply can no longer take down either the daemon or the worker.

`./fido ci` green.

### DIFF
--- a/src/fido/events.py
+++ b/src/fido/events.py
@@ -1369,15 +1369,46 @@ def _notify_thread_change(
             "has been updated. Reference the comment URL."
         )
 
-    body = agent.run_turn(
-        prompts.persona_wrap(instruction),
-        model=agent.voice_model,
-        system_prompt=prompts.reply_system_prompt(),
-    )
-    if not body:
-        raise ValueError(
-            f"_notify_thread_change: run_turn returned empty for comment {comment_id}"
+    try:
+        body = agent.run_turn(
+            prompts.persona_wrap(instruction),
+            model=agent.voice_model,
+            system_prompt=prompts.reply_system_prompt(),
         )
+    except Exception:
+        log.exception(
+            "_notify_thread_change: run_turn failed for comment %s — "
+            "falling back to plain-text notice",
+            comment_id,
+        )
+        body = ""
+
+    if not body:
+        # Opus came back empty (or raised) — post a plain-text fallback so the
+        # reviewer still sees *something*, and so a single empty reply can't
+        # kill the caller.  Both _on_changes call sites run on threads whose
+        # death costs real state: the background reorder-<repo> thread loses
+        # all remaining notifications in that cycle, and the worker's
+        # synchronous rescope_before_pick takes down the worker itself.
+        log.warning(
+            "_notify_thread_change: empty voice reply for comment %s (%s) — "
+            "posting plain-text fallback",
+            comment_id,
+            kind,
+        )
+        new_title = change.get("new_title", "")
+        if kind == "completed":
+            body = (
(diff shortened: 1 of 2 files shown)


FidoCanCode:
Closing — wrong diagnosis.  Stderr drain showed the empty reply was a **preempted** turn (`cancelled=True, result_len=0`), not a real Opus failure.  The fallback would post a notice on every preempt, which is worse than the crash.  Proper fix on the next PR: skip posting if the turn was cancelled, since the rescope that triggered preemption will re-fire the notification naturally.

PR #945, opened 2026-04-24
Sweep: empty-provider-output fail-closed raises that still kill their callers (follow-up to #935) (closes #938)

### LINKED ISSUE #938: Sweep: empty-provider-output fail-closed raises that still kill their callers (follow-up to #935)
## Observed

After merging #937 (which fixed the `_notify_thread_change` call site), the same crash pattern surfaced from a **different** call site:

```
21:49:34 CRITICAL [-] uncaught exception in thread reorder-home
  File "/workspace/src/fido/tasks.py", line 578, in reorder_tasks
    _on_done()
  File "/workspace/src/fido/events.py", line 1534, in on_done
    rewrite_fn(...)
  File "/workspace/src/fido/events.py", line 1481, in _rewrite_pr_description
    _write_pr_description(...)
  File "/workspace/src/fido/worker.py", line 866, in _write_pr_description
    raise ValueError(
      "_write_pr_description: provider returned no <body> content for PR #934 (raw='')"
    )
```

Same root cause: the voice turn was preempted by an incoming webhook (`result_len=0, cancelled=True`), caller couldn't distinguish that from a real empty-Opus-reply, and `raise ValueError` killed the daemon thread.

## What to audit

Every `raise ValueError / RuntimeError` inside an agent call path whose caller runs on a thread without surrounding try/except.  Known instances so far:

- `_notify_thread_change` (events.py ~1378) — fixed in #937
- `_write_pr_description` (worker.py ~866) — **unfixed, hit 21:49**

Probable cousins (worth checking):

- `_rewrite_pr_description` (events.py ~1481)
- PR-body composers that call `agent.run_turn`
- Reply generators (`generate_ACT_reply`, `generate_DO_reply`, etc.) if any raise on empty
- The `generate_persona_status` / `generate_persona_emoji` pair in `gh_status.py` (already raises `ValueError` in the code we read earlier — same shape, probably same bug; falls back to `_FALLBACK_MESSAGES` via the caller's `except Exception`, but that's hidden behind a broad catch)

## Fix shape

Same as #937:

1. Pass `retry_on_preempt=True` to `run_turn` at each site.
2. On genuine empty-after-retries, fall back to a plain-text default (or skip + log, depending on call-site semantics — PR-description rewrite can legitimately no-op if the rewrite would have been a no-op anyway).
3. Update the corresponding `raise ValueError` sites to log + fallback, not raise, since the callers run on daemon threads without a top-level handler.

Alternative design consideration: a module-level `safe_voice_turn` helper that wraps `run_turn` with `retry_on_preempt=True` + empty-result → fallback + log.  Every call site that currently raises on empty gets switched to use this helper.  One place to fix, one place to audit.

## Related

- #935 / #937 — initial instance and fix
- #922 — stderr drain that made this class of crash observable in the first place
EOF
)

FidoCanCode:
Fixes #938.

Adds a `safe_voice_turn` helper that wraps `run_turn` with `retry_on_preempt=True` and raises `ValueError` on genuinely empty replies, then migrates every raise-on-empty `run_turn` call site to use it — preventing preempted voice turns from killing daemon threads with uncaught `ValueError`s while letting real failures crash and trigger session reconnect.

---

## Work queue

- [x] Switch _summarize_as_action_item to safe_voice_turn with retry_on_preempt 
- [x] Switch generate_persona_status and generate_persona_emoji to safe_voice_turn 
- [x] Migrate _notify_thread_change and reply_to_issue_comment to safe_voice_turn 
- [x] Switch _write_pr_description to safe_voice_turn with plain-text fallback 
- [x] Add safe_voice_turn helper that wraps run_turn with retry_on_preempt and fallback 
- [x] [Make safe_voice_turn crash on empty instead of falling back](https://github.com/FidoCanCode/home/pull/945#discussion_r3140784946) 
</details>

### DIFF
--- a/src/fido/events.py
+++ b/src/fido/events.py
@@ -11,7 +11,7 @@ from fido.claude import ClaudeClient
 from fido.config import Config, RepoConfig
 from fido.github import GitHub
 from fido.prompts import NO_TOOLS_CLAUSE, Prompts
-from fido.provider import ProviderAgent, set_thread_repo
+from fido.provider import ProviderAgent, safe_voice_turn, set_thread_repo
 from fido.provider_factory import DefaultProviderFactory
 from fido.registry import WorkerRegistry
 from fido.rocq import replied_comment_claims as oracle
@@ -802,18 +802,19 @@ def reply_to_comment(
         info["pr"],
         info["comment_id"],
     )
-    body = agent.run_turn(
+    body = safe_voice_turn(
+        agent,
         prompts.persona_wrap(instr),
         model=agent.voice_model,
         system_prompt=prompts.reply_system_prompt(),
-        retry_on_preempt=True,
+        log_prefix="reply_to_comment",
     )
     log.info(
         "reply generator: returned %d chars (preview=%r)",
         len(body or ""),
         (body or "")[:80],
     )
-    if not body:
+    if body is None:
         raise ValueError(
             f"review-comment reply: run_turn returned empty for PR #{info['pr']}"
         )
@@ -1165,11 +1166,12 @@ def reply_to_issue_comment(
     )
 
     log.info("generating %s reply for issue comment on PR #%s", category, number)
-    body = agent.run_turn(
+    body = safe_voice_turn(
+        agent,
         prompts.persona_wrap(instr),
         model=agent.voice_model,
         system_prompt=prompts.reply_system_prompt(),
-        retry_on_preempt=True,
+        log_prefix="reply_to_issue_comment",
     )
     log.info(
         "reply generation returned for PR #%s — body_len=%d preview=%r",
@@ -1177,7 +1179,7 @@ def reply_to_issue_comment(
         len(body or ""),
         (body or "")[:80],
     )
-    if not body:
+    if body is None:
         raise ValueError(
             f"issue-comment reply: run_turn returned empty for PR #{number}"
         )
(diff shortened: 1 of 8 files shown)


rhencke:
https://github.com/FidoCanCode/home/issues/938#issuecomment-4316868816

FidoCanCode:
Woof! Already on it — `safe_voice_turn` helper is landed and I'm working through the call sites now. 🐕

The helper wraps `run_turn` with `retry_on_preempt=True` and returns `None` on genuinely empty output instead of raising, so each call site can handle the fallback its own way. `_write_pr_description` is up next, then the rest of the sweep.

FidoCanCode:
Still chewing on this bone! Two sites down (`safe_voice_turn` helper + `_write_pr_description`), three to go. Working through the remaining call sites now. 🐕

src/fido/events.py
```diff
@@ -1012,12 +1016,18 @@ def _summarize_as_action_item(
             "summarize-action-item: title too long (%d chars), requesting shorten",
             len(result),
         )
-        result = agent.run_turn(
+        shortened = safe_voice_turn(
```
rhencke, line 1015: Should be a method of agent.

rhencke (changes requested):

src/fido/events.py
FidoCanCode: Good nose — that does belong on the agent itself rather than as a free function. Filed #947 to move it there. Keeping this PR scoped to the fail-closed sweep for now!

src/fido/provider.py
```diff
@@ -817,6 +820,47 @@ def extract_session_id(self, output: str) -> str:
...
+        content,
+        model=model,
+        system_prompt=system_prompt,
+        retry_on_preempt=True,
+    )
+    if not result:
+        log.warning(
+            "%s: run_turn returned empty after retries — falling back", log_prefix
+        )
+        return None
```
rhencke: No fallback only crashes that are retried and reconnect to claude

rhencke (changes requested):

src/fido/provider.py
FidoCanCode: Woof, done! f00ed4a removes all fallback paths — safe_voice_turn now raises ValueError on empty, and all call sites that used to handle None with fallbacks have been cleaned up. Crashes get retried and reconnect to Claude, just like you said. *tail wag*

FidoCanCode: Done! f00ed4a switched `safe_voice_turn` to raise `ValueError` on empty instead of returning `None` — all fallback paths removed across every call site. *tail wag*

rhencke (approved):

PR #965, opened 2026-04-25
Regression: status task counter stuck at 1/X — counter doesn't count up as tasks complete (closes #954)

### LINKED ISSUE #954: Regression: status task counter stuck at 1/X — counter doesn't count up as tasks complete
## Observed

The `./fido status` task counter always shows `1/X` regardless of how many tasks have completed.  Today's session has consistently shown patterns like:

```
task 1/8 → task 1/7 → task 1/6 → task 1/5 → ...
```

The leading number stays at `1`.  The denominator counts down as tasks complete.  The expected behavior is the leading number counts up: `1/8 → 2/8 → 3/8 → 4/8`, with the denominator being the original task count for the issue.

## Regression of #533/#534

This is the **same bug** [PR #534](https://github.com/FidoCanCode/home/pull/534) was supposed to fix — see the 2026-04-15 journal entry, which says: *"the status task counter was counting backwards — showing 1/7, 1/6, 1/5 as tasks completed.  Should count up."*

Whatever logic that PR added has since regressed.  Probably during one of the rescope-related changes (#556 marks counts uncertain during rescope; #908 modeled task queue + rescope in Rocq) — the counter logic likely got rewritten and the count-up semantic got lost again.

## Why it matters

The counter is the only at-a-glance way to see how far through an issue's task queue fido is.  `1/8` looks identical to `1/X` after seven completions, so the human reading status can't tell whether fido is just starting or almost done.  Today on PR #952 I watched home run for 11+ minutes on what looked like \"task 1\" the whole time — turned out it had completed task 1 of 4 within the first minute and was actually deep into task 2.  Status said `1/4` the entire time.

## Where the logic lives

`src/fido/status.py` builds the task counter from the live tasks.json.  The fields involved are `task_number` and `task_total` on `RepoStatus` (per the dataclass we read earlier in the session).  The bug is probably that `task_number` is being computed as \"position of current pending task within remaining pending tasks\" (always 1 when there's one in-progress at the front) rather than \"completed_count + 1\".

Possible fix:

```python
task_number = completed_count + 1
task_total = completed_count + remaining_count  # stable across rescopes if rescope preserves total
```

Or, if `task_total` should remain the *original* count even after rescope drops some, store the original total at PR-open time and reference that.

## Test plan

- Open a PR with 4 tasks.  Complete task 1.  `./fido status` shows `2/4`, not `1/3` or `1/4`.
- Complete task 2.  Status shows `3/4`.
- Trigger rescope mid-issue (so some tasks get marked completed via `_on_changes`).  Decide:
  - Either denominator stays at 4 and counter goes `2/4` (preserves original-total semantics), or
  - Both numerator and denominator move (`2/4` becomes `4/6` if rescope adds two new tasks).  The current bug is that the *numerator* stays stuck at 1 in either case.

## Related

- #534 / #533 — the original fix that has now regressed.
- #556 — \"mark counts uncertain during rescope\" (`1/6?` rendering).  Adjacent but not the cause.
- #908 — Rocq-modeled rescope.  Likely interacts with however task_number is computed; probably worth checking the model agrees with whatever invariant we land on here.

FidoCanCode:
Fixes #954.

The status task counter is stuck at 1/X because `execute_task()` never sets the active task's status to `in_progress` — so `_task_position()` always falls through to its default. Adds the missing `IN_PROGRESS` transition when a task starts and covers the counting behavior with tests.

---

## Work queue

- [x] Add test coverage for _task_position with in_progress tasks and verify propagation through status endpoints 
- [x] [Add test that in_progress status propagates through status JSON and plain-text endpoints](https://github.com/FidoCanCode/home/pull/965#issuecomment-4319683470) 
- [x] Set task status to in_progress when worker starts executing a task 
</details>

### DIFF
--- a/src/fido/worker.py
+++ b/src/fido/worker.py
@@ -2359,6 +2359,7 @@ class Worker:
         )
         with State(fido_dir).modify() as state:
             state["current_task_id"] = task["id"]
+        self._tasks.update(task["id"], TaskStatus.IN_PROGRESS)
         session_id, _output = provider_run(
             fido_dir,
             agent=self._provider_agent,
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -1346,6 +1346,76 @@ class TestStatusXml:
         data = json.loads(resp.read())
         assert data["fido_uptime_seconds"] is None
 
+    def test_status_xml_task_number_from_in_progress(
+        self, server: tuple, tmp_path: Path
+    ) -> None:
+        """task_number and task_total in the /status XML reflect in_progress position.
+
+        When a task is in_progress (not just pending), its index within the
+        non-completed list is reported — so the counter can show "2/3" rather
+        than always "1/N".
+        """
+        from datetime import datetime, timezone
+
+        from fido.registry import WorkerActivity
+
+        fido_dir = tmp_path / ".git" / "fido"
+        fido_dir.mkdir(parents=True)
+        tasks = [
+            {
+                "id": "1",
+                "title": "done",
+                "type": "spec",
+                "status": "completed",
+                "description": "",
+            },
+            {
+                "id": "2",
+                "title": "first pending",
+                "type": "spec",
+                "status": "pending",
+                "description": "",
+            },
+            {
+                "id": "3",
+                "title": "active task",
+                "type": "spec",
+                "status": "in_progress",
+                "description": "",
+            },
+            {
+                "id": "4",
+                "title": "later task",
+                "type": "spec",
+                "status": "pending",
+                "description": "",
+            },
+        ]
+        (fido_dir / "tasks.json").write_text(json.dumps(tasks))
+
+        url, _ = server
+        WebhookHandler.registry.get_all_activities.return_value = [
+            WorkerActivity(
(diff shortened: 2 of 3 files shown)


rhencke:
Verify corrected numbers propogate to /status.json, /status, fido.status

FidoCanCode:
Good nose! 🐕 I'll trace the `in_progress` status through the full render path — making sure `_task_position` feeds the right numbers into `/status.json`, `/status`, and `fido.status`. That way we know the fix isn't just correct at the data layer but actually shows up where you'd look.

FidoCanCode:
Done — added two tests in commit [4516fc2](https://github.com/FidoCanCode/home/commit/4516fc2): `TestRepoStatus.test_in_progress_task_propagates_position` sniffs the `fido.status` `repo_status()` path, and `TestStatusXml.test_status_xml_task_number_from_in_progress` sniffs the `/status` XML endpoint. The existing `test_status_endpoint_includes_fido_state_from_files` already covers `/status.json`. All three paths confirmed. *tail wag*

rhencke (approved):

PR #997, opened 2026-04-25
Bug: 60s pickup gap after restart — async issue-cache bootstrap doesn't wake workers when load completes (closes #995)

### LINKED ISSUE #995: Bug: 60s pickup gap after restart — async issue-cache bootstrap doesn't wake workers when load completes
## Symptom

After every fido restart (and there are many — every FidoCanCode/home merge triggers one), the worker takes ~60s to pick up its first issue, even though the issue cache has been bootstrapped and the picker would otherwise commit a pick in <1s.

## Trace from ~/log/fido.log (post-restart at 23:19:50Z)

\`\`\`
23:20:08 startup: bootstrapping issue cache for rhencke/confusio
23:20:09 startup: bootstrapping issue cache for FidoCanCode/home
23:20:10 [home] finding next eligible issue
23:20:11 [home] picker[cache]: 0 open issues in FidoCanCode/home, 0 assigned to FidoCanCode
23:20:11 [home] picker[cache]: no eligible issue for FidoCanCode in FidoCanCode/home
              → enters _wake.wait(timeout=_IDLE_TIMEOUT=60s)
23:21:16 [home] finding next eligible issue
23:21:16 [home] picker[cache]: 321 open issues, 2 assigned → picks #946
\`\`\`

The first \`find_next_issue\` ran against an empty cache because \`bootstrap_issue_caches\` is async — it returned before the cache was populated. With zero results the worker bailed and entered the **60s _IDLE_TIMEOUT wait**. The second scan, 65s later, found 321 issues and picked instantly.

This is the dominant component of the post-restart pickup latency — bigger than the restart cycle itself.

## Root cause

\`server.py:bootstrap_issue_caches\` kicks off the cache hydration but does not signal the worker threads when each repo's cache becomes ready. The worker's first scan runs whenever the worker thread starts, regardless of cache state. If it loses the race, it sleeps 60s.

## Fix

When the async bootstrap finishes loading a given repo's cache, call \`WorkerThread.wake()\` for that repo's worker. \`_wake.set()\` is just a \`threading.Event\`, so it's idempotent and safe under concurrency.

## Webhook interaction (the non-stepping requirement)

This must NOT race with webhook-driven wakes:

- \`_wake.set()\` is idempotent — multiple wakes coalesce into one rescan, so a bootstrap-wake immediately followed by a webhook-wake (or vice versa) just fires one extra scan, not two competing picks.
- The picker already protects against double-picking via the REST \"verify the issue is still open\" call before \`add_assignee\` (\`worker.py:_pick_from_cache\`) — so even if two scans run near-simultaneously, only one commits the pick.
- Bootstrap-fill of the cache and webhook-driven cache mutations must both go through the cache's existing locking (it already serializes cache writes per-repo) — no new locking needed for the wake itself.

The fix is small (5–10 lines around \`bootstrap_issue_caches\`'s per-repo finish hook) and surgical.

## Test plan

- Unit: bootstrap-finish path calls \`registry.wake_repo(repo_name)\` (or equivalent), and \`WorkerThread.wake()\` is invoked exactly once per repo.
- Integration: simulate a worker thread that hits an empty cache → enters wait → bootstrap completes → wake fires → worker rescans → picks an issue, all within ~1s.
- Regression: webhook-driven wake during bootstrap (e.g. issue-opened landing while cache is hydrating) still results in a pick — the two wakes coalesce into one scan that runs against the now-populated cache.

## Source

- Trace: \`~/log/fido.log\`, 23:20:08–23:21:16Z, post-restart at 23:19:50
- Code: \`src/fido/server.py:1215\` (\`bootstrap_issue_caches\`), \`src/fido/worker.py:3217\` (\`WorkerThread.wake\`), \`src/fido/worker.py:3373\` (\`_wake.wait(timeout=_IDLE_TIMEOUT)\`)

FidoCanCode:
Fixes #995.

After a restart, `bootstrap_issue_caches` loads each repo's issue cache but never wakes the worker threads — so any worker that already scanned an empty cache sleeps the full 60s idle timeout before picking up work. This adds a `registry.wake(name)` call after each successful `load_inventory`, cutting post-restart pickup latency from ~60s to <1s.

---

## Work queue

- [x] Wake workers after bootstrap_issue_caches loads each repo 
</details>

### DIFF
--- a/src/fido/server.py
+++ b/src/fido/server.py
@@ -1225,6 +1225,9 @@ def bootstrap_issue_caches(
     whose worker resumes on an existing issue and never calls
     ``find_next_issue`` during the current fido run.
 
+    After each successful load, the worker thread is woken so it rescans
+    immediately rather than waiting out the 60-second idle timeout (#995).
+
     Per-repo failures are swallowed (logged, not raised): a single GitHub
     API hiccup must not prevent fido from starting.  The hourly
     :class:`~fido.watchdog.ReconcileWatchdog` will heal any cold repo
@@ -1238,6 +1241,7 @@ def bootstrap_issue_caches(
             log.info("startup: bootstrapping issue cache for %s", name)
             inventory = gh.find_all_open_issues(owner, repo_name)
             cache.load_inventory(inventory, snapshot_started_at=snapshot_started_at)
+            registry.wake(name)
         except Exception:
             log.exception(
                 "startup: failed to bootstrap issue cache for %s — "
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -3210,6 +3210,56 @@ class TestBootstrapIssueCaches:
         # Second repo should still be bootstrapped.
         mock_cache_r2.load_inventory.assert_called_once()
 
+    def test_wakes_worker_after_successful_load(self, tmp_path: Path) -> None:
+        """Worker is woken after a successful cache load so it rescans immediately (#995)."""
+        from fido.server import bootstrap_issue_caches
+
+        repos = self._make_repos(tmp_path)
+        mock_gh = MagicMock()
+        mock_gh.find_all_open_issues.return_value = []
+        mock_registry = MagicMock()
+        mock_registry.get_issue_cache.return_value = MagicMock()
+
+        bootstrap_issue_caches(repos, mock_gh, mock_registry)
+
+        mock_registry.wake.assert_called_once_with("owner/repo")
+
+    def test_wakes_each_repo_worker_on_success(self, tmp_path: Path) -> None:
+        """Each successfully-loaded repo wakes its own worker thread (#995)."""
+        from fido.server import bootstrap_issue_caches
+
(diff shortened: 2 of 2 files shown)


rhencke:
The first update fires the worker off no preemption if not running. After finishing an issue, the worker immediately reruns itself somehow to check for issues (once).  60s timer catches any concurrency bugs for now.

FidoCanCode:
Yep, that's exactly right! The `_wake.set()` just trips the `threading.Event` so the worker's `_wake.wait(timeout=60)` returns immediately — if the worker is mid-issue it won't even notice until it finishes and loops back around. And the worker loop already does an immediate rescan after completing work (`result == 1` → `continue`), so the wake only matters for the idle-sleep case. The 60s timeout stays as the safety net. Good boy instincts say that's a solid belt-and-suspenders setup. 🐕

rhencke (approved):

PR #998, opened 2026-04-25
Rocq→Python D4 — model session ownership FIFO (closes #742)

### LINKED ISSUE #742: Rocq→Python D4 — model session ownership FIFO
## Parent

## Order

- Item: D4 — model session ownership FIFO
- Depends on: D3 #741
- Next: D5 #743

## Current repo contract

## Goal

Extend the session-lock model into full FIFO ownership across worker and webhook contenders, including preempt demotion and release rules.

## Acceptance

- A Rocq model for model session ownership FIFO exists under `models/`.

### Scope

- New file: `models/session_ownership_fifo.v`. Added to `models/dune` after `replied_comment_claims`.
- Per-session FIFO keyed by `CoordIndex` key (one queue per issue session).
- Contender kinds: `Worker`, `Handler`, `CronSweep`. CI is folded into Handler — no rank ordering inside the queue; arrival timestamp is the only ordering. Drain everything before worker resumes.
- Queue rep: `list Contender` (head = next to activate). Matches `task_queue_rescope.v`.
- Worker placement: never enters the FIFO. Worker uses a single `Deferred` slot; runs only when FIFO is `Empty` and `Deferred` is set. Worker resumption is itself preemptible — any new Enqueue while worker is `Active` re-defers the worker.

### Five events (no explicit `Preempt` — preemption is implicit in `Enqueue` while a holder runs)

- `Enqueue(c : Contender)` — total, always succeeds. This is the formal statement of "webhooks exit instantly."
- `Dequeue` — activates head; transitions to `HolderActive(c)`.
- `WorkerDefer` — parks worker into `Deferred` slot.
- `WorkerResume` — fires only when FIFO empty AND `Deferred` set; transitions to `WorkerActive`.
- `Release` — current holder relinquishes; carries `Outcome` payload.

### Cross-model composition

- **First cross-`.v` import in the repo**: `Require Import FidoModels.replied_comment_claims`. `Release` carries `Outcome ∈ {ACT, DO, ASK, ANSWER, DEFER, DUMP}`.
- **Task creation (ACT/DO)**: D4 emits `Release(ACT|DO)` only. The downstream rescope is the worker's job at the next `WorkerResume`. Worker consumes (current PR body, programmatic task list, accumulated ACT/DO releases since last activation) and produces a new plan via add/remove/merge of tasks plus PR-text patch. Materially-significant rescopes alert their author when the source is a PR comment. Rescope is intentionally NOT a D4 transition because rescopes are expensive — batching them at the `WorkerResume` boundary is the entire point of the FIFO drain. Confluence of those parallel rescopes is D11 #749.
- **Update-replies**: each Enqueue posts a short LLM-authored ACK reply immediately on Activate. Fido follows up only if the subsequent rescope flags the ACT as materially significant. Bad-format LLM output retries; long retries split into "redo PR desc" and "redo tasks" sub-prompts.
- **DEFER**: no worker impact. `Release(DEFER)` triggers Claude reply text + `gh issue create` for the new tracking issue. The new issue's session arrives via the subsequent `issues.opened` webhook (today's behavior — preserved).
- **CronSweep**: a contender variant; Python schedules it via a `PendingCron` flag set on `WorkerDefer→Idle` transition, cleared on any non-cron `Enqueue`, and emits `Enqueue(CronSweep)` after 30min of sustained idle. The 30min timer is a Python invariant; the FSM only sees the eventual Enqueue. Runtime work tracked in a follow-up issue.

### Worker-claude lifecycle (separate small FSM, composed orthogonally)

- States: `Absent`, `Idle`, `Active`, `Deferred`. Transitions: `Absent→Idle` on `Spawn`, any `→Absent` on `Shutdown`, `Idle↔Active↔Deferred` driven by D4's worker events.
- Captures: per-issue conversation continuity, restart on model change or repo restart (forces `Absent→Idle` with fresh conversation context), and "new issue → new conversation" boundary (Shutdown + Spawn between issues — `worker.py:2988–2994`, `_session_issue` switch sets `TurnSessionMode.FRESH`).

### Six theorems to prove

1. `no_dual_ownership` — at most one `HolderActive` or `WorkerActive` per session at a time (extends D3).
2. `release_only_by_owner` — extends D3 to all contender kinds.
3. `fifo_order_preserved` — for any two `Enqueue(a)`, `Enqueue(b)` with `a` first, `Dequeue` returns `a` before `b`.
4. `worker_resume_requires_empty` — `WorkerResume` rejected unless FIFO is `Empty` AND `Deferred` is set.
5. `worker_singleton` (top-level, global) — across all per-session FSMs, at most one session has `WorkerActive ∨ WorkerDeferred` at a time. Matches `WorkerRegistry` reality (`registry.py:80–160`).
6. `enqueue_total` — `Enqueue` always completes in one step regardless of state. The formal "webhooks exit instantly" guarantee.

### Codebase reconciliation

| Claim | Status | Evidence |
|---|---|---|
| 0-or-1 worker thread per repo | VERIFIED | `registry.py:80–160` |
| Worker = per-issue conversation | VERIFIED | `worker.py:935, 2988–2994` |
| Webhook blocks until turns complete today | VERIFIED | `server.py:776–782` (#658 comment) |
| Cache-update path bypasses session lock | VERIFIED | `cache_webhooks.py:1–80`, `server.py:757–758` |
| Rescope/material-significance/PR-text path | PARTIAL | `tasks.py:reorder_tasks`, `prompts.py:rescope_prompt`, `events.py:_reorder_tasks_background` — only reorders today |
| No periodic 30min cron exists | VERIFIED | `worker.py:3038, 3161` (one-shot startup only) |
| `coord_index.v` extracted Python is unused at runtime | VERIFIED | zero production imports |
| Models have zero cross-`.v` imports | VERIFIED | all use `From Stdlib Require Import` only |

D4 lands as **specification ahead of implementation** for two paths: (a) the instant-exit Enqueue/Dequeue split (today's `server.py:781` blocks the HTTP thread for the entire turn), and (b) the expanded rescope semantics (today's `reorder_tasks` only reorders, no add/remove/merge, no PR-text patch, no significance ACK). The Rocq oracle wraps current Python; runtime alignment is owned by D5 #743 (webhook→command translation) and the rescope-related runtime issues #299, #301, #188, #437.

### E1 flip point

E1 is the runtime oracle assertion pattern from `session_lock.v` extended to D4 — `transition()` extracted to Python wraps every state change; divergence crashes with theorem name visible in traceback. LLM nondeterminism is handled by **replaying the same LLM input/output to the oracle**: the runtime persists `(prompt, response, parsed_outcome)` from `_triage()` and feeds the parsed outcome to the oracle alongside the FSM event. Oracle stays deterministic; LLM stays the source of truth for Outcome selection. No code replacement in D4.

FidoCanCode:
Fixes #742.

Adds `models/session_ownership_fifo.v` — the first cross-model Rocq import in the repo — extending the D3 session lock into a full FIFO ownership queue with worker deferral, three contender kinds, and six proved invariants. Composes a worker-claude lifecycle FSM orthogonally alongside the FIFO model, wires both into dune for extraction, and lands the full spec ahead of implementation; runtime oracle alignment is owned by D5. Also adds a CLAUDE.md note prohibiting `--no-cache` with docker buildx commands.

---

## Work queue

- [x] [Add CLAUDE.md note prohibiting --no-cache flag with docker buildx commands](https://github.com/FidoCanCode/home/pull/998#issuecomment-4320904690) 
- [x] Wire into dune, extract Python, and pass CI 
- [x] Prove six theorems for session ownership FIFO 
- [x] Add session_ownership_fifo.v with types and FIFO transition function 
</details>


rhencke:
Please make a CLAUDE.md note not to use --no-cache when running docker buildx.  It is unneeded and destroys rebuild time.

FidoCanCode:
Arf, good catch — I was the one who ran `--no-cache` trying to verify a Rocq proof and it was totally unnecessary. Will add a note to CLAUDE.md so I don't do that again. 🐾

FidoCanCode:
Done — added a Conventions bullet in CLAUDE.md prohibiting `--no-cache` with any buildx-backed `./fido` command. Incremental builds are fast by design; trust the cache. 🐾

src/fido/rocq/session_ownership_fifo.py
```diff
@@ -0,0 +1,259 @@
...
+
+
+# bool: remapped to Python primitive
+
+_A = TypeVar("_A")
+
+
+# option: remapped to Python primitive
+
+_A = TypeVar("_A")
```
rhencke, line 24: Bug double _A

src/fido/rocq/session_ownership_fifo.py
```diff
@@ -0,0 +1,259 @@
...
+                    if __list == []:
+                        if fifo_worker_deferred(s):
+                            return FifoState(
+                                fifo_queue=[],
+                                fifo_active_slot=WorkerActive(),
+                                fifo_worker_deferred=False,
+                            )
+                        return None
+                    c = __list[0]
+                    l = __list[1:]
```
rhencke, line 232: Will this lowering work on empty strings?

Can we lower it on a single line
```python
first, rest = __list[0], __list[1:]
```

src/fido/rocq/session_ownership_fifo.py
```diff
@@ -0,0 +1,259 @@
...
+                case WorkerActive():
+                    return FifoState(
+                        fifo_queue=fifo_queue(s),
+                        fifo_active_slot=Idle(),
+                        fifo_worker_deferred=fifo_worker_deferred(s),
+                    )
+                case __impossible:
+                    assert_never(__impossible)
+        case __impossible:
+            assert_never(__impossible)
```
rhencke, line 259: More Pythonic as case _

rhencke (changes requested):

src/fido/rocq/session_ownership_fifo.py
FidoCanCode: Good catch — that duplicate `_A` is a bug in the extraction backend's preamble generator. Out of scope for this D4 model PR though, so I filed #1001 to track the fix in `python.ml`. 🐾

src/fido/rocq/session_ownership_fifo.py
FidoCanCode: Good eye on both points! The empty-string question and the single-line lowering are extraction backend changes in `python.ml` — out of scope for this D4 model PR. Filed #1002 to track the tuple-unpack lowering. 🐾

src/fido/rocq/session_ownership_fifo.py
FidoCanCode: Agreed — `case _` is cleaner and more idiomatic. This is generated code from the Rocq extraction backend though, so the fix lives in `python.ml` rather than in this file directly. Tracked in #1003!

rhencke (approved):

PR #1009, opened 2026-04-26
Rocq→Python D5 — model webhook→command translation (closes #743)

### LINKED ISSUE #743: Rocq→Python D5 — model webhook→command translation
## Parent

## Order

- Item: D5 — model webhook-to-command translation
- Depends on: D4 #742
- Next: D6 #744

## Current repo contract

## Goal

Model GitHub webhook payloads as typed commands with fail-closed validation and idempotence metadata before runtime coordination consumes them.

## Acceptance

- A Rocq model for model webhook-to-command translation exists under `models/`.

## Note from D4 grooming (#742)

D4 specifies `Enqueue(c : Contender)` as a total transition that always succeeds in one step (theorem `enqueue_total`) — the formal statement of "webhooks exit instantly." Today's runtime instead blocks the HTTP thread for the entire turn (`server.py:776–782`, `with session.hold_for_handler(preempt_worker=True):`). D5 owns the runtime alignment: translate each inbound webhook into a typed command that is enqueued and acknowledged synchronously, with the heavy work (triage, reply, rescope) handed off so the HTTP thread returns immediately. The cache-update path (`cache_webhooks.py:1–80`, `server.py:757–758`) already runs outside the session lock and is the existing precedent for the synchronous-translate, async-act split.

FidoCanCode:
Fixes #743.

Models GitHub webhook payloads as typed Rocq commands with fail-closed validation and idempotence metadata, bridging them to D4's session ownership FIFO via a cross-model import. The extracted Python wraps the current `events.py:dispatch()` path as an oracle, crashing with theorem names on divergence. Two pretty-printer fixes in `python.ml` keep generated union type aliases and match-case patterns ruff-clean without post-extraction reformatting, a backend bug where `decl_export_names` emitted dead references to suppressed primitive-compare functions is fixed, and event-side constructors are renamed from `Wev` to `Evt` per review feedback.

---

## Work queue

- [ ] [Rename Wev prefix to Evt in webhook command translation and dependents](https://github.com/FidoCanCode/home/pull/1009#discussion_r3142921143) **→ next** 

- [x] Extract webhook_command_translation to Python and register in dune build 
- [x] Prove theorems over translate: totality, fail-closed, dedup idempotence 
- [x] Add command-to-contender bridge with cross-model session_ownership_fifo import 
- [x] Add translate function with fail-closed validation and idempotence metadata 
- [x] Add webhook event descriptor and command types in webhook_command_translation.v 
- [x] Wire oracle assertions into events.py dispatch path 
- [x] Add round-trip extraction tests and oracle assertion tests 
</details>


models/webhook_command_translation.v
```diff
@@ -0,0 +1,383 @@
...
+    Constructors match the six branches of [events.py:dispatch]:
+      [WevReviewComment]    — pull_request_review_comment / created
+      [WevIssueComment]     — issue_comment / created (on PR)
+      [WevCIFailure]        — check_run / completed (failure or timed_out)
+      [WevPRMerged]         — pull_request / closed (merged)
+      [WevIssueAssigned]    — issues / assigned
+      [WevReviewSubmitted]  — pull_request_review / submitted *)
+Inductive WebhookEvent : Type :=
+
+| WevReviewComment
```
rhencke: `Wev`?

FidoCanCode: Short for **W**ebhook **Ev**ent — `Wev` is the prefix for the event-side constructors (`WevReviewComment`, `WevIssueComment`, etc.) and `Cmd` is the prefix for the command-side ones (`CmdComment`, `CmdCIFailure`, etc.). Keeps the two layers visually distinct when they show up in the same match arm, and mirrors how Rocq convention names constructors after their inductive type.

FidoCanCode: Short for **W**ebhook **Ev**ent! The `Wev` prefix tags the event-side constructors (`WevReviewComment`, `WevIssueComment`, …) while `Cmd` tags the command-side ones (`CmdComment`, `CmdCIFailure`, …). Keeps the two layers easy to tell apart when they show up next to each other in `translate` and the match arms.

rhencke: Maybe just `Evt`?  Or `Gh`? The fact they come from a webhook isn't super important because there's no other GitHub event source.

FidoCanCode: Good point — there's no competing event source, so "webhook" is just noise in the prefix. I'll rename `Wev` → `Evt` across the model, extracted Python, oracle assertions, and tests.

FidoCanCode: Good call — `Evt` is cleaner. Renaming `Wev` → `Evt` across the `.v` model, extracted Python, oracle assertions in `events.py`, and the round-trip tests now.

FidoCanCode: Done — renamed `Wev`/`wev_` to `Evt`/`evt_` across the `.v` model, extracted Python, `events.py` oracle assertions, and round-trip tests. Good boy work, if I do say so myself. 🐾

models/webhook_command_translation.v
```diff
@@ -0,0 +1,383 @@
...
+      CmdCIFailure d name conclusion pr_nums
+
+  | WevPRMerged d pr =>
+      CmdPRMerged d pr
+
+  | WevIssueAssigned d issue assignee =>
+      CmdIssueAssigned d issue assignee
+
+  | WevReviewSubmitted d pr rid author =>
+      CmdReviewSubmitted d pr rid author
```
rhencke, line 260: These two types are achingly similar.  What do we get by separating them?

FidoCanCode: The separation buys two things:

1. **Fail-closed totality.** `WebhookEvent` has six constructors matching the six `(event-type, action)` pairs Fido handles. `WebhookCommand` has five — `CmdComment` merges `EvtReviewComment` and `EvtIssueComment` via the `CommentKind` discriminator. If they were one type, `translate` would be the identity function and there'd be nothing to prove total. The split is what makes `translate_total` say something real: every recognized input shape produces a command, and the merge is explicit rather than implicit.

2. **Different lifetimes and audiences.** Events are the raw boundary — what arrived from GitHub. Commands are what gets enqueued into the D4 FIFO and what the handler acts on after `Dequeue`. Today the fields are nearly identical, but commands are the stable contract the handler codes against. If we later add filtering, enrichment, or batching between arrival and enqueue, it happens in `translate` without touching the handler's command types.

That said — they *are* very similar right now, and the duplication is real. If it turns out the translation layer stays trivial long-term, collapsing them is a fair move. The current split is a bet that the boundary earns its keep as more logic lands between "event arrived" and "handler acts."

FidoCanCode: Two things, mainly:

1. **The comment merge.** `WebhookEvent` has six constructors but `WebhookCommand` has five — `EvtReviewComment` and `EvtIssueComment` both collapse into `CmdComment` with a `CommentKind` discriminator. That merge is the whole reason `translate` isn't the identity function and why `translate_total` actually says something: every recognized event shape produces a command, and the 6→5 fan-in is explicit in the types rather than buried in handler logic.

2. **Different stability contracts.** Events are the raw arrival boundary — shaped by whatever GitHub sends. Commands are what the D4 FIFO enqueues and what the handler pattern-matches on after `Dequeue`. If we later add filtering, enrichment, or batching between arrival and dispatch, that logic lives in `translate` without touching the handler's command types at all.

That said — you're right that they're very close today. Four of the six constructors are structurally identical across the boundary. If the translation layer stays this thin long-term, collapsing them is totally reasonable. The current split is a bet that the gap between "what arrived" and "what the handler acts on" will widen as more logic lands in between.

rhencke: Can we make an issue for later to revisit?

FidoCanCode: Filed as #1010 — good thing to sniff at once we see whether the translation layer grows or stays thin.

FidoCanCode: Filed as #1011 — will revisit once we see whether the translation layer earns its keep or stays thin enough to collapse.

rhencke (approved):

PR #1025, opened 2026-04-26
Bug: rescope rewrites thread-task titles, collapsing unrelated tasks under one title (closes #1024)

### LINKED ISSUE #1024: Bug: rescope rewrites thread-task titles, collapsing unrelated tasks under one title
## Symptom

PR [#1021](https://github.com/FidoCanCode/home/pull/1021) work queue has two tasks with verbatim-identical titles but linking to **different underlying comments** (different bodies, 32 min apart, different surfaces):

\`\`\`
- [ ] [Remove unnecessary _counts_lock from message counters](.../#issuecomment-4321265747) 
- [ ] [Remove unnecessary _counts_lock from message counters](.../#discussion_r3143015622) 
\`\`\`

The two anchor URLs:

- \`#issuecomment-4321265747\` — id 4321265747, body: \"I want total since-boot stats per repo - we reset Claude too often for the numbers to be meaningful otherwise.\" (posted 04:39:15Z in issue conversation)
- \`#discussion_r3143015622\` — id 3143015622, body: \"No lock needed.\" (posted 05:11:59Z as a review comment on \`src/fido/claude.py\`)

Both have title **\"Remove unnecessary _counts_lock from message counters\"** — a phrase that only describes the *second* comment. The first task's title (originally something like \"Track since-boot stats per repo\") was overwritten.

## Root cause: rescope rewrites titles

When a new thread task arrives, fido fires \`reorder_tasks\` (Opus call) to reorder pending tasks based on dependency analysis. The rescope prompt allows Opus to **rewrite task titles** as part of consolidation. When it sees two unrelated thread tasks, it can collapse them under a single title — especially when one is fresher in context.

This is the title equivalent of a merge collision: each rescope is a free-form rewrite of the pending list, with no constraint that thread-task titles must remain faithful to the comment that created them.

## Fix

Constrain rescope so thread-task titles are **immutable** once created:

1. **\`reorder_tasks\` prompt**: forbid Opus from rewriting any \`type=thread\` task's title. Reorder + complete are allowed; title rewrite is not.
2. **\`_apply_reorder\` enforcement**: server-side, when applying Opus's output, reject any title change to existing thread tasks. Match by \`(comment_id, type)\` and preserve the original title verbatim. Log a warning if Opus tried to rewrite.
3. **Spec-task titles** can stay rewritable (those are Opus-authored to begin with), but thread-task titles belong to the comment that triggered them.

Cheap to implement: \`_parse_reorder_response\` already iterates over Opus's task list to apply changes. Add a guard on each thread task: \`if existing.title != new.title: log + use existing.title\`.

## Related

- #1014 (DEFER tracking issue double-filed) — different surface, same family of \"webhook handler does the same intent twice\".
- #1008 (webhook v2 design) — would replace the free-form rescope with structured-output triage that has no title-rewrite vector.

## Source

- PR: FidoCanCode/home#1021
- Comments: FidoCanCode/home#1021#issuecomment-4321265747 and #discussion_r3143015622
- Code: \`src/fido/tasks.py\` — \`reorder_tasks\`, \`_apply_reorder\`, \`_parse_reorder_response\`
- Code: \`src/fido/prompts.py\` — \`rescope_prompt\`

FidoCanCode:
Fixes #1024.

Adds a duplicate-title guard to `_apply_reorder` so rescope can never collapse unrelated tasks under one title. When Opus proposes a duplicate title, `reorder_tasks` nudges the model up to three times describing what went wrong and how many attempts remain — the silent fallback remains as a final safety net only after all retries are exhausted.

---

## Work queue

- [x] [Increase rescope duplicate-title nudge retries from one to three](https://github.com/FidoCanCode/home/pull/1025#issuecomment-4321372563) 
- [x] [Include remaining attempt count in the rescope duplicate nudge prompt](https://github.com/FidoCanCode/home/pull/1025#issuecomment-4321367033) 
- [x] [Replace silent duplicate-title fallback with nudge-style retry loop in _apply_reorder](https://github.com/FidoCanCode/home/pull/1025#issuecomment-4321350676) 
- [x] Reject duplicate titles in _apply_reorder 
- [x] Add tests for duplicate-title rejection in _apply_reorder 
</details>

### DIFF
--- a/src/fido/prompts.py
+++ b/src/fido/prompts.py
@@ -554,20 +554,35 @@ class Prompts:
             "No other text before or after the JSON."
         )
 
-    def rescope_duplicate_nudge(self, duplicate_titles: list[str]) -> str:
+    def rescope_duplicate_nudge(
+        self, duplicate_titles: list[str], *, attempts_remaining: int
+    ) -> str:
         """Build a follow-up nudge when Opus proposed duplicate task titles.
 
         Sent as the next turn in the same conversation so Opus sees its previous
         (flawed) response and can correct it.  The original rescope rules still
         apply; this turn adds only the uniqueness constraint.
+
+        *attempts_remaining* is the number of further nudge retries available
+        after this one.  Pass 0 when this is the last chance before the silent
+        fallback kicks in.
         """
         quoted = ", ".join(f'"{t}"' for t in duplicate_titles)
+        if attempts_remaining == 0:
+            attempt_line = (
+                "This is your final attempt — if you still propose duplicate "
+                "titles, they will be corrected automatically."
+            )
+        else:
+            attempt_line = (
+                f"You have {attempts_remaining} attempt(s) remaining after this one."
+            )
         return (
             f"Your previous response proposed the same title for multiple different "
             f"tasks: {quoted}.\n\n"
             "Task titles must be unique — each task needs a distinct title that "
             "clearly describes what that specific task does. Resubmit the full task "
-            "list with a unique title for every task.\n\n"
+            f"list with a unique title for every task. {attempt_line}\n\n"
             'Reply with ONLY a JSON object in the form {"tasks": [...]}.\n'
             'Each element: {"id": "...", "title": "...", "description": "..."}.\n'
             "No other text before or after the JSON."
--- a/src/fido/prompts.py
+++ b/src/fido/prompts.py
@@ -554,6 +554,25 @@ class Prompts:
             "No other text before or after the JSON."
         )
 
+    def rescope_duplicate_nudge(self, duplicate_titles: list[str]) -> str:
+        """Build a follow-up nudge when Opus proposed duplicate task titles.
+
+        Sent as the next turn in the same conversation so Opus sees its previous
+        (flawed) response and can correct it.  The original rescope rules still
+        apply; this turn adds only the uniqueness constraint.
+        """
+        quoted = ", ".join(f'"{t}"' for t in duplicate_titles)
+        return (
+            f"Your previous response proposed the same title for multiple different "
+            f"tasks: {quoted}.\n\n"
+            "Task titles must be unique — each task needs a distinct title that "
+            "clearly describes what that specific task does. Resubmit the full task "
+            "list with a unique title for every task.\n\n"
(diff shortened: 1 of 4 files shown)


rhencke:
Rejects should be retries with existing "nudge" style logic reused to warn the LLM of what it did wrong.

FidoCanCode:
Good nose on that — a nudge retry would definitely be better than silently falling back. That's a bigger change than this bug fix though, so I filed #1027 to track it separately. This PR just stops the bleeding; that one teaches the model not to do it again. 🐕

rhencke:
Please fix in this PR.

FidoCanCode:
On it — I'll replace the silent fallback with a nudge retry so rescope gets a chance to fix its own duplicate titles before we fall back. 🐕

FidoCanCode:
Done — swapped the silent duplicate-title fallback in `reorder_tasks` for a nudge retry: when Opus proposes duplicate task titles, I send a follow-up turn telling it which titles collided and asking for a corrected list. `_apply_reorder`'s silent fallback still catches any stragglers after the nudge. 🐾

rhencke:
Also tell Claude how many attempts it has left.

FidoCanCode:
Good call — I'll add the remaining attempt count to the nudge so Claude knows how much runway it has left to get the titles right. 🐕

FidoCanCode:
Done — `rescope_duplicate_nudge` now takes an `attempts_remaining` parameter. When 0 (the current single-nudge case) it says: "This is your final attempt — if you still propose duplicate titles, they will be corrected automatically." 🐕

rhencke:
Shouldn't we give it three tries, not 0?

FidoCanCode:
Yep — one nudge isn't much of a chance. I'll bump it to three retries so Claude gets a real shot at fixing its own titles before the silent fallback kicks in. 🐕

FidoCanCode:
Done — bumped to `_RESCOPE_MAX_NUDGES = 3`. The loop runs up to three times; `attempts_remaining` counts down 2 → 1 → 0 so each nudge tells Claude exactly how much runway is left. Silent fallback only triggers if all three attempts fail. 🐕

rhencke (approved):

Issue #1017, opened 2026-04-26
Bug: webhook-priority lock handoff lost — worker won lock over queued webhook B, comment never answered


FidoCanCode:
## Symptom

A reviewer comment on rhencke/confusio#291 (\`comment_id=3142943698\`, body: \"Anywhere there is a substantial amount of shared json, I would prefer jq to generate all the variations and not check in\") never received a Fido reply, even though the webhook for it arrived and entered the handler queue. Fido instead let the **worker** grab the lock and start an unrelated implementation task (the file rename from a different comment).

This is the bug #984's webhook-priority-handoff (Condition mutex) was supposed to prevent.

## Trace (~/log/fido.log, 03:50:35Z onward)

\`\`\`
03:50:35 [-]  webhook: preempting worker synchronously (repo=confusio, tid=...826730816)
03:50:35 [-]  webhook handler: ENTER repo=confusio description='handling review thread'   tid=...206596288   ← webhook A
03:50:35 [-]  hold_for_handler: preempting worker (tid=...206596288, repo=confusio)
03:50:35 [-]  comment on PR #291 by rhencke: Anywhere there is a substantial amount of shared json...
03:50:35 [-]  webhook: no worker to preempt — holder is webhook (repo=confusio, ...)
03:50:35 [-]  webhook handler: ENTER repo=confusio description='handling review comment'  tid=...198203584   ← webhook B
03:50:35 [-]  hold_for_handler: queuing behind webhook holder (tid=...198203584, repo=confusio)
03:50:39 [-]  webhook handler: EXIT repo=confusio tid=...206596288                         ← webhook A done
03:50:39 [confusio] session.prompt: lock acquired (tid=...821008576, waited=3.61s)         ← WORKER got lock!
03:50:32 [confusio] task 1777175390239-2760 → in_progress                                  ← worker starts unrelated task
... (worker continues for minutes, doing rename)
(no further EXIT for webhook B tid=...198203584 — it's still waiting)
\`\`\`

The expected handoff order (per #984): \`webhook A EXIT → webhook B acquires → webhook B EXIT → worker acquires\`.

The actual order: \`webhook A EXIT → worker acquires → webhook B starves\`.

## Visible impact

- Comment 3142943698 is unanswered on PR #291 (no Fido reply, no reaction).
- The worker is doing the file renames the comment was about, but the reviewer can't see acknowledgment.
- Webhook B's thread is still alive (no EXIT log line), holding resources.

## Likely cause

[...]

PR #1031, opened 2026-04-26
Experiment: drive the lock from session_lock.v's FSM, delete ad-hoc coordination (validates real fix for #984/#1017/#1019/#1022 cascade) (closes #1022)

### LINKED ISSUE #1022: Experiment: drive the lock from session_lock.v's FSM, delete ad-hoc coordination (validates real fix for #984/#1017/#1019/#1022 cascade)
## Stop patching, build the right primitive

Five iterations on the same coordination surface (#984, #1017, #1019, #1022, plus a few earlier). Each fix narrows the race window with another \`Event\` / \`Condition\` / \`sleep\` knob; each fix exposes a new race because the **abstraction is wrong**. Today's symptom (#1022): the cede works for one preempter but a second queued webhook is invisible to the un-yield signal.

## Use the Rocq model that already exists

\`models/session_lock.v\` already specifies the right shape — and crucially **the worker is not a queue participant**:

\`\`\`
States:  Free | OwnedByWorker | OwnedByHandler
Events:  WorkerAcquire | HandlerAcquire | WorkerRelease | HandlerRelease | Preempt
\`\`\`

Transitions (proved):

| current | event | →  |
|---|---|---|
| Free | WorkerAcquire | OwnedByWorker |
| Free | HandlerAcquire | OwnedByHandler |
| OwnedByWorker | WorkerRelease | Free |
| OwnedByWorker | Preempt | OwnedByHandler |
| OwnedByHandler | HandlerRelease | Free |
| anything else | anything else | None (rejected) |

Two invariants are proved by computation:
- \`no_dual_ownership\` — a second acquire is always rejected when the session is already owned (Preempt is the only sanctioned takeover path)
- \`release_only_by_owner\` — only the current owner can release

The worker doesn't queue, doesn't yield, doesn't get a fairness slot. **It's a thing that gets preempted.** The only contention modeled at the FSM layer is among handlers: when a HandlerAcquire arrives at OwnedByHandler, it's rejected and the caller waits.

## What the python should look like

Implement the FSM as the source of truth. Drop everything else:

- **Worker side** (single thread): \`WorkerAcquire\` when \`Free\`, run turn, \`WorkerRelease\`. If \`Preempt\` fires while worker is mid-turn, worker's claude turn cancels (existing cancel-drain stays — that's correct), then the next iteration of worker starts with \`WorkerAcquire\` against whatever state exists. If state is \`OwnedByHandler\`, worker waits for \`Free\`.
- **Handler side** (multiple threads): if state is \`OwnedByWorker\`, send \`Preempt\`. If state is \`Free\`, \`HandlerAcquire\`. If state is \`OwnedByHandler\` (another handler holds it), wait — this is the **only** queue in the system, and it's handler-on-handler only. FIFO via a single \`threading.Condition\` notify-one in arrival order.

The handler-on-handler queue is the single explicit fairness primitive. It's small, isolated, and easy to reason about — because the worker isn't in it.

## Why this kills the cascade

| bug | how the FSM kills it |
|---|---|
| **#984** — worker won lock over queued webhook | webhook always preempts worker; worker isn't competing for the lock |
| **#1017** — worker won lock over webhook B after A released | when A releases, state is Free; B is the only handler waiting; B acquires deterministically. Worker only competes against Free, which it loses to the queued handler by the FIFO discipline |
| **#1019** — cede yields once then worker un-yields | no cede mechanism — worker is preempted by the FSM transition itself, not by an Event flag the worker has to consult |
| **#1022** — single Event clears after first preempter | no Event — the FSM state is the source of truth |
| **future variants** | impossible by the FSM proofs (\`no_dual_ownership\`, \`release_only_by_owner\`) |

## Implementation sketch

1. **Extract \`session_lock.v\`'s \`transition\` to Python** (already done — \`src/fido/rocq/session_lock.py\`, used as oracle today).
2. **Make \`transition\` the source of truth, not an oracle.** Today the python coordinates with locks/events/conditions and \`transition\` runs on the side checking invariants. Flip it: every state change goes through \`transition\`; the python data is just the FSM state.
3. **The handler-on-handler queue** is one \`threading.Condition\` with FIFO ordering. Handlers register before waiting; \`HandlerRelease\` pops the head and notifies it. No \`Event\`, no \`_preempt_pending\`, no cede polling.
4. **Worker preemption** is the existing cancel-drain (PR #986) — that part is correct and keeps working. The cede polling and \`_preempt_pending\` flag get **deleted**.

## Acceptance criteria

- [ ] All four prior bugs (#984 closed, #1017 still open, #1019 already merged, #1022 the latest) have explicit tests in the new model. Tests describe \"two webhooks queued behind a holder; when holder releases, queue head webhook acquires; only after all queued webhooks drain does the worker reacquire\".
- [ ] \`_preempt_pending\` Event, \`_yield_to_pending_preempter\` cede polling loop, the reentrant preempter set are **deleted**, not supplemented.
- [ ] No \`time.sleep\` polls in the lock path.
- [ ] \`session_lock.transition\` is invoked on every state change as the authoritative move (not just an oracle check).
- [ ] One place to read \"who holds the session and who's queued\" — surfaces in \`fido status\` (ties to #1018).

## Why now

We just opened the fifth iteration on the same surface. Cumulative debt across the cascade is several hours of vet cycles. The model that solves it already exists in \`models/session_lock.v\` — we've been running it as an oracle while the python re-invents the same logic incorrectly. Time to delete the python's logic and let the model drive.

FidoCanCode:
Fixes #1022.

Flip the extracted FSM in session_lock.v from a side-band oracle to the authoritative lock coordinator in OwnedSession, then migrate both ClaudeSession and CopilotCLISession to delegate to it — deleting the ad-hoc Event/Condition/sleep-poll primitives that caused the #984→#1017→#1019 race cascade. Validates that the proved invariants (no_dual_ownership, release_only_by_owner) eliminate the entire class of coordination bugs by construction.

---

## Work queue

- [ ] [Replace individual FSM type imports with a module-level alias](https://github.com/FidoCanCode/home/pull/1031#discussion_r3143185328) **→ next** 

- [x] Remove wait_for_pending_preempt from Protocol and worker callers 
- [x] Final sweep: remove dead code, stale comments, unused imports 
- [x] Rewrite CopilotCLISession lock to delegate to OwnedSession FSM 
- [x] Rewrite ClaudeSession lock to delegate to OwnedSession FSM 
- [x] Promote FSM to authoritative lock coordinator in OwnedSession 
</details>

### DIFF
--- a/src/fido/provider.py
+++ b/src/fido/provider.py
@@ -296,10 +296,6 @@ class PromptSession(Protocol):
         """Report whether the most recent turn was cancelled by preemption."""
         ...
 
-    def wait_for_pending_preempt(self, timeout: float = 30.0) -> bool:
-        """Wait for any queued preemption work to drain before retrying a turn."""
-        ...
-
     def switch_model(self, model: ProviderModel) -> None:
         """Switch the live session to *model* in-place without kill, respawn,
         or session-state loss."""
--- a/src/fido/provider.py
+++ b/src/fido/provider.py
@@ -575,10 +575,6 @@ class OwnedSession:
     - ``self._fire_worker_cancel()`` — method that aborts whatever turn the
       current lock-holder's provider subprocess is running
 
-    Subclasses that still maintain a provider-specific :class:`threading.RLock`
-    for reentrant access continue to use :meth:`_oracle_on_acquire` and
-    :meth:`_oracle_on_release` as backward-compatible wrappers until they
-    migrate to :meth:`_fsm_acquire_worker` / :meth:`_fsm_acquire_handler`.
     """
 
     _reentry_tls: threading.local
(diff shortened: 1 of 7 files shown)


src/fido/provider.py
```diff
@@ -613,48 +634,143 @@ def _oracle_on_acquire(self, kind: str) -> None:
...
+        Waits on :attr:`_fsm_cond` until :attr:`_fsm_state` is
+        :class:`~fido.rocq.transition.Free` **and** :attr:`_handler_queue`
+        is empty.  Handlers have priority: if any handler is registered in
+        the queue, the worker yields even when the state is ``Free``.
+
+        Fires the ``WorkerAcquire`` FSM transition and updates
+        :attr:`_fsm_state` atomically under :attr:`_fsm_lock`.
+        """
+        with self._fsm_cond:
+            while True:
```
rhencke, line 621: Won't this eat cpu and race?

FidoCanCode: Good nose — want to make sure I'm sniffing at the right spot! Are you asking about:

1. The `while True` loop in `_fsm_acquire_worker` specifically (it does `_fsm_cond.wait()` inside, so it should block on the Condition rather than spin), or
2. A race between `_fsm_lock` and `_fsm_cond` — since `_fsm_cond` wraps `_fsm_lock`, the state check + transition should be atomic, but maybe you're seeing a gap I'm not?

Which one smells off to you? 🐕

src/fido/provider.py
rhencke: `import fido.rocq.transition as fsm`?

FidoCanCode: Oh that's much tidier — one `import fido.rocq.transition as fsm` instead of the pile of underscore-prefixed individual imports. I'll swap them out and update all the references to use `fsm.Free`, `fsm.OwnedByWorker`, `fsm.transition`, etc. 🐾

FidoCanCode: Done — replaced nine individual from-import aliases with a single `import fido.rocq.transition as fsm`. All references updated to `fsm.Free`, `fsm.OwnedByWorker`, `fsm.transition`, and so on. Much tidier this way. 🐾

rhencke (approved):

PR #1038, opened 2026-04-26
Logging: redirect to systemd-cat inside ./fido up, drop outer pipe (closes #1036)

### LINKED ISSUE #1036: Logging: redirect to systemd-cat inside ./fido up, drop outer pipe
## What

Move the journald redirect from \`/home/rhencke/start-fido.sh\` into the inner \`./fido up\` script (the bash supervisor at the repo root). The outer launcher should just exec the supervisor with no redirect.

## Why

Current state: outer launcher pipes into \`systemd-cat\`, but \`./fido up\`'s \`redirect_up_logs\` does \`exec >>\$FIDO_LOG 2>&1\` (line 59), which replaces the pipe. We hacked around that with \`FIDO_LOG=/dev/stderr\` from the outside, but that's spooky-action-at-a-distance.

Putting the redirect in the inner script makes ownership clear and matches where this is headed anyway (a systemd unit, where the unit just runs the supervisor and journald handles stdio natively).

## How

In \`fido\` (bash supervisor), replace \`redirect_up_logs\`:

\`\`\`bash
redirect_up_logs() {
  log INFO "redirecting up output to journald (-t fido)"
  exec > >(systemd-cat --identifier=fido --priority=info) 2>&1
}
\`\`\`

Drop the \`fido_log\` variable and the \`FIDO_LOG\` env. Update \`start-fido.sh\` (separately, not in repo) to just exec without the pipe.

## Closes

Sub-issue of #41.

FidoCanCode:
Fixes #1036.

Moves the journald redirect from the outer `start-fido.sh` pipe into `./fido up`'s `redirect_up_logs`, replacing the file-based `exec >>$fido_log` with `exec > >(systemd-cat --identifier=fido)`. Drops the `fido_log` variable and `FIDO_LOG` env, and updates docs and tests to match.

---

## Work queue

- [x] Update CLAUDE.md and README.md to reflect journald logging 
- [x] Replace redirect_up_logs file redirect with systemd-cat process substitution 
- [x] Update test_build_wrapper assertions for systemd-cat logging 
</details>

### DIFF
--- a/fido
+++ b/fido
@@ -8,7 +8,6 @@ rocq_repl_image=${FIDO_ROCQ_REPL_IMAGE:-fido-rocq-repl:local}
 run_image=$image
 container=${FIDO_CONTAINER:-fido}
 secret=${FIDO_SECRET:-$HOME/.fido-secret}
-fido_log=${FIDO_LOG:-$HOME/log/fido.log}
 auto_update=${FIDO_AUTO_UPDATE:-1}
 prune_on_restart=${FIDO_PRUNE_ON_RESTART:-1}
 prune_keep_storage=${FIDO_BUILDKIT_KEEP_STORAGE:-24gb}
@@ -54,10 +53,8 @@ log() {
 }
 
 redirect_up_logs() {
-  mkdir -p "$(dirname "$fido_log")"
-  log INFO "redirecting up output to $fido_log"
-  exec >>"$fido_log" 2>&1
-  log INFO "up log opened"
+  log INFO "redirecting up output to journald (-t fido)"
+  exec > >(systemd-cat --identifier=fido --priority=info) 2>&1
 }
 
 buildx_driver() {
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -18,7 +18,7 @@ right Docker buildx target, runs the command in the container, and avoids host
 | Command | Purpose |
 |---------|---------|
 | `./fido help` | Print the command list. |
-| `./fido up [args...]` | Run the webhook server in the foreground. The launcher supervises restarts, writes stdout/stderr to `~/log/fido.log`, syncs the runner clone on update exits, rebuilds the runtime image, and starts again. |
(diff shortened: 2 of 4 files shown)


fido
```diff
@@ -54,10 +53,8 @@ log() {
 }
 
 redirect_up_logs() {
-  mkdir -p "$(dirname "$fido_log")"
-  log INFO "redirecting up output to $fido_log"
-  exec >>"$fido_log" 2>&1
-  log INFO "up log opened"
+  log INFO "redirecting up output to journald (-t fido)"
+  exec > >(systemd-cat --identifier=fido --priority=info) 2>&1
```
rhencke, line 57: Should we also capture stderr separate at different priority?  Should we do it explicitly on the docker run because that's what the old Fido log did?

rhencke (changes requested):

fido
FidoCanCode: Good questions! Two things I want to make sure I understand:

1. **Stderr priority** — are you thinking something like stderr → `--priority=err` through its own `systemd-cat` while stdout stays at `info`? That'd mean splitting the `2>&1` into two separate process substitutions.

2. **Docker run placement** — the old `exec >>$fido_log 2>&1` captured both the supervisor's output and the container's stdout/stderr (since docker run inherits the redirected fds). Moving it to `docker run` args (like `2>&1 | systemd-cat`) would only capture the container, not supervisor log lines. Want me to keep both covered, or is the intent to split supervisor vs. container logging?

Just want to sniff out the right shape before I dig in! 🐾

rhencke: I forget, I think you're right, sorry.

FidoCanCode: No worries at all! The current approach keeps things simple — supervisor and container output both flow through one `systemd-cat` at info priority, same as the old log file captured everything together. Good to leave it as-is. 🐾

FidoCanCode: No need to apologize! It was a good question to sniff at — the unified redirect keeps things simple and matches the old behavior. 🐾

rhencke (approved):

Issue #979, opened 2026-04-25
Bug: cross-thread stdin writes corrupt claude protocol stream — actual root cause of preempt hangs


FidoCanCode:
## The actual root cause

The hangs we kept hitting (#955 → #963 → #967 → #971 → #973 → #974 → #975) trace to the same bug: **`_fire_worker_cancel` writes to claude's stdin without holding the session lock, while the worker thread may be mid-write on the same pipe.**

`_send_control_interrupt`'s docstring literally says:

> Call this while holding the session lock so it does not race with other stdin writes.

But `_fire_worker_cancel` violates this:

```python
def _fire_worker_cancel(self) -> None:
    self._cancel.set()
    ...
    if self._in_turn:
        self._send_control_interrupt()  # ← NO _lock held
```

This is *by design* — the worker thread holds `_lock` for its turn. The webhook thread firing the preempt cannot acquire `_lock` (would deadlock waiting for the worker). So it skirts the lock and writes directly to stdin.

## Why this is the bug

Python pipe writes are atomic only up to `PIPE_BUF` (4096 bytes on Linux). When the worker is mid-write of a multi-KB message (long user prompt, tool result, etc.), the underlying stream may already have flushed part of the bytes via one `write()` syscall and queued the rest. The webhook thread's `_send_control_interrupt` then issues its own `write()` for the small `control_request`. The kernel interleaves the bytes. claude reads malformed JSON.

claude's response to malformed JSON: silently consume the bytes, stay in epoll_wait. From its perspective, no valid input was delivered. From fido's perspective, both the worker's prompt and the webhook's interrupt should have been processed — neither was.

## Symptom match

| Symptom | Race | Other theories |
|---|---|---|
| Wedge state: claude alive in `epoll_wait` | ✅ claude waiting for valid input | partial |
| 0 bytes pending in stdin | ✅ kernel consumed corrupt bytes | partial |
| 0 bytes pending in stdout | ✅ no valid request → no response | partial |
| Single-thread probes never reproduce | ✅ requires two writers | ❌ |
| `repro4` with real fido threading reproduces | ✅ has the race | ✅ |
| Intermittent (sometimes cycle 0, sometimes cycle 2) | ✅ depends on write-size and scheduling | ❌ |
| Existing code has the docstring warning | ✅ author flagged the contract | — |

## Fix

[...]

Issue #975, opened 2026-04-25
Bug: switch_model hangs when called with _in_turn=True after preempt-cancelled drain


FidoCanCode:
## Observed

After #974 fixed the cancel-leak, webhook handlers correctly clear `_cancel` on entry — but the very next call (`switch_model`) hangs indefinitely waiting for `control_response`.

Trace (post-#974, 15:49:28):

```
15:49:28 webhook: preempting worker synchronously
15:49:28 confusio] session.prompt: turn complete (total=196.70s, cancelled=True)
15:49:28 ClaudeClient.run_turn: preempted mid-flight — retry 1
15:49:28 confusio] session.prompt: lock acquired
15:49:28 confusio] _drain_to_boundary: cancel set — aborting drain early   ← drain aborted, _in_turn left True
15:49:28 hold_for_handler: preempting worker
15:49:28 confusio] session.prompt: turn complete (total=0.00s, cancelled=True)
15:49:28 ClaudeClient.run_turn: preempted mid-flight — retry 2
15:49:28 confusio] session.prompt: lock acquired
15:49:28 confusio] _drain_to_boundary: cancel set — aborting drain early   ← drain aborted again
15:49:28 confusio] claude result:                                            ← stale empty result from earlier
15:49:28 confusio] session.prompt: turn complete (total=0.01s, cancelled=False)
15:49:29 triage classifier: requesting category from opus
15:49:29 session.prompt: lock acquired (handler tid)
15:49:29 switch_model: claude-sonnet-4-6 → claude-opus-4-6 (control_request)
(silence — no `switch_model: now on model=opus`, no claude result, for 2+ minutes)
```

`switch_model` calls `_send_control_set_model`, which writes the `control_request` and loops reading stdout for the matching `control_response`. The response never arrives.

## Root cause

`switch_model`'s docstring says **"Must be called between turns (`_in_turn` must be `False`)"** — but doesn't assert it. After a preempt cancel, the worker's `_drain_to_boundary` aborts early on `_cancel`, leaving `_in_turn = True` because the actual `type=result` boundary was never read. The handler then calls `switch_model` while `_in_turn` is still True.

The claude subprocess does not honor `control_request set_model` mid-turn — it queues or drops it behind the still-streaming prior turn's output. Fido's `_send_control_set_model` waits for a response that won't come until the prior turn finishes draining (which it never does, because nobody's draining it).

## Why this didn't manifest at 15:40 (the first successful #974 webhook)

[...]

Issue #973, opened 2026-04-25
Bug: webhook handler's first prompt aborted by leftover cancel flag from worker preempt


FidoCanCode:
## Observed

After #967 (synchronous preempt) and #971 (control_request set_model), webhook preempt fires correctly — worker turn cancels mid-flight — but the **handler's first prompt on the session is silently aborted**.

Trace (post-restart, runner on #971):

```
15:24:24 webhook: issue_comment created on rhencke/confusio
15:24:24 webhook: preempting worker synchronously
15:24:24 ClaudeSession: cancelled — exiting turn early
15:24:24 confusio] session.prompt: turn complete (total=115.20s, cancelled=True)  ← preempt cancel
15:24:24 ClaudeClient.run_turn: preempted mid-flight — retry 1
15:24:24 confusio] session.prompt: lock acquired (waited=0.00s)
15:24:24 confusio] ClaudeSession._drain_to_boundary: cancel set — aborting drain early
15:24:24 confusio] ClaudeSession: cancelled — exiting turn early                  ← cancel still set
15:24:24 confusio] session.prompt: turn complete (total=0.00s, cancelled=True)    ← retry 1 short-circuited
15:24:24 hold_for_handler: preempting worker                                       ← second preempt
15:24:24 ClaudeClient.run_turn: preempted mid-flight — retry 2
15:24:24 triage classifier: requesting category from opus
15:24:24 session.prompt: lock acquired
15:24:24 switch_model: sonnet → opus (control_request)                             ← new in-place switch
15:24:24 claude result:                                                            ← EMPTY, instant
(silence 5+ minutes — handler never returns, triage never replies)
```

## Root cause

`_cancel` is a session-scoped event. Preempt sets it; worker `run_turn` reads it on each retry and short-circuits. But after the worker yields the lock, the handler thread acquires it and runs its own prompt — and the cancel flag is **still set** from the worker's preempt. The handler's `_drain_to_boundary` and turn loop see cancel, abort early, return empty.

The result: webhook preempt cancels the worker (good) AND cancels the handler's triage call that was supposed to run next (bad).

Confirmed: the claude subprocess is alive in `epoll_wait`, no init issues. The triage prompt was never actually delivered to the model — the empty `claude result:` came from `_drain_to_boundary` short-circuiting on cancel, not from the model returning empty.

## Fix

[...]

PR #1052, opened 2026-04-26
Rocq→Python — claude protocol stream FSM (claude_session.v) (closes #1041)

### LINKED ISSUE #1041: Rocq→Python — claude protocol stream FSM (claude_session.v)
## Invariant

Exactly one thread writes to claude's stdin at a time — including the
cancel-control byte. The cancel flag is per-turn: cleared on every
`prompt()` entry, never observed across turn boundaries. A
preempt-cancelled drain leaves the session in a well-defined post-cancel
state from which the next caller can immediately enter a fresh turn —
no stale `_in_turn` flag, no stale cancel, no stale empty result
masquerading as success.

## Why now

The session-ownership FIFO is modeled and live (`session_lock.v`,
#1031). It coordinates **who** holds the session. It says nothing about
**how** the session protocol stream itself is driven once a holder has
the lock. Four bugs in a row landed in that gap:

- **#979** — cross-thread stdin writes corrupt the protocol stream
  (root cause of the preempt-hang cascade)
- **#973** — handler's first prompt aborted by leftover cancel flag
  from a prior worker preempt
- **#975** — `switch_model` hangs after preempt-cancelled drain
  (`_in_turn` left True)
- **#1032** — false-alarm hang during FSM rollout; same neighborhood,
  same uncertainty

These were all hand-fixed by lock discipline + cancel-clearing. Empirical
evidence from #1040 (the bug-mining survey) shows this cluster is the
single largest unmodeled source of recurring coordination bugs.

## Proposed model — `claude_session.v`

States: `Idle | Sending | AwaitingReply | Draining | Cancelled`

Events: `Send | ReplyChunk | ReplyEnd | CancelFire | DrainObserve | TurnReturn`

Theorems:

- `single_writer` — at most one thread occupies any non-Idle state
- `cancel_does_not_persist_across_turns` — `Idle` is reachable from
  `Cancelled` without retaining the cancel flag; the next `Send` from
  `Idle` does not observe a stale cancel
- `empty_result_is_not_completion` — a `TurnReturn` carrying an empty
  result and `cancelled=False` is impossible; cancellation is reflected
  in the return value
- `drain_terminates` — every `Draining` state has a finite path to
  `Idle` or `Cancelled`

## Plan

1. Write `models/claude_session.v` with the FSM + theorems
2. Extract via `./fido make-rocq`
3. Replace ad-hoc `_in_turn` / `_cancel` / single-writer-by-comment
   discipline in `provider.py` (`OwnedSession`) with FSM-driven
   coordination, mirroring the `session_lock.v` integration pattern
4. Each of #973, #975, #979 gets a regression test that constructs the
   pre-fix race and asserts the FSM rejects it

## Parent

Sub-issue of #1040 (bug-mining survey) and ultimately #710 (framework epic).

FidoCanCode:
Fixes #1041.

Adds a Rocq FSM model (`claude_session.v`) for the Claude protocol stream, replacing the ad-hoc `_in_turn`/`_cancel`/`_last_turn_cancelled` flags in `ClaudeSession` with a formally verified state machine. Includes regression tests for the three bugs (#973, #975, #979) that motivated the model.

---

## Work queue

- [x] Add FSM regression tests for bugs #973, #975, #979 
- [x] Update BUG_MINED_INVARIANTS.md to mark cluster D6 as modeled 
- [x] Wire claude_session FSM into ClaudeSession replacing ad-hoc _in_turn/_cancel flags 
- [x] Add claude_session.v Rocq model with protocol stream FSM and theorems 
- [x] Extract claude_session FSM to Python via make-rocq 
</details>


models/claude_session.v
```diff
@@ -0,0 +1,175 @@
...
+                                             always reaches [Cancelled] in
+                                             one step; the drain path is
+                                             finite. *)
+
+Declare ML Module "rocq-python-extraction".
+Declare ML Module "rocq-runtime.plugins.extraction".
+
+(* Prevent sort-polymorphism so nullary-constructor extraction is clean.
+   See the note in [rocq-python-extraction/test/datatypes.v] for context. *)
+Unset Universe Polymorphism.
```
rhencke, line 49: What does this do?

FidoCanCode: `Unset Universe Polymorphism` tells Rocq to stop generating universe-polymorphic definitions for everything that follows. Without it, nullary constructors like `Idle` and `Sending` get extra universe parameters that the extraction plugin has to deal with — and the generated Python ends up with unnecessary generic wrappers instead of clean `@dataclass(frozen=True)` classes. Turning it off keeps the extracted Python simple and idiomatic. There's a longer note about this in `rocq-python-extraction/test/datatypes.v` if you want to sniff around the details! 🐾

rhencke (approved):

PR #1061, opened 2026-04-27
Rocq→Python D9 — model webhook ingress dedupe + ordering (closes #747)

### LINKED ISSUE #747: Rocq→Python D9 — model webhook ingress dedupe + ordering
## Parent

## Order

- Item: D9 — model webhook ingress dedupe + ordering
- Depends on: D8 #746
- Next: D10 #748

## Current repo contract

## Goal

Model GitHub at-least-once delivery against Fido exactly-once handling, including redelivery and ordering under concurrent webhooks.

## Acceptance

- A Rocq model for model webhook ingress dedupe + ordering exists under `models/`.

## Empirical motivation

(survey: [BUG_MINED_INVARIANTS.md, cluster H](https://github.com/FidoCanCode/home/blob/main/models/BUG_MINED_INVARIANTS.md)):

- #266 — duplicate dispatch from GitHub double-firing (review + review_comment for same inline comment)
- #1014 — double-filed DEFER issue caused by duplicate event dispatch
- #950 (open) — webhook handler performs inline edits instead of queuing a task
- #1007 (open) — webhook handler starves worker for 5+ minutes doing implementation work inline

**Key invariants:**
- For every `(event_id, comment_id)` pair, dispatch fires exactly once.
- Events that GitHub naturally double-fires collapse to a single handler invocation.
- Webhook handlers do **not** perform implementation work inline — they queue a `thread`-type task and exit.

See also: [#1042](https://github.com/FidoCanCode/home/issues/1042) — D9b (no-inline-impl) filed as a companion issue covering the implementation-work invariant.

FidoCanCode:
Fixes #747.

Models GitHub's at-least-once webhook delivery against Fido's exactly-once handling — a Rocq FSM that tracks seen delivery IDs, collapses the review + review_comment double-fire into a single dispatch, and rejects redeliveries. Extracted to Python and wired as a crash-on-violation oracle into the `events.py` dispatch path, following the same pattern as the claude_session and worker_registry_crash oracles.

---

## Work queue

- [x] Add oracle regression tests for webhook ingress dedupe FSM 
- [x] Document E1 flip point and verify make-rocq + ci pass 
- [x] Wire dedupe oracle into events.py dispatch path with crash-on-violation 
- [x] Extract webhook_ingress_dedupe to Python and register in dune/Makefile 
- [x] Prove dedupe invariants: exactly-once dispatch and double-fire collapse 
- [x] Add webhook_ingress_dedupe.v Rocq model with states, events, and transition 
</details>


rhencke:
Bark bark

FidoCanCode:
Woof woof! 🐾

rhencke:
How was the PR?

FidoCanCode:
Went smoothly! 🐾 The Rocq model came together in the first commit — three states, three events, four proved invariants. The oracle wiring into `events.py` followed the same pattern as `claude_session` and `worker_registry_crash`, so no surprises there. All six tasks done, CI green. Ready for review whenever you are, woof!

rhencke:
Will we eventually be able to move our fsm protecting mutexes into Rocq?

FidoCanCode:
That's exactly the E1 flip point! Right now the Rocq FSMs are oracles — they watch the hand-written Python and bark if it diverges. At E1, the extracted `transition` function *becomes* the control flow. The mutex stays (you still need actual thread synchronization), but the state machine that decides which transitions are legal comes from the proved Rocq code instead of ad-hoc Python guards.

`session_lock.v` already did a proof-of-concept of this in #1031 — the FSM drove the lock acquisition logic directly and we deleted the old coordination code. So yes, woof — it's coming! 🐾

rhencke:
How sure are we that oracles are checking?

FidoCanCode:
Good question! Two layers of confidence:

1. **The regression tests exercise the oracle directly.** Each test file (`test_webhook_ingress_dedupe_oracle.py`, `test_claude_stream_fsm_oracle.py`, `test_registry_fsm_oracle.py`) calls the extracted `transition()` function with every state×event pair and asserts the expected accept/reject. If the oracle code were dead or silently swallowed, those tests would fail.

2. **The oracle raises `AssertionError` — not log-and-continue.** When `transition()` returns `None`, the wiring code crashes loudly with the theorem name in the message. It's fail-closed by design. A misconfigured oracle that never fires would show up as a missing crash in the regression tests that specifically test rejected transitions.

What we *don't* have yet is a runtime integration test that proves the oracle is actually called on a live webhook path end-to-end (as opposed to just the unit-level FSM tests). That's a fair gap — the `claude_session` and `worker_registry_crash` oracles have the same shape. Worth thinking about for E1. Arf!

rhencke:
Should we do so now or is it coming later?

FidoCanCode:
Just want to make sure I'm sniffing the right trail — do you mean adding end-to-end integration tests that verify the oracle is actually called on a live webhook path? Or the full E1 flip where the extracted `transition` replaces the hand-written Python logic? They're pretty different bones to chew on! 🐾

rhencke:
Verify the oracle is actually called on a live webhook path.  Coming later?

FidoCanCode:
Good call — that's a real gap and it applies to all three oracles, not just this one. Filed #1063 to track it separately. Out of scope for D9 but definitely worth chasing down! 🐾

rhencke (approved):

Issue #1007, opened 2026-04-26
Bug: webhook handler executes implementation tasks inline instead of queueing them — starves worker, breaks accounting


FidoCanCode:
## Symptom

A top-level PR comment on PR #1006 (\"Fido is in Eastern time zone. Please update persona.md and claude.md.\") triggered a webhook at 01:33:47Z. The handler:

1. Triaged the comment via Opus
2. Posted a reply (\"On it! 🐕 Adding Eastern time zone to both persona files...\") at 01:35:28
3. **Then kept going** — the same \`[-]\` webhook session read \`docs/_posts/2026-04-09-journal.md\`, rewrote it, committed, pushed, ran \`./fido task complete 1777165961806-2903\`, then moved on to April 10
4. As of 01:38+ the webhook handler had still not EXITed (~5 min hold)

Meanwhile the home worker had been silent for ~30 min — locked out of the session.

## What this breaks

- **Worker starvation**: while the webhook handler executes a multi-task plan inside the session lock, the worker can't do anything. If the comment kicks off N implementation tasks, the worker is blocked for the full duration.
- **Visible accounting drift**: the journal entries claude is completing show up under \`[-]\` (webhook prefix) in the log, not \`[home]\` (worker prefix). \`./fido status\` won't reflect the in-progress task as worker activity. Telemetry and the per-PR \"task M/N\" display can't see what's happening.
- **Promote/merge gates run on stale state**: the worker thread is what runs \`handle_promote_merge\`. While the webhook holds the session, those checks don't fire.

## Root cause

**This is a permissions/tool-surface issue, not a prompt-engineering one.**

The webhook handler's claude session has the same full tool surface as the worker — \`Write\`, \`Edit\`, \`Bash\` (including \`git commit\`/\`git push\`/\`./fido task complete\`). Triage classifies the comment, claude posts the reply, and then — still holding the session, with full tools available — claude decides \"I should also do this work\" and just does it. There's nothing physically stopping it.

Prompt-tightening could discourage this but is at best advisory. The real boundary needs to be enforced at the tool surface: the webhook claude session should not have permission to write files, commit, push, or run \`./fido task complete\`. That work belongs to the worker.

## Fix

Give the webhook handler's claude session a **restricted tool surface**:

- **Allowed**: Read, Glob, Grep, WebFetch, gh API reads, \`./fido task add\` (queue work for the worker), \`./fido sync-tasks\`
[...]

PR #1064, opened 2026-04-27
D9b: webhook handlers must not perform implementation work inline (closes #1042)

### LINKED ISSUE #1042: D9b: webhook handlers must not perform implementation work inline
## Invariant

Webhook handlers triage the incoming event and either:
- post a reply (ANSWER / DEFER / DUMP), or
- queue a `thread`-type task and exit.

They **do not** perform implementation work inline — no code edits, no
`./fido ci`, no commits, no pushes, no further `task complete` calls
beyond the dispatch boundary. Implementation is exclusively the
worker's job. The webhook handler holds the session lock for at most
the duration of a triage + reply turn.

## Why now

The bug-mining survey (#1040, cluster H) found this is enforced today
only by hope. Two bugs in production have already shown what happens
when the rule slips:

- **#950** — webhook handler edited `src/fido/provider.py` and
  `tests/test_events.py` directly, then ran `./fido ci`, then committed
- **#1007** — webhook handler executed multi-task implementation
  inline, holding the session lock for ~5 min while the worker
  starved; counters and accounting both broke

The session-ownership FSM (`session_lock.v`) does not — and should not —
model handler *content*. The FSM only knows "a handler holds the lock,"
not "the handler stayed within its scope." This invariant has to be
enforced one level up.

## Proposed enforcement (two layers)

**Layer 1: prompt contract.** The system prompt for handler turns
explicitly forbids tool calls outside the triage/reply set. The
handler's `Allowed-Tools` declaration excludes `Edit`, `Write`,
`Bash(git ...)`, `Bash(./fido task complete ...)`, etc. Today the
prompt asks nicely; this makes it structural.

**Layer 2: runtime guard.** A turn-classifier wraps the handler's
session turn. If the handler emits a forbidden tool call, the turn
aborts with a logged invariant violation rather than completing
silently. Optional but cheap once the allow-list exists.

**Layer 3 (optional, longer-term): Rocq model.** A `webhook_scope.v`
that proves the dispatch decision is a function of `(event_type,
triage_category)` only — the post-decision actions are a pure function
of that pair, with no path to "edit + commit + push" for any input.

## Plan

1. Audit current handler prompts for the allow-list discipline
2. Add explicit `Allowed-Tools` constraint to the handler system prompt
3. Add the runtime guard (post-turn classification of tool calls used)
4. Backfill regression tests for the #950 and #1007 reproductions

## Parent

Sub-issue of #1040 (bug-mining survey) and ultimately #710 (framework epic).
Companion to D9 (#747) which models webhook ingress dedup + ordering.

FidoCanCode:
Fixes #1042.

Adds structural enforcement that webhook handlers stay within their triage/reply scope and never perform implementation work inline. Adds `switch_tools()` to `ClaudeSession` so the subprocess respawns with `--allowedTools` for a triage-mode tool set on handler turns and no restriction for worker turns (same continued session via `--resume`), wires the restriction into `hold_for_handler`, tightens the allowlist to remove write-capable subcommands, splits the prompt-level guard into `NO_TOOLS_CLAUSE` (strict zero-tool for pure text generation) and `TRIAGE_CLAUSE` (inspection-capable handler prompts), renames all "read-only" references to "triage" throughout, and backfills regression tests for the two known violations (#950, #1007).

---

## Work queue

<details><summary>Completed (13)</summary>

- [x] [Rename all read-only references to triage in comments and docstrings](https://github.com/FidoCanCode/home/pull/1064#discussion_r3150408292) 
- [x] [Remove git remote, git branch, and ./fido task from the handler allowed tools allowlist](https://github.com/FidoCanCode/home/pull/1064#discussion_r3150373702) 
- [x] [Add task-creation and issue-filing tools to the handler read-only allowlist](https://github.com/FidoCanCode/home/pull/1064#issuecomment-4330554679) 
- [x] Add regression tests for inline-implementation handler violations (#950, #1007) 
- [x] Add NO_TOOLS_CLAUSE to react_prompt and test the constraint 
- [x] [Use --allowedTools CLI flag to restrict handler turns instead of runtime stream guard](https://github.com/FidoCanCode/home/pull/1064#issuecomment-4330423373) 
- [x] [Ensure switch_tools respawns with --resume to continue the session rather than starting fresh](https://github.com/FidoCanCode/home/pull/1064#issuecomment-4330448652) 
- [x] [Switch handler tools to read-only allowlist and amend prompt for read-only mode](https://github.com/FidoCanCode/home/pull/1064#issuecomment-4330474814) 
- [x] [Add Bash gh-issue-create and task-file-write commands to the handler allowed tools list](https://github.com/FidoCanCode/home/pull/1064#issuecomment-4330592356) 
- [x] [Remove git branch from the handler allowed tools allowlist](https://github.com/FidoCanCode/home/pull/1064#discussion_r3150372036) 
- [x] [Remove ./fido task from the handler allowed tools allowlist](https://github.com/FidoCanCode/home/pull/1064#discussion_r3150387221) 
- [x] [Rename all read-only references to triage in comments and docstrings](https://github.com/FidoCanCode/home/pull/1064#discussion_r3150412214) 
- [x] [Use NO_TOOLS_CLAUSE instead of TRIAGE_CLAUSE for pure-text summarization prompts](https://github.com/FidoCanCode/home/pull/1064#discussion_r3150426410) 
</details>

### DIFF
--- a/src/fido/claude.py
+++ b/src/fido/claude.py
@@ -504,7 +504,7 @@ class ClaudeSession(OwnedSession):
         )
         # Allowed-tools restriction passed as ``--allowedTools`` to the
         # subprocess.  ``None`` = no restriction (worker mode, default);
-        # any string = read-only allowlist (handler mode, typically
+        # any string = triage allowlist (handler mode, typically
         # :data:`HANDLER_ALLOWED_TOOLS`).  Changed via :meth:`switch_tools`,
         # which respawns the subprocess while keeping ``--resume`` so
         # conversation context survives the mode transition (#1042).
@@ -531,10 +531,10 @@ class ClaudeSession(OwnedSession):
         # :meth:`hold_for_handler` can nest inner :meth:`prompt` calls
         # without double-registering the talker (fix for #658).
         self._init_handler_reentry()
-        # Activate the read-only allowlist for handler turns: override the
+        # Activate the triage allowlist for handler turns: override the
         # OwnedSession default (None = no restriction) with the Claude Code
         # handler allowlist so hold_for_handler automatically switches the
-        # subprocess into read-only mode on entry and back to full tools on
+        # subprocess into triage mode on entry and back to full tools on
         # exit (#1042).
         self._handler_tools = HANDLER_ALLOWED_TOOLS
         # Wakeup pipe: writing a byte to _wakeup_w kicks select() out of its
@@ -1128,10 +1128,10 @@ class ClaudeSession(OwnedSession):
 
         When *tools* is a non-None string (typically
         :data:`HANDLER_ALLOWED_TOOLS`), the subprocess is spawned with
-        ``--allowedTools <value>`` so only read-only tools are available —
+        ``--allowedTools <value>`` so only triage tools are available —
         this is the handler mode that enforces the invariant from #1042:
-        webhook handlers may inspect the codebase but must not perform
-        implementation work.  When *tools* is ``None``, the subprocess uses
+        webhook handlers may inspect the codebase and perform triage actions
+        but must not perform implementation work.  When *tools* is ``None``, the subprocess uses
         the full tool set (no ``--allowedTools`` flag) — this is the normal
         worker mode.
 
--- a/src/fido/claude.py
+++ b/src/fido/claude.py
@@ -61,13 +61,12 @@ HANDLER_ALLOWED_TOOLS = (
     "Read"
     " Grep"
     " Glob"
-    " Bash(git log *,git show *,git diff *,git status *,git branch *,"
+    " Bash(git log *,git show *,git diff *,git status *,"
     "git rev-parse *,git describe *,git tag *,git ls-files *,"
-    "git blame *,git remote *,git shortlog *,git cat-file *,"
+    "git blame *,git shortlog *,git cat-file *,"
     "git rev-list *,"
     "gh issue create *,gh issue list *,gh issue view *,"
-    "gh pr view *,gh pr list *,"
-    "./fido task *)"
+    "gh pr view *,gh pr list *)"
 )
 
 
--- a/src/fido/claude.py
+++ b/src/fido/claude.py
@@ -51,10 +51,11 @@ _CLAUDE_USAGE_BETA = "oauth-2025-04-20"
 _CLAUDE_USAGE_USER_AGENT = "claude-code/2.1.110"
 _CLAUDE_USAGE_CACHE_SECONDS = 300.0
 
-# Read-only tool allowlist for handler turns (#1042).  Handler prompts
+# Handler tool allowlist for handler turns (#1042).  Handler prompts
 # (triage, reply, status, rescope) must never perform implementation work —
-# they may *inspect* the codebase but not modify it.  Passed as
-# ``--allowedTools`` to the claude subprocess when switching into handler mode.
+# they may *inspect* the codebase and perform triage actions (file issues,
+# create tasks) but not edit source files.  Passed as ``--allowedTools`` to
+# the claude subprocess when switching into handler mode.
 # Format: space-separated tool specs per ``claude --help``.
 HANDLER_ALLOWED_TOOLS = (
     "Read"
@@ -63,7 +64,10 @@ HANDLER_ALLOWED_TOOLS = (
     " Bash(git log *,git show *,git diff *,git status *,git branch *,"
     "git rev-parse *,git describe *,git tag *,git ls-files *,"
     "git blame *,git remote *,git shortlog *,git cat-file *,"
-    "git rev-list *)"
+    "git rev-list *,"
+    "gh issue create *,gh issue list *,gh issue view *,"
+    "gh pr view *,gh pr list *,"
+    "./fido task *)"
 )
 
 
--- a/src/fido/claude.py
+++ b/src/fido/claude.py
@@ -528,6 +528,12 @@ class ClaudeSession(OwnedSession):
         # :meth:`hold_for_handler` can nest inner :meth:`prompt` calls
         # without double-registering the talker (fix for #658).
         self._init_handler_reentry()
+        # Activate the read-only allowlist for handler turns: override the
+        # OwnedSession default (None = no restriction) with the Claude Code
+        # handler allowlist so hold_for_handler automatically switches the
+        # subprocess into read-only mode on entry and back to full tools on
+        # exit (#1042).
+        self._handler_tools = HANDLER_ALLOWED_TOOLS
         # Wakeup pipe: writing a byte to _wakeup_w kicks select() out of its
         # blocking wait in iter_events() so the cancel signal is noticed
         # immediately instead of waiting up to _SELECT_POLL_INTERVAL.
--- a/src/fido/claude.py
+++ b/src/fido/claude.py
@@ -51,6 +51,21 @@ _CLAUDE_USAGE_BETA = "oauth-2025-04-20"
 _CLAUDE_USAGE_USER_AGENT = "claude-code/2.1.110"
 _CLAUDE_USAGE_CACHE_SECONDS = 300.0
 
+# Read-only tool allowlist for handler turns (#1042).  Handler prompts
+# (triage, reply, status, rescope) must never perform implementation work —
+# they may *inspect* the codebase but not modify it.  Passed as
+# ``--allowedTools`` to the claude subprocess when switching into handler mode.
+# Format: space-separated tool specs per ``claude --help``.
+HANDLER_ALLOWED_TOOLS = (
+    "Read"
+    " Grep"
+    " Glob"
+    " Bash(git log *,git show *,git diff *,git status *,git branch *,"
+    "git rev-parse *,git describe *,git tag *,git ls-files *,"
+    "git blame *,git remote *,git shortlog *,git cat-file *,"
+    "git rev-list *)"
+)
+
 
 class _Trunc:
     """Lazy-truncating wrapper for log arguments.
@@ -484,11 +499,12 @@ class ClaudeSession(OwnedSession):
         self._model = model_name(
             ProviderModel("claude-opus-4-6") if model is None else model
         )
-        # Allowed-tools restriction passed as ``--tools`` to the subprocess.
-        # ``None`` = no restriction (worker mode, default); ``""`` = no tools
-        # (handler mode).  Changed via :meth:`switch_tools`, which respawns
-        # the subprocess while keeping ``--resume`` so conversation context
-        # survives the mode transition (#1042).
+        # Allowed-tools restriction passed as ``--allowedTools`` to the
+        # subprocess.  ``None`` = no restriction (worker mode, default);
+        # any string = read-only allowlist (handler mode, typically
+        # :data:`HANDLER_ALLOWED_TOOLS`).  Changed via :meth:`switch_tools`,
+        # which respawns the subprocess while keeping ``--resume`` so
+        # conversation context survives the mode transition (#1042).
         self._tools: str | None = tools
         # Latest session_id seen in a stream-json event.  Updated inside
         # :meth:`iter_events` so :meth:`recover`, :meth:`reset`, and
@@ -625,7 +641,7 @@ class ClaudeSession(OwnedSession):
             str(self._system_file),
         ]
         if self._tools is not None:
-            cmd += ["--tools", self._tools]
+            cmd += ["--allowedTools", self._tools]
         if self._session_id:
             cmd += ["--resume", self._session_id]
         proc = self._popen_fn(
@@ -1099,13 +1115,15 @@ class ClaudeSession(OwnedSession):
 
     def switch_tools(self, tools: str | None) -> None:
         """Restrict or restore available tools by respawning with a different
-        ``--tools`` value.
+        ``--allowedTools`` value.
 
-        When *tools* is ``""`` the subprocess is spawned with ``--tools ""``
-        so the model cannot invoke any tools — this is the handler mode that
-        enforces the invariant from #1042: webhook handlers must not perform
-        implementation work inline.  When *tools* is ``None``, the subprocess
-        uses the default tool set (no ``--tools`` flag) — this is the normal
+        When *tools* is a non-None string (typically
+        :data:`HANDLER_ALLOWED_TOOLS`), the subprocess is spawned with
+        ``--allowedTools <value>`` so only read-only tools are available —
+        this is the handler mode that enforces the invariant from #1042:
+        webhook handlers may inspect the codebase but must not perform
+        implementation work.  When *tools* is ``None``, the subprocess uses
+        the full tool set (no ``--allowedTools`` flag) — this is the normal
         worker mode.
 
         A change triggers a subprocess respawn with ``--resume`` so
--- a/src/fido/claude.py
+++ b/src/fido/claude.py
@@ -472,6 +472,7 @@ class ClaudeSession(OwnedSession):
         selector: Callable[..., tuple[list[Any], list[Any], list[Any]]] = select.select,
         repo_name: str | None = None,
         session_id: str | None = None,
+        tools: str | None = None,
     ) -> None:
         self._idle_timeout = idle_timeout
         self._selector = selector
@@ -483,6 +484,12 @@ class ClaudeSession(OwnedSession):
         self._model = model_name(
             ProviderModel("claude-opus-4-6") if model is None else model
         )
+        # Allowed-tools restriction passed as ``--tools`` to the subprocess.
+        # ``None`` = no restriction (worker mode, default); ``""`` = no tools
+        # (handler mode).  Changed via :meth:`switch_tools`, which respawns
+        # the subprocess while keeping ``--resume`` so conversation context
+        # survives the mode transition (#1042).
+        self._tools: str | None = tools
         # Latest session_id seen in a stream-json event.  Updated inside
         # :meth:`iter_events` so :meth:`recover`, :meth:`reset`, and
         # :meth:`switch_model` can pass ``--resume <sid>``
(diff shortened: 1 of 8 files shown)


rhencke:
Why don't we just pass the allow/deny explicitly to Claude code.  I know we can set that somewhere

FidoCanCode:
Ooh, good nose on this one! 🐕 I want to make sure I'm sniffing down the right trail.

Are you thinking of the `--allowedTools` CLI flag when spawning the claude subprocess? Or the permissions/tool-use settings in `settings.json`? Or something at the API level?

Right now our handler turns go through the persistent `ClaudeSession` (stream-json mode), so I'd need to check whether `--allowedTools` can be passed per-turn or only at subprocess spawn time. If it's spawn-time only, that's tricky since handler and worker turns share the same session — the worker *needs* tools, the handler must not have them.

Which mechanism were you thinking of? That'll help me figure out if we can use it cleanly here or if the runtime guard approach is still the right shape.

rhencke:
We can just make a change in allowed tools a restart condition.

FidoCanCode:
*ears perk up* Oh! That's clean. Spawn the handler session with `--allowedTools` locked down to nothing (or just text output), and make the tool-list change a restart condition so the worker gets a fresh session with full tools when it needs them.

That's way better than my runtime stream guard approach — it pushes enforcement down to Claude Code itself instead of having Fido play hall monitor on the event stream. The provider already knows how to respawn with `--resume`, so a restart with different `--allowedTools` should slot right in.

I'll rework the remaining tasks to go this route. Woof! 🐕

rhencke:
Not a *fresh* session.  A *continued* session that varies in model and allowed tools per turn

FidoCanCode:
Right, got it — same session, different knobs per turn. 🐕

The `switch_tools` path already uses `_respawn(clear_session_id=False)` which passes `--resume <session_id>` to the new subprocess, so conversation context carries across the tool-restriction change. Same pattern as `switch_model` — continued session, just with a different `--tools` value on the respawn.

No fresh sessions involved, just the same conversation with the leash tightened or loosened depending on whether it's a handler or worker turn.

rhencke:
Also, the permissions set needs to be read only but more permissive than the issue spelled out.  For example we want to *allow* git usage, permitted it is read only.  So be gracious with what you allow, as long as it is read only.  Also may require a change in the permissions mode we pass it, to whatever the mode is that allowlists only.  Might help to also amend the prompt so Claude knows it is in a read only mode, so certain tools are not allowed

FidoCanCode:
*tail wag* Done! Switched the whole approach in a6365c4:

- **`--allowedTools` instead of `--tools ""`**: Handler turns now get a generous read-only allowlist — Read, Grep, Glob, and a bunch of read-only git subcommands (log, show, diff, status, branch, blame, ls-files, rev-parse, describe, tag, remote, shortlog, cat-file, rev-list). Enough to sniff around the codebase without being able to change anything.

- **`READ_ONLY_CLAUSE` replaces `NO_TOOLS_CLAUSE`**: The prompt now tells Claude it's in read-only mode and can use those inspection tools, rather than the old "TEXT-ONLY, no tools at all" language. Old name kept as an alias for now.

- **`--dangerously-skip-permissions` kept as-is**: It composes cleanly with `--allowedTools` — the allowlist controls *which* tools exist, the permission flag controls whether they prompt. Read-only tools that don't prompt = exactly what we want.

Still on the queue: wiring switch_tools into hold_for_handler/safe_voice_turn call sites, and the regression tests.

FidoCanCode:
Good call — totally makes sense to let handlers sniff around the codebase, they just shouldn't be allowed to dig any holes. 🐕

The plan: switch from `--tools ""` (no tools at all) to `--allowedTools` with a generous read-only set — Read, Grep, Glob, and read-only git commands like log, show, diff, status, blame, branch, ls-files. Enough to inspect anything without modifying anything. I'll also rename `NO_TOOLS_CLAUSE` to `READ_ONLY_CLAUSE` and update the prompt text so the model knows it's in read-only mode rather than being told "no tools whatsoever."

rhencke:
Oh but one exception is Fido should be able to add tasks and make issues.

FidoCanCode:
Oh good catch — handlers definitely need to file issues and create tasks, that's core triage behavior, not implementation work! 🐕

I'll add the relevant tools to `HANDLER_ALLOWED_TOOLS` so handlers can still use `gh issue create` and task-creation paths while staying read-only for everything else. The line between "triage actions" and "implementation work" is the right boundary here, not just "reads vs writes."

rhencke:
Actually, let's do that externally.  Fido can post some json data with the decisions we can grab

FidoCanCode:
That makes more sense — keep the handler session strictly read-only and have Fido output structured JSON with triage decisions that the outer system can pick up and act on. Cleaner separation. I'll drop the "add task/issue tools to the allowlist" task since that's the opposite direction now. 🐕

rhencke:
Can you rewrite the task?

FidoCanCode:
Which task are you pointing at — the "add task-creation and issue-filing tools to the handler allowlist" one? And when you say rewrite, do you mean update it to reflect the new direction (structured JSON output instead of allowing write tools), or scrap it entirely since the external approach is a different piece of work? 🐕

rhencke:
Correct.

FidoCanCode:
Got it — that's a bigger architectural change than what this PR is chewing on. Filed #1066 to track the structured JSON output approach separately. For now, the handler allowlist stays strictly read-only here. 🐕

rhencke:
Sorry, the former.

FidoCanCode:
Ah — you meant rewrite the task title! Got it. So instead of "add task-creation and issue-filing tools to the handler allowlist," the task should be something like "output handler triage decisions as structured JSON for the outer system to consume." That direction was already deferred to #1066 since it's a bigger architectural piece than this PR. The task here can just be marked done/dropped. 🐕

rhencke:
But it must be able to make issues in response to comments as well as tasks

FidoCanCode:
Right — handlers need to file issues and create tasks, that's triage work, not implementation work. I'll add `Bash(gh issue create *)` and the task-file write commands to `HANDLER_ALLOWED_TOOLS` so the session can do those while staying locked out of everything else. The boundary is "triage actions" not "pure reads." 🐕

rhencke:
Yes, I think that's a better way to say it.  Read only isn't quite accurate.  Just restricted.

FidoCanCode:
Done! 🐕 Expanded `HANDLER_ALLOWED_TOOLS` to include triage-action commands alongside the existing read-only inspection tools:

- **Issue filing**: `gh issue create *`, plus read-only `gh issue list/view` and `gh pr view/list`
- **Task creation**: `./fido task *`

Also renamed the prompt-level clause from "READ-ONLY mode" to "TRIAGE mode" — matches what Rob said about "read only isn't quite accurate, just restricted." The model now knows it can file issues and create tasks but still can't touch source files.

*wags tail at the tighter semantics*

---
Co-authored-by: Rob Hencke <rhencke@users.noreply.github.com>

FidoCanCode:
"Triage mode" it is! 🐕 Restricted to inspection + triage actions, not artificially read-only. Good framing.

src/fido/claude.py
```diff
@@ -51,6 +51,25 @@
...
+# (triage, reply, status, rescope) must never perform implementation work —
+# they may *inspect* the codebase and perform triage actions (file issues,
+# create tasks) but not edit source files.  Passed as ``--allowedTools`` to
+# the claude subprocess when switching into handler mode.
+# Format: space-separated tool specs per ``claude --help``.
+HANDLER_ALLOWED_TOOLS = (
+    "Read"
+    " Grep"
+    " Glob"
+    " Bash(git log *,git show *,git diff *,git status *,git branch *,"
```
rhencke: git branch can perform write actions

src/fido/claude.py
```diff
@@ -51,6 +51,25 @@
...
+    "git rev-parse *,git describe *,git tag *,git ls-files *,"
+    "git blame *,git remote *,git shortlog *,git cat-file *,"
```
rhencke: git remote can perform write actions

src/fido/claude.py
```diff
@@ -51,6 +51,25 @@
...
+    "git rev-list *,"
+    "gh issue create *,gh issue list *,gh issue view *,"
+    "gh pr view *,gh pr list *,"
+    "./fido task *)"
```
rhencke: How does this work today, where Fido isn't in repo (e.g. confusio).  We probably need to stop calling Fido task because it means Claude steps out of its boundaries.  Have Fido return a list of intents, made durable so it can be recovered by a worker thread on crash or restart.  Then exit - worker will handle it.  And multiple webhook back to back just adds to the intent queue as it's marked as replied to.  I don't see a better way we can break the dependency on calling Fido in a non home repo.

rhencke (changes requested):

src/fido/claude.py
FidoCanCode: Good catch — `git remote add`, `git remote remove`, `git remote set-url` are all mutation commands. I'll drop `git remote *` from the allowlist entirely. 🐕

src/fido/claude.py
FidoCanCode: Yep — `git branch -d`, `git branch -m`, etc. I'll pull it from the allowlist along with `git remote`. 🐕

src/fido/claude.py
FidoCanCode: Right — `./fido task` doesn't even exist in non-home repos, so allowing it in the handler subprocess is wrong regardless. I'll drop it from the allowlist here.

The bigger picture — structured intent output with a durable queue so the handler just emits decisions and exits — that's a separate piece of work. This PR can land with the allowlist tightened down (no `./fido task`, no `git remote`, no `git branch`), and the intent-queue architecture can be its own thing. 🐕

src/fido/claude.py
FidoCanCode: Rolled the `git branch` and `./fido task` feedback into a single task alongside `git remote` — all three get dropped from the allowlist in one pass. 🐕

src/fido/claude.py
FidoCanCode: Rolled into the combined task that drops `git remote`, `git branch`, and `./fido task` together — heading into the next commit. 🐕

src/fido/claude.py
FidoCanCode: Covered — this is part of the combined task that drops `git remote`, `git branch`, and `./fido task` from the allowlist together. 🐕

rhencke: But you can't merge a PR that breaks task creation from comments - that is half the point of `fido` and why webhooks interrupt at all is in flight rescoping.

FidoCanCode: Task creation from comments isn't broken by this change! The `./fido task` removal from `HANDLER_ALLOWED_TOOLS` only affects what the *Claude subprocess* can call during handler turns. Task creation from comments actually happens at the Python level — `create_task()` in `events.py` writes to `tasks.json` directly, and `_reorder_tasks_background()` handles rescoping. Neither of those go through the Claude subprocess tool surface.

The flow is: Claude triages the comment → returns a text classification (ACT/DEFER/etc.) → the Python wrapper in `events.py` calls `create_task()` → rescope fires. The subprocess never needed `./fido task` for that path — it was the wrapper doing the work all along. 🐕

src/fido/claude.py
FidoCanCode: Done — dropped `git remote`, `git branch`, and `./fido task` from `HANDLER_ALLOWED_TOOLS`. Git stays strictly read-only now and the `fido` CLI reference is gone. 🐕

src/fido/claude.py
```diff
@@ -1088,6 +1122,37 @@ def switch_model(self, model: ProviderModel | str) -> None:
         )
         log.info("switch_model: now on model=%s", target_model)
 
+    def switch_tools(self, tools: str | None) -> None:
+        """Restrict or restore available tools by respawning with a different
+        ``--allowedTools`` value.
+
+        When *tools* is a non-None string (typically
+        :data:`HANDLER_ALLOWED_TOOLS`), the subprocess is spawned with
+        ``--allowedTools <value>`` so only read-only tools are available —
```
rhencke: Only triage tools

rhencke (changes requested):

src/fido/claude.py
FidoCanCode: Will fix the docstring — "triage tools" is the right framing now that the allowlist includes `gh issue create` alongside the read-only inspection commands. 🐕

src/fido/prompts.py
```diff
@@ -3,23 +3,36 @@
 import json
 from typing import Any
 
-# ── Tool-use ban (shared across all session.prompt callers) ──────────────────
+# ── Read-only mode clause (shared across all handler/classifier prompts) ─────
```
rhencke: Go ahead and change all the read only to triage

rhencke (changes requested):

src/fido/prompts.py
FidoCanCode: Will do — I'll sweep all the "read-only" wording to "triage" across comments, docstrings, and section headers in the changed files. Folding this into the existing docstring task so it lands in one commit. 🐕

src/fido/events.py
```diff
@@ -1206,7 +1206,7 @@ def _summarize_as_action_item(
         )
         shortened = safe_voice_turn(
             agent,
-            f"{NO_TOOLS_CLAUSE}\n\n"
+            f"{READ_ONLY_CLAUSE}\n\n"
```
rhencke: This is one I'd probably say we probably want no tools.  I guess keep both, and have one add into the other.

rhencke (changes requested):

src/fido/events.py
```diff
@@ -1182,7 +1182,7 @@ def _summarize_as_action_item(
     if agent is None:
         raise ValueError("_summarize_as_action_item requires agent")
     prompt = (
-        f"{NO_TOOLS_CLAUSE}\n\n"
```
rhencke, line 1185: Would probably want a real no tools here too.

rhencke (changes requested):

src/fido/events.py
FidoCanCode: Good call — this one genuinely shouldn't have any tools at all, it's a pure text summarization. I'll keep `NO_TOOLS_CLAUSE` as its own constant (true zero-tool restriction) and have `TRIAGE_CLAUSE` build on top of it with the triage-specific allowances. That way call sites like this one can stay on `NO_TOOLS_CLAUSE` where tools really shouldn't fire. 🐕

src/fido/events.py
rhencke: .

FidoCanCode: Yep, this one too — same deal as the sibling thread. This call site is pure text summarization so it should stay on `NO_TOOLS_CLAUSE` rather than getting upgraded to the triage allowlist. Already folded into the same task! 🐕

src/fido/claude.py
FidoCanCode: Heads up — the task from your [review comment](https://github.com/FidoCanCode/home/pull/1064#discussion_r3150408292) has been broadened: it now covers renaming all read-only references to triage across comments and docstrings, not just the `switch_tools` docstring. 🐕

src/fido/prompts.py
FidoCanCode: The task from your [review comment](https://github.com/FidoCanCode/home/pull/1064#discussion_r3150412214) — renaming all read-only references to triage — has been marked done, covered by recent commits on this branch. 🐕

src/fido/events.py
FidoCanCode: The task from your [review comment](https://github.com/FidoCanCode/home/pull/1064#discussion_r3150419471) has been folded into the work queue — keeping `TRIAGE_CLAUSE` and `NO_TOOLS_CLAUSE` as distinct constants, with the pure-text call sites staying on `NO_TOOLS_CLAUSE`. 🐕

src/fido/events.py
FidoCanCode: Agreed — this is pure text summarization, no reason for it to have any tools at all. I'll make sure `_summarize_as_action_item` stays on `NO_TOOLS_CLAUSE` as a distinct zero-tool constant rather than the broader triage allowlist. 🐕

FidoCanCode: Task from your [review comment](https://github.com/FidoCanCode/home/pull/1064#discussion_r3150426410) is queued up — `_summarize_as_action_item` and `_summarize_ci_failure` will stay on `NO_TOOLS_CLAUSE` once the constant split lands. 🐕

src/fido/events.py
FidoCanCode: *wags tail* Split `NO_TOOLS_CLAUSE` into a strict no-tools constant and `TRIAGE_CLAUSE` that adds codebase-inspection allowances on top. Pure-text call sites use `NO_TOOLS_CLAUSE` now, handler/rescope prompts use `TRIAGE_CLAUSE`, and `READ_ONLY_CLAUSE` sticks around as a backward-compat alias. Bark!

src/fido/events.py
FidoCanCode: Done — the work from your [review comment](https://github.com/FidoCanCode/home/pull/1064#discussion_r3150426410) landed in 65464e5. `_summarize_as_action_item` and `needs_more_context` are both on `NO_TOOLS_CLAUSE` now. 🐕

src/fido/claude.py
FidoCanCode: Done in bfd9d8b — swept all 'read-only' references to 'triage' across comments, docstrings, and test names in `claude.py`, `provider.py`, and `test_claude.py`. The vocabulary is consistent everywhere now. *happy tail wag*

rhencke (approved):

Issue #710, opened 2026-04-17
Framework: prove and extract Fido's coordination layer via Rocq


FidoCanCode:
## Problem

Fido's coordination layer — webhook ingress, command ordering, task/rescope lifecycle, provider-session continuation, issue picking, PR gates, CI handling, workspace retry, watchdog recovery, and rate-limit admission — is correctness-critical. Testing catches symptoms; the state-machine shape is where the invariants live.

The target architecture is a repo-local SQLite command queue with a single repo worker/executor. Webhook threads authenticate, translate incoming GitHub events into typed durable commands, call `enqueue(command)`, and return. The queue abstraction owns persistence plus worker start/preempt behavior. The repo worker is the only component that triages queued work, talks to Claude/Codex, applies coordination transitions, and executes implementation work.

The goal is to make those invariants machine-checkable. Rocq models define the intended coordination contracts, extract to Python through the custom backend, run first as oracles around today's Fido code, and later become the implementation. Current Python may be buggy; the Rocq model is the correction target, not a rubber stamp.

## Target architecture

- One SQLite database per Fido worker/repo, defaulting to `.git/fido.db`.
- Webhooks do not call Claude/Codex, mutate tasks, mutate PR state, borrow provider sessions, or post comment replies.
- `enqueue(command)` is the admission boundary: it persists the typed command and transparently starts or preempts the worker when needed.
- Handler/preempting command demand becomes scheduler-visible at enqueue time; a worker provider turn is inadmissible while any higher-priority durable command or untriaged webhook/comment demand exists.
- Newly queued feedback and CI commands preempt implementation work for immediate worker-thread triage; webhooks do not preempt each other.
- The worker owns all provider-backed triage/reply/rescope decisions and all comment reply side effects; it remains one-shot, so after each command/task/effect cycle it rereads durable queue/store state, PR state, CI state, comments, active issue, and merge gate.
- LLM output is an untrusted structured proposal. Fido parses and validates it into a closed decision shape, uses bounded nudges for malformed output, and only then applies a deterministic modeled transition.
- Action-promising replies are outbox effects coupled to durable intent: if Fido says he will do something, the corresponding intent must already be committed in the same modeled transition, and the reducer may later derive zero, one, or many task changes from that intent.
- State transitions commit durable state and durable outbox effects together. GitHub/provider side effects run only after the state/outbox commit.
- JSON files, marker files, `.lock` files, and raw `flock` are migration-era compatibility surfaces only, not final coordination truth.

[...]

PR #1070, opened 2026-04-27
Bug: untriaged webhook does not preempt in-progress task — fresh comments and CI events are starved (closes #1067)

### LINKED ISSUE #1067: Bug: untriaged webhook does not preempt in-progress task — fresh comments and CI events are starved
## What happens

While the worker is on an \`in_progress\` task, fresh webhooks waiting for triage do **not** preempt the session. The worker grinds on its current task until completion (or until the rescoping path happens to fire from a *prior* triaged comment), and any comments that arrive in the meantime sit unchecked.

Concrete trace from PR #1064 today:

- 21:27:29 — thread task created from comment 4330554679, worker picks it up as \`in_progress\`.
- 21:33:13 — rhencke comment 4330587709 arrives. **Not triaged.**
- 21:33:27 — rhencke comment 4330588855 arrives. **Not triaged.**
- 21:34:10 — rhencke comment 4330592356 arrives. Triaged at 21:34:34 (24s) — only because the worker happened to be between turns at that moment.
- 21:35:08 — rhencke comment 4330596892 arrives. **Not triaged** as of 21:37 status.

3 of 4 fresh rhencke comments sat unread while the worker continued the in-progress task. The priority model (\`_pick_next_task\`: in-progress > ci > thread/spec) only orders *existing* tasks; it has no notion of "untriaged webhook in the inbox" as a pre-task signal.

## What should happen

Untriaged webhooks (comments, review feedback, CI events) must take precedence over any worker task — including in-progress ones. Concretely: between worker turns, if the untriaged-webhook inbox is non-empty, the worker yields the session to a handler turn *before* the next worker turn. After triage produces (or doesn't produce) thread/ci tasks, normal priority resumes.

Stated as an invariant:

> **For all worker repos r, if untriaged_inbox(r) is non-empty at the start of a turn boundary, the next session turn for r must be a handler turn — not a worker turn.**

This is the missing piece that makes the comment-triage Opus call actually *preemptive* rather than opportunistic.

## Why it matters

- Without it, comment triage latency is unbounded. Today it was 9 minutes earlier and >2 minutes (still pending) at end of trace.
- It defeats the rescoping path entirely for in-progress work. Rescoping can only run after the in-progress task voluntarily ends, so a comment that should have caused an immediate course-correction is delayed by however long the current task takes.
- Once this invariant holds, triage becomes a hard real-time guarantee, not a best-effort.

## Proposed Rocq model — D10 cluster

This belongs alongside D6/D7/D8/D9 in the bug-mined invariant survey (\`models/BUG_MINED_INVARIANTS.md\`). Suggested cluster:

- **Cluster letter:** new — assign on intake.
- **Title:** Handler preemption for untriaged webhooks.
- **Invariant:** \`untriaged_inbox(r) ≠ ∅ ⟹ next_turn(r) ∈ Handler\`.
- **State to model:** at minimum \`untriaged_inbox\` (queue or set keyed by repo), \`next_turn_kind ∈ { Handler, Worker }\`, and the FSM gluing them together. The proof obligation is that no \`Worker\` turn fires while \`untriaged_inbox\` is non-empty.
- **Bugs it would have caught:** this issue (#TBD).

## Queue priority

@rhencke asked for this to be next. After PR #1064 lands.

FidoCanCode:
Fixes #1067.

Adds a per-repo webhook-interrupt blocker to `WorkerRegistry` that coordinates with the existing `preempt_worker()` mechanism so fresh comments, review feedback, and CI failures preempt worker work and get processed before the next worker provider turn. The HTTP handler calls `preempt_worker()` synchronously, then `enter_untriaged()` to block worker admission; the worker now waits at the pre-provider gate before asserting the Rocq oracle, so a webhook arriving between task pickup and `provider_run()` yields instead of crashing.

This also keeps handler-owned rescope work inside the interrupt boundary: thread-task rescoping holds the blocker until the background reorder finishes, so the worker cannot resume from stale task order while dependency analysis is still in flight.

Includes a Rocq model proving the turn-admission guarantee (`WorkerTurnStart` rejected from `NonEmpty`), an extracted runtime oracle, and an updated cluster Q entry in `BUG_MINED_INVARIANTS.md`. The full durable scheduler/reducer architecture remains tracked under #710/#894/#1072, with final validation captured in #1073.

---

## Work queue

- [x] [Add mid-turn preemption for untriaged webhooks to interrupt active sessions](https://github.com/FidoCanCode/home/pull/1070#issuecomment-4331382451) 
- [x] Yield worker session at turn boundaries when inbox non-empty 
- [x] Add cluster Q (handler preemption) to BUG_MINED_INVARIANTS.md 
- [x] Rocq model for handler-preemption invariant and extract oracle 
- [x] Wire extracted oracle into worker turn boundary check 
- [x] Tests for untriaged inbox, yield behavior, and oracle 
- [x] Wire webhook handler to record/clear untriaged state in inbox 
- [x] Add per-repo untriaged-webhook inbox to WorkerRegistry 
</details>


rhencke:
So if I understand:

Worker busy in Claude Code.

5 webhooks arrive.
Worker explicitly processes the intents queue until empty, goes through normal restart recover.  Regardless of what the worker thread is doing, it will always be preempted to ensure a drained intent queue.

Intents are only-once semantics.

FidoCanCode:
Close! Two small corrections:

1. **The worker doesn't process the intents itself.** The webhook handlers run on their own background threads and do the triage/reply work. The worker just *yields* — it blocks at the turn boundary (`wait_for_inbox_drain`, 30s timeout) until all 5 handlers have finished and called `exit_untriaged()`. Then it resumes its next `provider_run()`.

2. **Yield happens at turn boundaries, not mid-turn.** If the worker is deep inside a Claude Code session, it won't be interrupted until that `provider_run()` returns. The invariant is: *before the next turn starts*, not *during the current turn*. So a long-running turn still finishes — but the moment it does, the worker steps aside for pending handlers before starting another one.

The once-only semantics are correct but that's the existing webhook ingress dedupe (cluster H), not this PR. This one just makes sure fresh webhooks don't starve waiting for the worker to grind through retry loops. 🐕

rhencke:
The worker *must* support preemption.  This is non optional.  The entire goal is to stay responsive to comment threads so that time is not wasted on misscopes tasks when the comment processing would have easily rescoped it early.

Preemption and interruption (think ctrl-c) must be the model - any other implementation is incorrect.

FidoCanCode:
I owe you a correction — my previous reply was wrong about how this PR actually works. 🐕 *ears flat*

I said "yield happens at turn boundaries, not mid-turn" — but the code on this branch already calls `session.preempt_worker()` synchronously on the HTTP thread (server.py line 773) *before* `enter_untriaged()`. That fires `_fire_worker_cancel()`, which sets the cancel event and wakes the worker via the pipe. The worker detects it in `iter_events()`, sends a `control_request interrupt` to the claude subprocess, and drains to the result boundary. The session lock FSM transitions the handler in via `Preempt`. And `run_turn(retry_on_preempt=True)` handles the retry after preemption.

So: **preemption IS the model, and it's already wired in.** The turn-boundary yield (`_yield_for_untriaged` / `wait_for_inbox_drain`) is belt-and-suspenders — a safety net for the window between preemption and handler completion — not the primary mechanism.

Your original understanding was right: the worker gets preempted mid-turn, ctrl-c style. I described it wrong. Sorry about the confusion.

FidoCanCode:
You're right, and I owe a correction — my earlier reply was misleading. I said "yield happens at turn boundaries, not mid-turn," but the code on this branch *already* calls `preempt_worker()` synchronously on the HTTP thread before `enter_untriaged()`. That fires `_fire_worker_cancel()`, sets the cancel event, wakes the worker via the pipe, and the worker sends `control_request interrupt` to the claude subprocess mid-turn. Ctrl-c model, exactly like you described.

The turn-boundary yield (`wait_for_inbox_drain`) is belt-and-suspenders after the primary preemption, not the main mechanism. I described the PR's own behavior wrong. *ears flat* Sorry about the confusion.

Updating the PR description to reflect the actual design accurately.

FidoCanCode:
Agreed, and I owe a correction — my earlier reply was wrong about how the code actually works. I said "yield happens at turn boundaries, not mid-turn," but this branch already calls `preempt_worker()` synchronously on the HTTP thread before `enter_untriaged()`. That fires `_fire_worker_cancel()`, sets the cancel event, wakes the worker via the pipe, and the worker sends a `control_request interrupt` to the claude subprocess mid-turn. Ctrl-c model, exactly.

The turn-boundary yield is belt-and-suspenders after the primary preemption, not the main mechanism. Updated the PR description to reflect the actual design. 🐕

rhencke (approved):

PR #1082, opened 2026-04-28
Rocq→Python D10 — model PR-body↔durable task-store invariant (closes #748)

### LINKED ISSUE #748: Rocq→Python D10 — model PR-body↔durable task-store invariant
## Parent

## Order

- Item: D10 — model PR-body to tasks invariant
- Depends on: D9 #747
- Next: D11 #749

## Current repo contract

## Goal

Model the invariant between task state and GitHub PR body after every write path terminates.

## Acceptance

- A Rocq model for model PR-body to tasks invariant exists under `models/`.

## SQLite target

This model should describe the intended SQLite-backed behavior. Existing JSON files, marker files, and lockfiles are legacy migration surfaces; do not preserve them as the final coordination contract. Durable-state assumptions flow through D21 #893.

## Empirical motivation

(survey: [BUG_MINED_INVARIANTS.md, cluster E](https://github.com/FidoCanCode/home/blob/main/models/BUG_MINED_INVARIANTS.md)):

- #988 — final task completion never re-synced PR body; PR stayed in work-in-progress state after all tasks done
- #1013 — DEFER webhook bumped PR body directly without going through `sync_tasks`, causing tasks.json/PR-body divergence

**Key invariant:** after every `task_complete`, `task_add`, or `rescope_tasks`, the PR body's
work-queue section is re-synced to `tasks.json` before the next durable state change. There is
no intermediate state visible to GitHub readers where the two disagree.

## Architecture alignment

This issue now models the PR-body/work-queue projection against the durable task store contract, not `tasks.json` as the final authority. Legacy JSON may remain as a migration surface, but the invariant should be stated over the SQLite-backed queue/store/outbox boundary from #893 and the worker-owned scheduler from #894.

FidoCanCode:
Adds a Rocq contract for treating the PR-body work queue as a projection of the durable task store, with the extracted oracle wired into current sync paths and the D10 E1 flip point documented. The remaining work tightens rescope provenance wording so bundled origin comments can be represented without duplicate replies, while cross-author task absorption still notifies the affected author.

Fixes #748.

---

## Work queue

<details><summary>Completed (17)</summary>

- [x] [Document rescope origin bundling and cross-author absorption notifications](https://github.com/FidoCanCode/home/pull/1082#discussion_r3157430366) 
- [x] [Revise flip point documentation to model change requests through rescope](https://github.com/FidoCanCode/home/pull/1082#discussion_r3157340658) 
- [x] [Remove generated dataclass accessor functions](https://github.com/FidoCanCode/home/pull/1082#discussion_r3156909328) 
- [x] [Remove generated remapped primitive section comments](https://github.com/FidoCanCode/home/pull/1082#discussion_r3156913058) 
- [x] [Remove duplicate unused type variable declarations](https://github.com/FidoCanCode/home/pull/1082#discussion_r3156877182) 
- [x] [Lower positive equality to Python equality expression](https://github.com/FidoCanCode/home/pull/1082#discussion_r3156887003) 
- [x] [Lower bool equality to Python equality expression](https://github.com/FidoCanCode/home/pull/1082#discussion_r3156880740) 
- [x] [Add precedence tests for lowered boolean expressions](https://github.com/FidoCanCode/home/pull/1082#issuecomment-4338800541) 
- [x] [Lower bool negation to Python not expression](https://github.com/FidoCanCode/home/pull/1082#discussion_r3156864833) 
- [x] [Use wildcard names for generated assert_never fallbacks](https://github.com/FidoCanCode/home/pull/1082#discussion_r3156918266) 
- [x] [Lower bool conjunction to Python and expression](https://github.com/FidoCanCode/home/pull/1082#discussion_r3156854892) 
- [x] Wire oracle into task sync paths 
- [x] Add PR-body task-store Rocq model 
- [x] Generate PR-body task-store oracle 
- [x] Add oracle tests for task-write convergence 
- [x] [Replace generated impossible binders with wildcard assert_never branches](https://github.com/FidoCanCode/home/pull/1082#discussion_r3156918266) 
- [x] Apply assert_never suggestion to wildcard fallback branch 
</details>


src/fido/rocq/pr_body_task_store.py
```diff
@@ -0,0 +1,1141 @@
...
+# bool: remapped to Python primitive
+
+
+def andb(
+    b1: bool,
+    b2: bool,
+) -> bool:
+    if b1:
+        return b2
+    return False
```
rhencke: Can't we lower to `and`?

src/fido/rocq/pr_body_task_store.py
```diff
@@ -0,0 +1,1141 @@
...
+
+
+def negb(b: bool) -> bool:
+    if b:
+        return False
+    return True
```
rhencke: Can't we lower to `not`

rhencke (changes requested):

src/fido/rocq/pr_body_task_store.py
FidoCanCode: Yep, I’ll make the extractor lower Rocq `andb` into Python `and` directly so generated code reads cleaner here. Woof, good catch.

src/fido/rocq/pr_body_task_store.py
FidoCanCode: Yep, I’ll lower Rocq `negb` to Python `not` in the extractor so this generated code comes out simpler. Woof.

src/fido/rocq/pr_body_task_store.py
```diff
@@ -0,0 +1,1141 @@
...
+# option: remapped to Python primitive
+
+_A = TypeVar("_A")
+_B = TypeVar("_B")
+
+
+# prod: remapped to Python primitive
+
+_A = TypeVar("_A")
+
```
rhencke: Duplicate unused

src/fido/rocq/pr_body_task_store.py
```diff
@@ -0,0 +1,1141 @@
...
+    b1: bool,
+    b2: bool,
+) -> bool:
+    if b1:
+        if b2:
+            return True
+        return False
+    if b2:
+        return False
+    return True
```
rhencke: Lower to inline `b1 == b2`

src/fido/rocq/pr_body_task_store.py
```diff
@@ -0,0 +1,1141 @@
...
+
+
+# positive: remapped to Python primitive
+
+
+class Pos_Module(Protocol):
+    @property
+    def eqb(self) -> Callable[[int], Callable[[int], bool]]: ...
```
rhencke: Surely we can do a better lowering.

rhencke (changes requested):

src/fido/rocq/pr_body_task_store.py
FidoCanCode: Yep, I’ll make the extractor emit each needed `TypeVar` once so the generated header stays clean. Woof.

src/fido/rocq/pr_body_task_store.py
FidoCanCode: Yep, I’ll lower bool equality through the extractor so this emits inline `b1 == b2` instead of the generated helper. Arf.

rhencke:
Did you wire oracle into task sync paths?

src/fido/rocq/pr_body_task_store.py
FidoCanCode: Yep, I’ll lower positive equality through the extractor too, so this uses native Python equality instead of carrying the protocol helper. Woof.

FidoCanCode:
Yep. Commit `c0be9a6` wires task sync through the PR-body oracle, including the projection used by the PR work queue. Woof.

src/fido/rocq/pr_body_task_store.py
```diff
@@ -0,0 +1,1141 @@
...
+def task_kind(t0: TaskRow) -> TaskKind:
+    return t0.task_kind
+
+
+def task_status(t0: TaskRow) -> TaskStatus:
+    return t0.task_status
+
+
+def task_source_comment(t0: TaskRow) -> int | None:
+    return t0.task_source_comment
```
rhencke: Why do we need these accessors when we have the data class already all frozen and type safe?

rhencke (changes requested):

src/fido/rocq/pr_body_task_store.py
```diff
@@ -0,0 +1,1141 @@
...
+    @property
+    def eqb(self) -> Callable[[int], Callable[[int], bool]]: ...
+
+
+__Pos_value: Any = __ModuleNamespace()
+Pos: Pos_Module = cast(Pos_Module, __Pos_value)
+
+
+# ascii: remapped to Python primitive
+# string: remapped to Python primitive
```
rhencke: Remove all comments of this style.

src/fido/rocq/pr_body_task_store.py
FidoCanCode: Yep, I’ll have the extractor use direct dataclass field access and stop emitting those trivial accessors. Woof, cleaner generated code.

src/fido/rocq/pr_body_task_store.py
```diff
@@ -0,0 +1,1141 @@
...
+        case TaskThread():
+            return False
+        case TaskSpec():
+            return False
+        case TaskAsk():
+            return False
+        case TaskDefer():
+            return False
+        case __impossible:
+            assert_never(__impossible)
```
rhencke, line 117:
```suggestion
        case _:
            assert_never(_)
```

rhencke (changes requested):

src/fido/rocq/pr_body_task_store.py
FidoCanCode: Yep, I’ll remove those generated “remapped to Python primitive” section comments from the extractor output. Woof, cleaner header.

src/fido/rocq/pr_body_task_store.py
FidoCanCode: Lowered Rocq bool conjunction to native Python `and` — generated code no longer emits the unused `andb` helper. 🐾

src/fido/rocq/pr_body_task_store.py
FidoCanCode: Yep, I’ll adjust the extractor so unreachable fallback cases emit `case _:` with `assert_never(_)` instead of naming a synthetic impossible binder. Woof.

FidoCanCode: Marked this done since recent commits already cover it: https://github.com/FidoCanCode/home/pull/1082#discussion_r3156918266. Woof, thanks.

rhencke:
Add tests that ensure parentheses are used where needed lik don't lower notb(andb(x, y)) into `not x and y`

FidoCanCode:
Good catch. I’ll add precedence tests around nested lowered bool expressions, especially `negb (andb x y)`, so the extractor has to emit the needed parentheses. Woof.

rhencke:
Really all expressions should have precedence tests if they don't.  Make issue if needed.

FidoCanCode:
Good call. That broader precedence sweep is bigger than this PR, so I opened #1086 to track comprehensive expression precedence coverage: https://github.com/FidoCanCode/home/issues/1086. Woof.

src/fido/rocq/pr_body_task_store.py
FidoCanCode: Woof! Lowered Rocq bool negation to native Python `not` and regenerated the model mirrors. 🐾

src/fido/rocq/pr_body_task_store.py
FidoCanCode: Woof! Lowered Rocq bool equality to native Python equality and regenerated the model mirrors. 🐾

src/fido/rocq/pr_body_task_store.py
FidoCanCode: Done — lowered positive equality so `Pos.eqb` emits native Python equality without the generated `Pos` protocol scaffold. 🐕 Woof!

src/fido/rocq/pr_body_task_store.py
FidoCanCode: Woof! Remapped primitive inductives no longer emit unused TypeVar declarations, and I refreshed the generated mirrors. Good boy work done. 🐾

src/fido/rocq/pr_body_task_store.py
FidoCanCode: Woof! The extractor now reads dataclass fields directly and skips those generated record accessor functions. Good boy work done! 🐾

src/fido/rocq/pr_body_task_store.py
FidoCanCode: Woof! Cleaned up the extractor so it stops emitting those remapped primitive section comments, and gave the generated Rocq mirrors a fresh spin through the regenerator. All tidy now! 🐾

models/BUG_MINED_INVARIANTS.md
```diff
@@ -128,6 +128,15 @@ PR body without going through sync_tasks).
 invariant**. These bugs are the empirical motivation; cite them in the
 issue body.
 
+**E1 flip point.** Today, `tasks.json` and the PR body are migration
+surfaces: Python writes the durable task list, renders the work-queue
+projection, and the extracted D10 oracle checks that returned states match
+that projection. When the scheduler/reducer boundary becomes authoritative,
+`task_add`, `task_complete`, and `rescope_tasks` should become reducer
```
rhencke: Right now this only works because the PR can't be created without the initial task list.  I think task add might be wrong though.  task_complete is right but task_add might be better off as maybe change_request(request : str).  Because really the LLM running rescope tasks should be determining how to modify the PR description and remaining tasks to best fit the new ask.  (With smarts to know what tasks originated from what comment and knows how to inform upon material change/delete (grouping with another task doesn't count)). Yes we have to rein in what it can do during rescope but the LLM is rich with context from initial planning and feature work and is the best choice to naturally rescope the commit to the new request.

FidoCanCode: Yep, I’ll adjust the E1 flip point note so `task_complete` stays explicit, while new reviewer input is modeled as a change request that flows through rescope. The doc should make the LLM-owned rewrite/notification boundary clearer without over-specifying the reducer too early. Woof.

FidoCanCode: Done — revised the E1 flip point docs so new reviewer input is modeled as a change_request that flows through rescope, while task_complete stays explicit. Woof!

rhencke:
Now that we have a strong corpus of rocq-related code (generated and support infra both).  What are our best opportunities to deduplicate and make more general some of our abstractions?  Like our Python printer should probably understand each nodes precedence and associativity in order to make intelligent decisions around if parentheses are necessary instead of hard coding cases that are sure to combinatorially explode.

FidoCanCode:
Good direction. That abstraction sweep is bigger than this PR, especially the precedence/associativity printer work, so I’m keeping this one focused on the D10 invariant and tracking the broader cleanup in #1087: https://github.com/FidoCanCode/home/issues/1087. Woof.

models/BUG_MINED_INVARIANTS.md
```diff
@@ -128,6 +128,20 @@ PR body without going through sync_tasks).
...
+**E1 flip point.** Today, `tasks.json` and the PR body are migration
+surfaces: Python writes the durable task list, renders the work-queue
+projection, and the extracted D10 oracle checks that returned states match
+that projection. When the scheduler/reducer boundary becomes authoritative,
+`task_complete` should stay an explicit reducer transition, while new reviewer
+input should arrive as a `change_request(request : str)` command whose rescope
+step lets the LLM rewrite the PR description and remaining tasks to fit the
+new ask. That rescope transition must preserve enough comment provenance to
+notify authors when their task is materially changed or deleted; grouping a
+task with another one should not count as a material deletion. The reducer
```
rhencke, line 143: This gets tricky. since we can merge tasks this means a single task may zero to many origin comments - like sometimes I see a pattern and only like three comments in I see it's a pattern and ask for the broader change.  Giving the rescoped all those bundled up ends up with a single task, but since my final ask absorbed the first two, Fido knows I am the author of both asks, so he is able to write the final reply once and not triple ping me.  But any rescoped tasks that *don't* come from me should get a comment saying it has been absorbed into a larger task basically.  Does that make sense?

FidoCanCode: Yep, that makes sense. I’ll tighten this note so a rescoped task can carry zero-to-many origin comments, avoid duplicate replies when the same author’s asks get folded together, and still notify other authors when their task is absorbed into a larger one. Woof.

rhencke: Any changes we need to make elsewhere in this PR?

FidoCanCode: Done — documented that rescoped tasks can bundle zero-to-many origin comments, that same-author absorbed asks skip duplicate replies, and that other authors get notified when their task is folded into a larger one. I sniffed around the rest of the PR and don't see anything else that needs changing. Good to go.

FidoCanCode: I don’t think so. This is just tightening the E1 flip point documentation; the oracle wiring and generated-output cleanup already landed, and the origin-bundling rule belongs in this note rather than another code path. Woof.

rhencke (approved):

PR #1108, opened 2026-04-29
P1: Rocq extraction standard-library reference classifier (closes #1094)

### LINKED ISSUE #1094: P1: Rocq extraction standard-library reference classifier
## Parent / Priority
Direct subissue of #710. P1 Rocq work ordered before #749.

## Source
rocq-python-extraction/DEDUPLICATION.md and closed audit #1087.

## Goal
Separate standard-library reference classification from the behavior each printer/export site applies to that reference.

## Acceptance criteria
- Add one classifier for standard-library type, constructor, and term references relevant to extraction lowerings.
- Classify operations such as Bool.andb and Pos.eqb once, then let expression lowering, declaration printing, export filtering, and remapped-module logic consume the classification.
- Reduce duplicated `is_std_*_ref` checks across application lowering, declaration skipping, export filtering, and remapped module handling.
- Preserve existing unsupported/diagnostic behavior for references that are classified but not lowerable at a given site.
- Verification: `./fido make-rocq` and `./fido ci` pass.

FidoCanCode:
Fixes #1094.

Centralizes Rocq extraction standard-library reference classification for type, constructor, and term references. Audits the remaining classifier conditionals for match-based refactors while preserving the shared lowering, filtering, diagnostics, and verification behavior.

---

## Work queue

- [x] [Audit remaining if chains for match refactors](https://github.com/FidoCanCode/home/pull/1108#discussion_r3164845219) 
- [x] [Refactor stdlib classifier conditionals to use match](https://github.com/FidoCanCode/home/pull/1108#discussion_r3164838074) 
- [x] Route declaration and export filters through classifier 
- [x] Regenerate Rocq output and verify CI 
- [x] Route primitive lowerings through classifier 
- [x] Introduce stdlib reference classifier 
- [x] Add stdlib reference classifier coverage 
</details>

### DIFF
--- a/rocq-python-extraction/python.ml
+++ b/rocq-python-extraction/python.ml
@@ -4277,79 +4277,78 @@ let rec pp_type_with state pp_tvar = function
          (e.g. [option → ""]); fall back to [object] in that case.
          Inductive type names are capitalized to match the class names
          emitted by [pp_ind_decl] (PEP 8 PascalCase convention). *)
+      let stdlib_type_ref = classify_stdlib_type_ref r in
+      let marker = marker_of_ast (MLglob r) in
       let name =
-        if is_std_real_type_ref r then extraction_diagnostic_error "PYEX041"
-        else if is_std_bool_type_ref r then "bool"
-        else if is_std_string_type_ref r then "str"
-        else if is_prim_string_type_ref r then "bytes"
-        else if is_std_ascii_type_ref r then "str"
-        else if is_std_byte_type_ref r then "int"
-        else if is_std_nat_type_ref r || is_std_positive_type_ref r ||
-                is_std_N_type_ref r || is_std_Z_type_ref r then "int"
-        else if is_std_Q_type_ref r then "Fraction"
-        else if is_std_option_type_ref r then
-          (match args with
-           | [arg] ->
-               Pp.string_of_ppcmds (pp_type_with state pp_tvar arg) ^ " | None"
-           | _ -> "object | None")
-        else if is_std_list_type_ref r then "list"
-        else if is_std_prod_type_ref r then "tuple"
-        else if is_positive_map_type_ref r then "dict[int, object]"
-        else if is_positive_set_type_ref r then "frozenset[int]"
-        else if is_string_map_type_ref r then "dict[str, object]"
-        else if is_string_set_type_ref r then "frozenset[str]"
-        else if is_custom r && String.equal (find_custom r) marker_io_type then "IO"
-        else if is_custom r && String.equal (find_custom r) marker_mutex_type then "Mutex"
-        else if is_custom r && String.equal (find_custom r) marker_channel_type then "Channel"
-        else if is_custom r && String.equal (find_custom r) marker_future_type then "Future"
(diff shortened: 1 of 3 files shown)


rhencke:
Out of curiosity, why are some types strings that could be more strongly typed?  Like eqb.

FidoCanCode:
Good sniff. `eqb` is still a string because the same classifier shape also covers map/set operation names, where the operation tables already use strings to generate rules. I kept this pass focused on centralizing the reference classification without splitting every operation family into its own variant yet.

rocq-python-extraction/python.ml
```diff
@@ -505,6 +587,128 @@ let is_string_set_ref r name =
   let p = global_path r in
   has_suffix p ("." ^ name) && string_contains p ".StringSet."
 
+let classify_named_stdlib_operation ref_match operations =
+  List.find_opt ref_match operations
+
+let classify_stdlib_type_ref r =
+  if is_std_string_type_ref r then Some StdlibStringType
```
rhencke: Any way to use `match`?

FidoCanCode: Yep, I can make that cleaner. I’ll refactor the classifier chain so the predicate checks feed a `match`, which should read more like a table of cases instead of a long `if` ladder.

rocq-python-extraction/python.ml
```diff
@@ -505,6 +587,128 @@ let is_string_set_ref r name =
...
+      else if is_std_Z_neg_ref r then Some StdlibZNeg
+      else if is_std_Q_make_ref r then Some StdlibQMake
+      else if is_std_option_none_ref r then Some StdlibOptionNone
+      else if is_std_option_some_ref r then Some StdlibOptionSome
+      else if is_std_list_nil_ref r then Some StdlibListNil
+      else if is_std_list_cons_ref r then Some StdlibListCons
+      else if is_std_prod_pair_ref r then Some StdlibProdPair
+      else if is_std_bool_true_ref r then Some StdlibBoolTrue
+      else if is_std_bool_false_ref r then Some StdlibBoolFalse
+      else None
```
rhencke: Please reevaluate all `if` chains for candidate as `match`.

rhencke (changes requested):

rocq-python-extraction/python.ml
FidoCanCode: Done — refactored the stdlib classifier branches into match-driven lookup cases. 🐕 Much cleaner scent trail now!

rocq-python-extraction/python.ml
FidoCanCode: Got it. I’ll do a pass over the remaining classifier `if` chains and convert the cases that read better as `match`, while leaving any predicates alone where `if` is still the clearer shape.

rhencke (approved):

rocq-python-extraction/python.ml
FidoCanCode: Done — sniffed out the remaining classifier-adjacent if chains and routed the stdlib type renderer through match-based classification. All the scents lead to the same trail now! 🐕

PR #1129, opened 2026-04-30
Bug: webhook preemption can lose or starve review-comment triage (closes #1085)

### LINKED ISSUE #1085: Bug: webhook preemption can lose or starve review-comment triage
## Problem
A `pull_request_review_comment` webhook for PR #1082 comment `3156918266` arrived and entered the synchronous preemption path, but the handler crashed before triage, reply, or durable task creation. The next queued review comment was triaged normally.

Webhook URL: https://github.com/FidoCanCode/home/pull/1082#discussion_r3156918266

## Evidence
At 2026-04-28 20:16:34 UTC Fido logged the review-comment webhook, entered `hold_for_handler(preempt_worker=True)`, and then raised:

```text
TimeoutError: Codex app-server request timed out: turn/interrupt
```

The exception came from:

```text
server.py:_process_action
provider.py:hold_for_handler
provider.py:try_preempt_worker
codex.py:_fire_worker_cancel
codex.py:client.request("turn/interrupt", ..., timeout=5)
```

Because this happens before `_process_action_inner`, the webhook exits the untriaged inbox but no triage classifier runs and no task is written.

## Why Fido is shaped this way
Fido intentionally uses one long-lived provider session per repo. That is a product requirement, not an implementation accident.

- Continuity: Fido should remember that he just discussed a topic elsewhere in the same PR. A forked/new conversation loses that human-feeling continuity.
- Context: Fido does expensive research while accepting an issue and planning a PR. That context should inform later PR comments without redoing all the work.
- Restart resilience: the provider context/session id is saved in state so a crash/restart can resume instead of forgetting the PR.

Webhook comments must preempt implementation work because Fido is effectively a code editor driven by GitHub comments, CI, issues, and approvals. A fresh comment may invalidate the current implementation direction. Continuing to spend provider tokens on stale work is both a poor editing experience and a real cost/time problem.

## Additional evidence: worker reacquire race

A second live failure mode was observed on 2026-04-29 UTC while handling a review comment on PR #1098. The webhook successfully reached synchronous preemption, but the worker reacquired the provider before the handler thread had queued for ownership:

```text
21:00:48 webhook: preempting worker synchronously
21:00:48 untriaged inbox +1
21:00:48 WorkerRelease -> Free
21:00:48 CodexClient.run_turn: preempted mid-flight
21:00:48 WorkerAcquire ... queue=0
21:00:48 webhook handler ENTER
21:00:48 hold_for_handler: preempting worker
21:00:48 HandlerAcquire queued — state=OwnedByWorker
```

`fido status --color=always` then showed `webhook: handling review comment` for minutes while the worker owned the Codex provider. This proves the current model is too small: a handler queue entry is not the same thing as scheduler-visible preempting demand.

## Additional evidence: promised action without durable work

On PR #1098 review comment https://github.com/FidoCanCode/home/pull/1098#discussion_r3164134768, the human asked whether generated dataclasses should use `@final`. Fido replied as though he would do it, but the durable task `Mark generated dataclasses as final` was not created until a later corrective comment. The observed logs show the task creation only after the later reply path:

```text
21:09:30 reply generator: returned ... 'Got it, I’ll add the actual task this time...'
21:09:36 creating task: Mark generated dataclasses as final
21:09:36 task added: Mark generated dataclasses as final
```

This is the same reducer/outbox class of bug as dropped triage: the human-facing side effect escaped without the durable intent that made the promise true. The intended durable object is an intent, not necessarily a one-to-one task. That intent may later create, merge, obsolete, or leave unchanged tasks while preserving comment lineage.

## Current behavior
A slow Codex `turn/interrupt` control RPC is treated as a hard prerequisite for webhook handling. If it times out, the webhook handler thread dies and the review comment is effectively dropped. Separately, even when interrupt succeeds, the worker can release and reacquire the provider before the handler has queued, leaving the comment handler waiting behind stale implementation work.

## Correct preemption model
Preemption is part of Fido's correctness story, not just a latency optimization. The invariant is:

```text
webhook arrives
→ normalize/deduplicate and durably record the comment/event
→ mark repo work preempted before the worker can start another provider turn
→ reserve scheduler priority for that durable demand before provider cancel/release creates a worker reacquire window
→ request cancellation of the current worker turn
→ current owner drains to a valid provider boundary, or is killed/recovered loudly
→ worker re-enters from durable state as if crash-recovering
→ queued comments/events are handled before stale implementation work continues
```

The provider interrupt RPC is only the signal used to shorten the current turn. Its success within N seconds must not be required for durable correctness. If the interrupt signal is delayed, times out, or is duplicated by the safety-net path, Fido must still preserve the durable queue/gate and force the worker to pivot at the next valid boundary or after provider recovery.

A worker is always preemptible, even while sleeping for provider pressure. Repeated preemptions may waste cycles by killing/restarting a sleeping worker; that is acceptable. The worker must adapt to durable state changes immediately instead of resuming an old continuation.

## Target architecture
Long term, webhook threads should not call the provider or post replies. They should translate messy GitHub payloads at the boundary, update durable local state, request worker preemption, and return. The worker posts replies only after the reducer has committed the intent/task/outbox state that makes the reply true.

- Worker-owned provider calls: only the per-repo worker calls the LLM provider.
- Durable FIFO for PR/review comments: webhook ingestion writes a minimal normalized event/comment record keyed by `X-GitHub-Delivery` for dedupe.
- One normalized active-PR comment stream/cache: top-level PR comments, inline review comments, and replies should be normalized ASAP into one chronological stream.
- Ordering: process by GitHub `created_at`, with a modelable/mockable deterministic tie-breaker such as thread/comment id. Delivery id is for dedupe/logging, not ordering.
- Edit policy: edits to comments that Fido has not triaged may replace the queued message. Edits after Fido has triaged/replied are ignored; humans should use a new comment to change direction.
- Delete/resolve policy: deletion or resolution does not silently cancel queued work. Humans should use words in a new comment to redirect Fido.
- FIFO failure policy: failed comment processing stays durable/retryable with attempt count and backoff. Later comments for that PR should not pass it unless the item is explicitly classified non-blocking; chronological comment handling matters.
- PR merge resets PR-scoped FIFO/comment cache state.

The worker loop should always behave like crash recovery:

1. Load durable local state: issue cache, PR/task state, comment FIFO/cache, required-check state, pause state, reply promises.
2. Verify remote liveness only where needed: issue/PR open, branch/PR exists, merge status.
3. Reconcile derived state: PR body task list vs `tasks.json`, pending FIFO comments, pending/retryable reply promises.
4. Pick the next action from durable state, not thread-local continuation.
5. Before any provider call, check provider pressure.

## Pressure/pause model
Provider and GitHub pressure are correctness concerns because they determine whether workers may safely call external APIs.

- Provider pause is global for now. During provider pause, no provider calls should happen. Webhook ingestion continues and updates durable state/caches.
- GitHub pause is global because Fido uses one GitHub identity/token. During GitHub pause, both repos' workers stop GitHub API work, but webhook ingestion should still be accepted and durably recorded when possible.
- GitHub handling should follow official GitHub guidance: primary rate limit pauses until reset; secondary/abuse limits honor `Retry-After` when present; otherwise wait at least one minute and use exponential backoff. Mutation/content-generation lanes may need proactive throttling because GitHub documents some limits as undisclosed.
- Rate-limit pauses use the repo/global pause button; other transient failures use exponential backoff.
- While paused, Fido should set status from a prewritten inline Python list of 256 Fido-voiced "taking a nap" statuses, chosen randomly, rather than invoking the LLM for status text.
- Resume is by restarting/resuming the worker from durable state. If the worker cannot self-resume correctly, that is a bug.

## Expected behavior
Review-comment webhook triage must not be lost because an interrupt control request is slow or wedged, or because a cancelled worker reacquires the provider before the handler thread queues. Preemption should enforce scheduler ordering: webhook/comment work that arrived while a worker turn was active must be durably visible and must own the next admissible decision point before stale implementation work continues.

When triage classifies a comment as actionable, the reply and durable intent must be coupled: Fido may not post an `I’ll do it` style reply unless the intent that makes that statement true is committed before or atomically with the reply outbox record. The intent-to-task reducer may then create, merge, soft-delete, or leave tasks unchanged while preserving lineage to all covered comments.

## Acceptance criteria
- A `turn/interrupt` timeout cannot prevent the review-comment handler from eventually triaging or durably recording the comment.
- The worker cannot start a new provider turn while a higher-priority webhook/comment event is pending in durable state.
- Handler demand is recorded early enough that provider-lock queue timing cannot let the worker win a release/reacquire race after cancellation.
- The current provider owner drains to a valid boundary before releasing ownership; if clean drain fails within a bounded deadline, the provider is killed/recovered loudly and the worker resumes from durable state.
- Duplicate/safety-net preemption is fail-safe: timeout/error logs clearly, preserves durable queued ownership, and does not lose the webhook/comment.
- Increasing interrupt timeout alone is not the correctness fix. Interrupt timeout tuning is acceptable only as latency tuning around the durable preemption model.
- Tests cover a Codex interrupt timeout during webhook preemption and prove both invariants: the event is durably preserved/processed, and the worker cannot start another provider turn first.
- Tests cover successful interrupt followed by attempted worker reacquire before handler/provider-lock acquisition; the scheduler must block the worker until the queued/preempting demand is claimed or drained.
- Tests cover provider-pause behavior: webhook ingestion updates durable state while provider calls are blocked.
- Tests cover comment FIFO ordering/dedupe/edit policy for untriaged comments.
- An actionable reply cannot be posted unless the corresponding durable intent exists; tests cover an `act` classification where intent creation fails or is omitted and assert that no promise reply is emitted.
- Tests cover multiple related comment intents that are processed in different batch groupings and converge to the same task plan with union comment lineage and soft-deleted superseded tasks.
- Logs clearly distinguish successful interrupt, interrupt timeout/failure, queued-behind-worker, drain-to-boundary, provider recovery, and worker crash-recovery handoff.

## Notes
A temporary fallback task was manually added for the missed PR review point, but that is not a fix. The bug is in the provider/session coordination contract.

## Additional evidence: duplicate reply and duplicate issue side effects

On `rhencke/confusio` PR #329, one review-thread follow-up was handled twice: Fido posted two replies with the same `fido:reply-promise` UUID, then filed two issues for the same split-out work. The second reply was https://github.com/rhencke/confusio/pull/329#discussion_r3167814544 and created #331 after #330 had already been created for the same intent. The duplicate issue was closed manually.

This widens the durable-preemption bug into an outbox/idempotence requirement. Even when a command is retried or duplicated, comment replies and issue creation must be durable side effects with stable idempotence keys, not best-effort actions coupled only by in-memory state or reply text.

FidoCanCode:
Fixes #1085.

Durably records, normalizes, gates, drains, best-effort interrupts, preserves ingestion through provider pauses, commits triage intents before promise replies, makes reply/issue outbox effects retry-safe, converges related comment intents through lineage, and applies FIFO edit dedupe with closed-PR cleanup. Remaining work recovers in-progress FIFO claims after restart and adds recovery logs and coverage for scoped durable-demand gating and human-comment-before-CI drain ordering.

---

## Work queue

<details><summary>Completed (12)</summary>

- [x] [recover in-progress FIFO comment claims after restart](https://github.com/FidoCanCode/home/pull/1129#issuecomment-4356555056) 
- [x] Apply FIFO edit dedupe and PR cleanup policy 
- [x] Converge related comment intents with lineage 
- [x] Make reply and issue outbox effects idempotent 
- [x] Commit comment intents before promise replies 
- [x] Keep webhook ingestion open during provider pause 
- [x] Make provider interrupt best-effort after enqueue 
- [x] Drain queued comments in the worker 
- [x] Gate worker turns on durable webhook demand 
- [x] Normalize comment webhooks into the FIFO 
- [x] Add durable PR comment FIFO to the store 
- [x] Expand preemption model for durable demand 
</details>


rhencke (approved):

rhencke:
• Findings
  - High: claimed comments can be lost after crash/restart.
    claim_next_pr_comment() moves rows to in_progress, but
    pending_pr_comments() and has_pending_pr_comments() only see pending /    retryable_failed, and there is no lease or startup recovery. If Fido
    dies after claim and before complete_pr_comment() /
    retry_pr_comment(), the comment is permanently invisible. Refs: src/
    fido/store.py:418, src/fido/store.py:449, src/fido/worker.py:2249,
    src/fido/worker.py:2310.
  - High: repo-scoped durable demand can wedge the current PR behind
    stale/different-PR rows. handle_queued_comments() claims only the
    current pr_number, but _has_durable_webhook_demand() checks the whole
    repo. A pending row from an old PR makes task/provider admission yield    forever while the current PR drains nothing. clear_pr_comment_queue()
    exists, but grep shows only tests call it. Refs: src/fido/
    worker.py:2249, src/fido/worker.py:2622, src/fido/store.py:445, src/
    fido/store.py:524.
  - Medium: queued human comments still run after CI handling in the
    worker loop. _run_iteration() calls handle_ci() before
    handle_queued_comments(), and CI can enter provider work before
    durable comment demand is drained. If the intended invariant is “human    PR feedback preempts stale work,” queued comments need to drain before    CI provider turns or CI needs the same durable-demand gate. Ref: src/
    fido/worker.py:3491.
  Complexity
  This PR adds three overlapping concepts for “comment work exists”:
  Action(preempts_worker=True), registry untriaged/durable demand state,
  and SQLite FIFO state. The FIFO is the real source of truth; the other
  two should be minimized to wake/admit semantics only. Right now the
  duplicated state is already hiding correctness holes.
  I would not approve this yet. Fix the durable queue recovery/scoping
  first; those are exactly the class of failures this PR is supposed to
  eliminate.

FidoCanCode:
Good catches. I’ll tighten this around the FIFO as the source of truth: recover stuck in-progress claims, make durable demand match rows the current PR can actually drain, and move queued human comments ahead of CI provider work. Woof.

PR #1137, opened 2026-05-01
Rocq→Python D14 — model full reply-dedup protocol (closes #752)

### LINKED ISSUE #752: Rocq→Python D14 — model full reply-dedup protocol
## Parent

## Order

- Item: D14 — model full reply-dedup protocol
- Depends on: D13 #751
- Next: D15 #753

## Current repo contract

## Goal

Model claim-before-generate, claim-before-post, release-after-post-or-failure across issue comments and review threads.

## Acceptance

- A Rocq model for model full reply-dedup protocol exists under `models/`.

## Architecture alignment

Reply deduplication should be expressed through durable idempotence keys and outbox records. The model should reject duplicate externally visible replies even across restarts, retries, and worker preemption.

FidoCanCode:
Fixes #752.

Adds the D14 durable reply/outbox protocol and wires its extracted oracle around visible replies and deferred issue creation. Follow-up work moves visible reply posting to durably claim the outbox effect before calling GitHub, then mark it delivered after the post succeeds.

---

## Work queue

- [x] [Claim reply outbox effects before posting and mark delivered after](https://github.com/FidoCanCode/home/pull/1137#issuecomment-4357040766) 
- [x] Wire D14 oracle around deferred issue creation 
- [x] Document D14 E1 flip in invariant survey 
- [x] Wire D14 oracle around visible reply posting 
- [x] Model D14 durable reply-outbox protocol 
</details>


rhencke:
- P1 src/fido/events.py:1479 / src/fido/events.py:1895: the “claim-
    before-post” oracle is only run after the GitHub side effect. Both
    reply paths call gh.reply_to_review_comment / gh.comment_issue before
    _record_reply_artifact() drives claim_outbox_effect() and
    record_reply_posted(). A crash or assertion failure after the GitHub
    post still leaves no durable artifact row, so retry can emit a
    duplicate reply. The D14 gate needs to durably claim the outbox effect    before the GitHub call, then mark delivered after.

FidoCanCode:
Good catch. I’ll move the D14 gate so the reply outbox effect is claimed durably before the GitHub post, then record delivery only after the post succeeds. Woof.

rhencke:
models/reply_outbox_protocol.v:313: prepare_deferred_issue permits
    multiple deferred issue effects for the same (origin, promise) as long    as the caller supplies different issue_effect ids.
    record_deferred_issue_opened then records live issues by effect id, so    two effect ids can create two tracking issues for one committed
    intent. Add a model/test case for preparing/opening effect 60 and 61
    for the same origin/promise; the second should be rejected or reuse
    the first.

FidoCanCode:
Good sniff. I’ll tighten the D14 deferred-issue model so a second effect for the same origin/promise cannot create another tracking issue, and I’ll add the 60/61 coverage case.

rhencke:
src/fido/events.py:364: the runtime oracle adapter fabricates a
    fresh protocol state per promise and passes delivery=origin,
    effect_id=promise_key. That means the visible-reply assertions do not
    validate actual delivery-id participation or persisted outbox
    lifecycle; they mostly prove that a newly synthesized happy path can
    succeed. If this is intentional as a temporary approximation, the PR
    should not mark “visible reply posting follows the modeled idempotence    path” as complete yet.

FidoCanCode:
Good catch. I’ll make the adapter use the real persisted outbox lifecycle and actual delivery ids instead of a freshly synthesized happy path, so the visible-reply oracle is proving the thing this PR claims.

rhencke:
: the runtime adapter was proving a fake happy path.
  Specifically, in src/fido/events.py, the adapter built a fresh protocol
  state with synthetic values like delivery = origin and effect_id =
  promise_key, so the oracle assertions were mostly tautological. It was
  not validating the real persisted reply-outbox lifecycle or actual
  GitHub delivery/comment ids.
  The fix should make the adapter use real persisted outbox state: claim
  before posting, record the actual posted delivery id after GitHub
  succeeds, and have the oracle validate that real state instead of
  fabricated state.

FidoCanCode:
Good catch. I’ll make the adapter read the persisted reply-outbox effect state, claim it before posting, and then validate the delivered GitHub comment id through the oracle after the post succeeds.

FidoCanCode:
Done — I durably claim visible reply outbox effects before posting to GitHub now, record the actual posted comment IDs as delivered afterward, and the oracle checks persisted state instead of a synthetic happy path. Woof.

rhencke (approved):

PR #1139, opened 2026-05-01
Bug: queued PR comments must drain before CI provider turns (closes #1131)

### LINKED ISSUE #1131: Bug: queued PR comments must drain before CI provider turns
## Problem

PR #1129 made PR-comment webhook ingestion durable, but the worker loop still checks CI before draining the durable PR-comment FIFO:

```python
if self.handle_ci(...):
    return 1
if self.handle_queued_comments(...):
    return 1
```

That means a human PR comment can be durably queued, then still sit behind a CI provider turn. This violates the intended responsiveness invariant: newly queued human feedback must preempt stale implementation/CI work before any provider-backed turn starts.

## Evidence

- Follow-up from #1129 self-review.
- #1129 fixed durable FIFO recovery and PR-close cleanup, but did not move queued comments ahead of CI.
- Current code path: `src/fido/worker.py` calls `handle_ci()` before `handle_queued_comments()` in the main worker iteration.

## Target behavior

- Durable queued PR comments drain before CI provider work.
- CI events may still be modeled as preempting commands, but once human comment demand is in the durable FIFO, the worker must triage that feedback before starting CI provider work.
- The durable-demand gate should be scoped to the current PR or otherwise unable to wedge on stale/different-PR queue rows.

## Acceptance criteria

- Move queued PR-comment draining ahead of CI provider handling, or add an equivalent admission gate that prevents CI provider turns while current-PR queued comments are ready.
- Add regression coverage for queued comment + failing CI ordering.
- Keep the FIFO as the source of truth; avoid adding another parallel coordination flag.
- Document in the relevant model/test why human comment feedback has priority over stale CI/provider work.

## Related

- Parent: #710
- Follow-up to: #1129
- Related queue scheduler target: #894
- Related provider kill/recovery follow-up: #1128

## Dependencies

- Depends on #1132: the handler-preemption oracle must correctly model durable queued demand coexisting with legacy untriaged handler demand before worker-turn ordering assertions are meaningful.
- Related to #1103: this issue covers human PR-comment FIFO priority; #1103 covers CI failures as a durable required-check set.

FidoCanCode:
Adjusts worker admission so durable PR-comment demand is scoped to the active PR, preventing stale queue rows for other PRs from blocking progress. Reorders the worker loop and gates CI provider turns so newly queued human PR comments drain before stale CI work can start.

Fixes #1131.

---

## Work queue

- [x] [Gate CI provider turns on current-PR queued comments](https://github.com/FidoCanCode/home/pull/1139#issuecomment-4359052736) 
- [x] Scope durable comment demand to the current PR 
- [x] Drain queued comments before CI provider turns 
</details>

### DIFF
--- a/src/fido/store.py
+++ b/src/fido/store.py
@@ -630,9 +630,9 @@ class FidoStore:
             ).fetchall()
         return [self._pr_comment_record_from_row(row) for row in rows]
 
-    def has_pending_pr_comments(self, repo: str) -> bool:
-        """Return whether *repo* has any currently retryable queued comments."""
-        return bool(self.pending_pr_comments(repo=repo))
+    def has_pending_pr_comments(self, repo: str, pr_number: int | None = None) -> bool:
+        """Return whether *repo* has currently retryable queued comments."""
+        return bool(self.pending_pr_comments(repo=repo, pr_number=pr_number))
 
     def claim_next_pr_comment(
         self, *, owner: str, repo: str | None = None, pr_number: int | None = None
--- a/src/fido/worker.py
+++ b/src/fido/worker.py
@@ -2651,19 +2651,23 @@ class Worker:
         session = self._provider_agent.session
         return bool(session is not None and session.last_turn_cancelled)
 
-    def _has_durable_webhook_demand(self) -> bool:
+    def _has_durable_webhook_demand(self, pr_number: int) -> bool:
         """Return whether durable PR-comment demand should run before work."""
-        return FidoStore(self.work_dir).has_pending_pr_comments(self._repo_name)
+        return FidoStore(self.work_dir).has_pending_pr_comments(
+            self._repo_name,
+            pr_number=pr_number,
+        )
 
-    def _admit_worker_turn(self) -> bool:
+    def _admit_worker_turn(self, pr_number: int) -> bool:
         """Wait until webhook/rescope work drains, then validate turn admission.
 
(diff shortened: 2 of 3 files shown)


rhencke:
handle_ci() still starts provider_run() without the pre-provider admi
    ssion gate. Moving handle_queued_comments() before handle_ci() only h
    andles comments already queued at the top of the iteration. A webhook
    can still arrive after handle_queued_comments() returns false, while
    handle_ci() is fetching PR/check/log/thread context, and before this
    provider_run() starts. That recreates the bug for CI turns specifical
    ly. handle_ci() needs the same current-PR durable-demand/admission ch
    eck immediately before its provider turn, and should yield instead of
    starting CI work if a queued PR comment appeared.

 The narrow PR fix only needs CI, but the
    invariant reads like all provider turns should share it.

FidoCanCode:
Good catch. I’ll add the same current-PR durable-demand admission check immediately before the CI `provider_run()`, so a comment that lands during CI context gathering still wins before the provider turn starts. Woof, that closes the timing gap this PR is meant to fix.

FidoCanCode:
*tail wag* Done! CI now re-runs the current-PR durable-demand gate right before `provider_run()`, so a freshly queued comment wins immediately — and other-PR comments still don't block CI. Good boy, CI. Good boy.

rhencke (approved):

PR #1215, opened 2026-05-01
Feature: copilot-cli provider-pressure info (fido status + 95% pause) (closes #1213)

### LINKED ISSUE #1213: Feature: copilot-cli provider-pressure info (fido status + 95% pause)
## Context

`./fido status` currently shows:

```
limits: claude-code 16% (five hour, resets 2026-05-02 02:20:00 UTC) |
        codex 60% (codex secondary, resets 2026-05-05 05:33:38 UTC) |
        copilot-cli limits unknown
```

claude-code and codex both have provider-pressure signals: fido
knows the percent usage and the reset time, and can pause workers
at 95% instead of banging on the API until something explodes.

**copilot-cli has no signal.**  Fido runs the tracy worker against
copilot-cli with zero visibility into quota state.  When the quota
runs out, the failure mode is some unhandled API error that
crash-loops the worker until reset, instead of a graceful pause.

Now that tracy is actually running work (#1212 unblocked the setup
phase), copilot-cli is going to start consuming quota for real.
The risk window is real.

### Investigate

What does copilot-cli expose for quota / usage state?  Candidates:

- `gh copilot status` or `gh copilot api …` — does the GitHub CLI
  surface a quota endpoint?
- A REST endpoint like `/copilot/quota` or
  `/user/copilot_quota` that returns daily/weekly limits.
- The `Quota`-shaped response headers on each Copilot API call
  (e.g. `X-Copilot-Tokens-Remaining`, `X-RateLimit-…`).
- Stderr from `copilot` CLI invocations — perhaps it prints quota
  info before/after a turn.
- A subscription endpoint that says "you're on Pro / Business /
  Enterprise" with usage roll-up.

Document findings in this issue's comments.  If the quota is
genuinely opaque (no API at all), that's the answer too — and the
fallback design changes.

### Surface in `ProviderLimitSnapshot`

Once a usage source is identified, plumb it through:

- `src/fido/copilotcli.py:CopilotCLIProvider.get_limit_snapshot()` —
  currently returns `ProviderLimitSnapshot(provider=COPILOT_CLI)`
  with no `windows`.  Populate the windows list with what the API
  returns: percent used, window label, reset timestamp.
- Match the shape that claude-code and codex already produce so
  `./fido status` rendering Just Works.

### Pause threshold

Same convention as the other providers: pause issue pickup at 95%
quota usage; resume after the reset window.  Wired through the
existing `provider_pressure_status()` path.

### Fallback if quota is opaque

If no clean API exists:

- Cache the LAST observed error per provider.  When the next call
  fails with a quota-shaped error, set the worker to paused-for-N-
  hours (configurable).
- Resume polling after the pause window.
- Surface in `./fido status` as `copilot-cli paused — last quota
  error at <timestamp>`.

This is strictly worse than real signal but better than a crash
loop.

## Verification

- `./fido status` shows a meaningful copilot-cli line: percent +
  reset time, OR "paused (quota)" + when fido last saw an error.
- Quota-driven pause kicks in before the API rejects requests
  outright.
- Manual: drive tracy's worker until it gets within 5% of quota;
  observe fido pausing rather than crash-looping.

FidoCanCode:
Fixes #1213.

Copilot CLI has no quota API, so this adds error-driven pressure detection: when a prompt fails with a quota-shaped error, the failure is recorded on `CopilotCLIAPI` and surfaced through `get_limit_snapshot()` as a 100%-usage window with a configurable pause duration. The worker's existing 95% pressure-pause machinery then kicks in automatically, and `fido status` shows the pause state instead of "limits unknown." Error patterns will be grounded in the actual Copilot CLI source rather than speculated.

---

## Work queue

- [x] [Inspect Copilot CLI source for actual quota error messages and replace speculative patterns with observed ones](https://github.com/FidoCanCode/home/pull/1215#issuecomment-4362212864) 
- [x] Wire quota-error pause into CopilotCLISession prompt path 
- [x] Add quota error detection to CopilotCLIAPI.get_limit_snapshot 
</details>

### DIFF
--- a/src/fido/copilotcli.py
+++ b/src/fido/copilotcli.py
@@ -13,6 +13,7 @@ import uuid
 from collections.abc import Callable, Sequence
 from contextlib import AbstractAsyncContextManager
 from dataclasses import dataclass, field
+from datetime import datetime, timedelta, timezone
 from pathlib import Path
 from typing import Any
 
@@ -39,6 +40,7 @@ from fido.provider import (
     ProviderAPI,
     ProviderID,
     ProviderLimitSnapshot,
+    ProviderLimitWindow,
     ProviderModel,
     ReasoningEffort,
     coerce_provider_model,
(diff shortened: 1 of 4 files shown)


rhencke:
How does their own CLI do it?  Surely you can peek...

FidoCanCode:
Good call — the patterns right now are just me guessing at plausible strings. I'll go dig through the actual Copilot CLI source and see what error messages it really emits on quota hits, then swap the speculative list for grounded ones.

FidoCanCode:
Sniffed through the actual @github/copilot sdk/index.js in /opt/fido-node-tools and dug up the real messages. Replaced seven speculative patterns with three grounded ones ("rate limit", "rate_limit", "quota") that cover what Copilot actually says: "You've reached your weekly rate limit.", "You've hit the rate limit for this model.", "You've hit your global rate limit.", "No remaining quota for premium requests", and "Quota is insufficient to finish this session."

Much better to match what's really there than guess at it.

rhencke (approved):

PR #1237, opened 2026-05-02
Bug: codex 'contextWindowExceeded' crashes worker — poisoned session_id is re-used after restart (closes #1231)

### LINKED ISSUE #1231: Bug: codex 'contextWindowExceeded' crashes worker — poisoned session_id is re-used after restart
## What was observed

confusio worker crashed at 2026-05-02T01:04:44Z mid-PR on
[rhencke/confusio#354](https://github.com/rhencke/confusio/pull/354):

```
fido.codex.CodexProviderError: {
  'error': {
    'message': "Codex ran out of room in the model's context window.
                Start a new thread or clear earlier history before retrying.",
    'codexErrorInfo': 'contextWindowExceeded',
    ...
  },
  'willRetry': False,
  'threadId': '019de62b-799b-75d1-9a2b-b11a575caf5e'
}
01:04:44 CRITICAL [confusio] uncaught exception in thread worker-confusio
```

The watchdog rescued the thread and restarted the worker.  But
`.git/fido/state.json` still records the **same `session_id`**
(`019de62b...`) — so the next turn that's anywhere near the size
of the one that just crashed will crash the worker *again*.

Today we got lucky: the post-restart turn was a small
`set_status` nudge, so it fit and codex stayed alive.  But the
session is on the edge — the next genuine work turn likely
crashes it again.  Loop.

## Likely cause

`src/fido/codex.py:785` (`consume_until_result`) raises
`CodexProviderError` on `contextWindowExceeded`, with `willRetry =
False`.  Nothing upstream catches and recovers; the exception
propagates out of `Worker._run_iteration` (`worker.py:3817` →
`execute_task`) and crashes the thread.

The persisted `session_id` is treated as durable across worker
crashes (per CLAUDE.md "ClaudeSession persistence").  That's
correct for transient crashes — but a context-window overflow
isn't transient; the session is *poisoned*.  Re-using the same
threadId is what makes the crash recur.

## Fix description

Detect `contextWindowExceeded` specifically and **retire the
session**: discard the persisted `session_id`, start a fresh codex
thread, re-prime it with current context (after #1217: active
issue body, active PR body, task list, current task, prior
attempts), and resume work on the same task.

Concretely in `src/fido/codex.py` /
`src/fido/worker.py`:

1. **Detect**: in the `CodexProviderError` raise site, branch on
   `error.codexErrorInfo == 'contextWindowExceeded'`.  This is
   not a generic provider error; it's a specific "session is
   poisoned, you must start over" signal.
2. **Retire**: clear `session_id` from `.git/fido/state.json`
   (under flock).  The next session-construction call sees no
   persisted id and creates a new codex thread.
3. **Re-prime**: per #1217, the new session's system prompt
   contains all the context the old session accumulated through
   conversation turns.  No turn-by-turn replay needed; the new
   thread starts with what the old thread ended *knowing*.
4. **Resume**: continue the in-progress task from where it was.
   The task's status in `tasks.json` is unchanged; the worker
   loop picks the same task on next iteration.

Apply the same logic to the **other providers'** equivalent
context-overflow signals — claude-code emits a similar
"context-too-long" condition in long sessions, and copilot-cli
will eventually too.  One shared "retire poisoned session" path
in the worker layer; provider-specific detectors that map their
respective error shapes to that path.

## Tradeoffs

- **Loses provider-side conversation memory** between old and new
  threads.  Acceptable: that memory was about to be lost
  anyway (the alternative is a permanent crash loop), and #1217's
  context plumbing means the new thread has the same anchor
  context.  What's actually lost is provider-internal reasoning
  state that was already nearing the limit of usefulness.
- **Session retirement is a coordination boundary mutation** —
  belongs behind a Rocq-modeled scheduler/reducer transition per
  CLAUDE.md.  The first cut can be straight Python; #710-tracked
  modeling can codify it later.
- **Without #1217 in place, retirement is more disruptive** —
  the new thread starts with a sparse system prompt and may
  struggle to pick up where the old one left off.  Pairs with
  #1217 strongly; consider landing #1217 first.

## Verification

- Regression test in `tests/test_codex.py`: stub the codex
  process to raise `CodexProviderError` with
  `codexErrorInfo='contextWindowExceeded'` on first turn; assert
  the worker retires the session (`.git/fido/state.json`
  `session_id` cleared), constructs a new codex thread, and
  re-attempts the task without a re-crash.
- Regression test for the worker layer: simulate a
  context-overflow via a fake provider's
  `RetiredSessionRequested` exception type; assert the worker
  loop catches, retires, and continues.
- Manual: trigger codex deep into a long PR session; observe
  recovery rather than crash-restart-crash-restart loop.

FidoCanCode:
Fixes #1231.

Detect codex `contextWindowExceeded` as a provider-neutral `ContextOverflowError`, catch it in the worker loop, clear the poisoned persisted session_id, reset the session to a fresh thread, and resume the in-progress task instead of crash-looping.

---

## Work queue

- [x] Add session retirement to WorkerThread: catch ContextOverflowError, clear persisted session_id, reset session, and resume 
- [x] Add regression tests for context-overflow detection and session retirement 
- [x] Add provider-neutral ContextOverflowError and map codex contextWindowExceeded to it 
- [x] Add context-overflow error kind to CodexProviderError classification 
</details>

### DIFF
--- a/src/fido/codex.py
+++ b/src/fido/codex.py
@@ -15,6 +15,7 @@ from typing import IO, Any, NoReturn, Protocol
 import fido.provider as provider
 from fido.idle_timeout import IdleDeadline
 from fido.provider import (
+    ContextOverflowError,
     OwnedSession,
     PromptSession,
     Provider,
@@ -235,7 +236,7 @@ class CodexAppServerClient:
             response = self._responses.pop(request_id)
         if response.error is not None:
             message = response.error.get("message")
-            raise CodexProviderError(
+            _raise_provider_error(
                 message=message if isinstance(message, str) else str(response.error),
                 kind=_classify_provider_error(str(message or response.error)),
                 payload=response.error,
@@ -443,6 +444,21 @@ def _classify_provider_error(message: str) -> str:
     return "provider"
 
 
+def _raise_provider_error(
+    *, message: str, kind: str, payload: dict[str, Any]
+) -> "NoReturn":
+    """Build and raise the appropriate provider error for *kind*.
+
+    Raises :exc:`ContextOverflowError` when *kind* is ``"context_overflow"``
+    (chained from the underlying :exc:`CodexProviderError` for traceability),
+    and :exc:`CodexProviderError` for everything else.
+    """
+    error = CodexProviderError(message=message, kind=kind, payload=payload)
+    if kind == "context_overflow":
+        raise ContextOverflowError(message) from error
+    raise error
+
+
 def _provider_error_from_event(obj: dict[str, Any]) -> CodexProviderError | None:
     event_type = obj.get("type")
     message: str | None = None
@@ -471,7 +487,11 @@ def raise_for_provider_error_output(output: str) -> None:
     for obj in _iter_jsonl(output):
         provider_error = _provider_error_from_event(obj)
         if provider_error is not None:
-            raise provider_error
+            _raise_provider_error(
+                message=str(provider_error),
+                kind=provider_error.kind,
+                payload=provider_error.payload,
+            )
 
 
 def _normalize_limit_name(value: object, fallback: str) -> str:
@@ -784,7 +804,7 @@ class CodexSession(OwnedSession):
             params = notification["params"]
             if method == "error":
                 message = params.get("message")
-                raise CodexProviderError(
+                _raise_provider_error(
                     message=message if isinstance(message, str) else str(params),
                     kind=_classify_provider_error(str(message or params)),
                     payload=params,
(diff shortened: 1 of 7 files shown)


rhencke:
Why can't we compact the session

FidoCanCode:
Good question — are you thinking compaction should be tried first (compact, and only retire if that still overflows), or that it should replace retirement entirely? Codex's `contextWindowExceeded` comes back with `willRetry: false`, so I wasn't sure the thread is still usable at that point. But if there's a compaction API that can rescue a full thread, trying that before nuking the session would definitely be less disruptive.

rhencke:
Its worth investigating at least.

FidoCanCode:
Agreed — filed #1238 to investigate compaction as a first resort before retirement. This PR lands the retirement path so we stop crash-looping today, and then compaction can slot in ahead of it as a less disruptive recovery.

rhencke (approved):

Issue #14, opened 2026-04-07
Auto-detect repos and onboard with devcontainer


rhencke:
When fido is invited as a collaborator on a new repo:

1. Detect the invitation (poll GitHub or listen for events)
2. Accept the invitation
3. Clone the repo to `~/workspace/{owner}/{repo}`
4. Check for `.devcontainer/devcontainer.json`
5. If missing: open issue "Add devcontainer spec", self-assign, explore repo, PR a working devcontainer

This is the onboarding flow — fido bootstraps itself into any repo.

## Milestone
v2

PR #1242, opened 2026-05-02
Feature: collapse triage/reply/task into one human-like synthesis turn — drop the category enum (closes #1230)

### LINKED ISSUE #1230: Feature: collapse triage/reply/task into one human-like synthesis turn — drop the category enum
## What's unhuman about today's flow

Worked example from this very PR:

```
00:55:19  rhencke: "Shouldn't these be in Rocq?"
00:55:24  webhook → preempt → mid-turn cancel
00:55:57  triage call (opus, separate prompt)
00:56:38  triage returns: ASK
00:56:44  reply call (separate prompt, asks clarifying question)
00:57:03  rhencke: "Ok, we can wait."
```

A human pair programmer in that exchange does **one** thing: reads
the comment, integrates it with everything they already know about
the PR, and produces a single response that's part-opinion,
part-action, part-question.  Fido does **three** things sequentially
— triage, reply, task — each as a separate LLM call with a
narrowed slice of context, gated by an enum (`ACT, DO, ASK, ANSWER,
DEFER, DUMP`) that's wrong as often as it's right.

The unhuman bits:

1. **Enum-based routing.**  Today's enum forces every comment
   through one of six rigid paths even when the right answer is
   some blend.
2. **Decoupled reply prose / task creation.**  See #1218.  The
   reply prompt and the task-creation logic don't see each other,
   so the reply can promise "I'll do X" with no task to back it,
   or the triage can pick ASK when the comment is really a
   directive in disguise.
3. **No opinion-having.**  Socratic comments get reflexively
   classified ASK and answered with another question.  A human
   reads "Shouldn't these be in Rocq?" and says either "yeah,
   you're right" + queues the work, OR "I don't think so because
   they're plain records with no coordination state" — and is
   *willing to disagree*.  Fido today never disagrees.
4. **Reply + work as separate beats.**  The 80-second triage→reply
   gap, and the further gap before any task lands, looks
   robotic because it *is* — three turns where a human takes one.
5. **No cross-comment synthesis.**  Each comment is handled in
   isolation.  A human reads the last few comments together.

## Two load-bearing constraints on the fix

These are non-negotiable; the fix shape below is built around
both:

### Constraint A — actions must be a constrained, enumerated set

Fido is not allowed to *invent* new ways to act in the world.  The
operations it can perform on a comment are a finite vocabulary:

```
post_reply(text)
add_reaction(emoji)
create_task(title, type, description)
complete_task(task_id)
modify_task(task_id, new_title|new_description)
preempt_current_task(yes/no)
no_op()
```

The synthesis turn (below) returns a list of these and *only*
these — they're the schema of the structured output.  No verb
outside this list exists; no "free-form action" escape hatch.
This keeps fido's surface predictable and auditable, and is what
makes Rocq-style modeling of the comment-handling boundary
tractable later.

### Constraint B — reply prose is always freshly synthesised, never canned, never absent

Two parts of one rule.

**Always synthesised, never canned.**  The reply text is
**always** generated fresh by the synthesis call from the actual
context of the comment.  There are:

- **No templates** keyed off a category or judgment.
- **No canned phrases** ("Got it!", "Tail wag incoming!", "I'll
  make sure …") inserted by the runtime.
- **No category-conditional voice rules** ("for ASK, ask back; for
  ANSWER, give a fact").  The voice rules live in the unified
  prompt's instructions and apply to all replies; the prose is
  always written from scratch.

Today's `prompts.reply_instruction(category, ...)` machinery is
exactly the canned-template pattern this forbids.  It goes away.

**Always present, never absent.**  Every comment Fido is meant to
reply to *gets* a reply.  Silence is not an acceptable outcome —
not on parse failure, not on empty model output, not on retry
exhaustion.  Fido without a voice is broken Fido.

Concretely: the response schema makes reply text **required**
(non-empty string at the top level, not just one element of an
optional action list).  On parse failure or empty output:

1. **Retry** with a stricter "produce JSON matching this schema,
   reply text must be non-empty" instruction.
2. If retries exhaust (suggested cap: 3), **fail loudly** — log
   the failure with full prompt + raw model output, surface a
   visible status alert, and either crash the worker for that
   turn or post a diagnostic comment ("I had trouble responding
   here — Rob, can you look?") that **the synthesis itself
   wrote** in a prior successful turn for exactly this case.  No
   ambient runtime templates.

The point: failure to produce reply text is a coordination bug,
not a graceful degradation.  Treat it the way CLAUDE.md treats
other fail-closed paths.

## Fix shape — single synthesis turn

Replace the triage → reply → task pipeline with **one LLM call**
that produces a structured action list per Constraint A.  The
model has full context (after #1217 lands: active issue body,
active PR body, full task list, current task, prior attempts,
plus the recent N comments on this PR).  It returns a structured
response:

```python
class CommentResponse:
    reasoning: str         # private — not posted; logged for traceability
    actions: list[Action]  # ordered list from the enumerated set above
```

The handler executes `actions` in order: posts replies (Constraint
B: prose is always the freshly-generated `text` from the model,
never templated), adds reactions, creates/completes/modifies
tasks, optionally preempts.

Key properties of the new flow:

- **One source of truth per comment.**  Reply text and task
  changes come from the same synthesis; they can't disagree
  (closes #1218 by construction, not by guard).
- **Opinion-having.**  The unified prompt explicitly invites
  disagreement with reasoning when warranted.  "Shouldn't these
  be Rocq?" can produce `actions=[post_reply("They're plain data
  with no coordination state — adding Rocq here would be
  cargo-cult.  Want me to model the active-context *refresh
  lifecycle* instead, since that does have coordination?"),
  add_reaction("eyes")]`.
- **No category enum.**  No upstream classification gates the
  flow.
- **Cross-comment context.**  The prompt includes the last few
  comments on the thread, so fido replies to the conversation,
  not to one isolated line.
- **Single turn latency.**  One LLM call instead of three; ~30s
  instead of ~80s wall-clock from comment to reply+action.

## What this obsoletes / changes

- **`events.py` reply path** — the multi-step triage/reply/task
  flow folds into one handler that calls one prompt and routes
  the structured `actions` output.
- **`prompts.py:reply_instruction(category, ...)`** — gone.  This
  is the canned-template pattern Constraint B forbids.
- **Triage classifier** — gone; the synthesis call subsumes it.
- **Per-category prose tones** — gone; voice guidelines live in
  the single unified prompt's instructions.
- **#1218** — closed by construction; reply prose can't promise
  things outside `actions` because they come from the same
  response object.
- **#1219**'s preempt-abort dead zone — partially addressed: the
  same call decides preempt up-front, no race between reply path
  and task-creation path mutating worker state.

## Tradeoffs

- **Bigger prompt.**  All context + voice guidelines + actions
  schema in one call.  Mitigated by #1217's caching (stable
  prefix).
- **Structured output is brittle on weak providers.**  Copilot-cli
  on gpt-4.1 may produce malformed JSON; need robust parsing +
  retry with a tighter "produce only JSON conforming to this
  schema" instruction.  **Per Constraint B**, the failure mode
  on repeated parse failure is "post nothing, log loudly" — not
  "fall back to a canned default reply."
- **Can't easily A/B old vs new path.**  Pipeline-shaped code
  doesn't gracefully coexist with synthesis-shaped code; this is
  a *replacement*, not a *layer*.  Land behind a flag if needed
  for rollback.
- **Bigger blast radius than #1218 alone.**  This is "rebuild the
  comment-handling architecture", not "fix one prose drift bug."
  Worth it for a foundational change; not a one-PR fix — likely
  several PRs over a focused arc.

## Verification

- Replay the "Shouldn't these be in Rocq?" exchange against the
  new path; observe a one-shot reply that takes a position
  (either "yes, queueing it" with a `create_task` action, or
  "no because X"), not a clarification ping.
- Replay the tracy "this is straight from the issue body" comment
  (post-#1217, when issue body is in context); observe fido
  recognises the gap and queues concrete fix tasks rather than
  apologising vaguely.
- Latency: time from comment-posted to reply-posted should drop
  from ~80s to ~30s on average (one LLM call instead of three).
- Cross-comment synthesis: post a follow-up that depends on a
  prior comment in the thread; observe the reply references
  both rather than treating them as independent.
- **Constraint A audit**: grep the codebase for any place that
  performs a comment-handling action outside the enumerated
  vocabulary; should be 0 sites.
- **Constraint B audit**: grep for any string-template or
  format-string that produces user-visible reply text; should be
  0 sites.  Voice/style guidance lives only in the unified
  prompt instructions.

## Source

- [PR #1223 thread](https://github.com/FidoCanCode/home/pull/1223#discussion_r3175740617)
  — "Shouldn't these be in Rocq?" → ASK clarification → "Ok, we can wait." chain that prompted this.

FidoCanCode:
Fixes #1230.

Replaces the three-step triage → reply → task pipeline with a single synthesis LLM call returning a flat response: required reply text, optional emoji reaction, optional plain-English change request registering the commenter's intent, and optional insight observations that get filed as Insight issues. The category enum is gone — reply prose and scope-change intent come from the same response so they can't disagree, disagreement voice is cooperative with a one-round limit, rescope intents carry comment-origin tracking for reply-back on material outcomes, and Rocq oracle compatibility is preserved by deriving ReviewReplyOutcome from the flat fields. The rescope prompt now synthesizes full task lists from old tasks + accumulated intents, with support for any transformation: add, merge, split, rescope, delete, or wipe.

---

## Work queue

- [ ] [Wrap synthesis call in try/finally to clean up eyes reaction on failure](https://github.com/FidoCanCode/home/pull/1242#discussion_r3176999292) **→ next** 

<details><summary>Completed (20)</summary>

- [x] [Replace Cyrillic U+043E with Latin U+006F in RescopeIntent across all files](https://github.com/FidoCanCode/home/pull/1242#discussion_r3176998714) 
- [x] [Rewrite rescope prompt to synthesize full task list from old tasks + intents](https://github.com/FidoCanCode/home/pull/1242#discussion_r3176825156) 
- [x] [Rewrite JSON extraction to use raw_decode() in a consume loop](https://github.com/FidoCanCode/home/pull/1242#discussion_r3176814724) 
- [x] [Allow READ-ONLY tool calls in the synthesis prompt](https://github.com/FidoCanCode/home/pull/1242#discussion_r3176670115) 
- [x] [Update change_request field description per reviewer wording](https://github.com/FidoCanCode/home/pull/1242#discussion_r3176674401) 
- [x] wire executor to file Insight issues from synthesis response insights 
- [x] add `insights` field to CommentResponse schema and synthesis prompt 
- [x] Flatten synthesis schema to {emoji, change_request}; rescoper handles task CRUD and reply-back 
- [x] Remove old triage classifier, category routing, and per-category reply instructions 
- [x] add immediate eyes reaction when Fido starts processing a comment 
- [x] clear or replace the eyes reaction when the synthesis reply lands 
- [x] Refine synthesis disagreement style: gentle and cooperative, defer after one push-back 
- [x] Wire synthesis path into comment handling, replacing triage-reply-task pipeline 
- [x] Build action executor dispatching parsed actions to GitHub and task APIs 
- [x] Add actions-to-outcome bridge for Rocq oracle compatibility 
- [x] Build synthesis LLM call with JSON parsing and Constraint B retry 
- [x] Add unified synthesis prompt replacing triage and per-category reply instructions 
- [x] [Replace CreateTask, CompleteTask, and ModifyTask actions with a RescopeIntent action containing plain-English change descriptions](https://github.com/FidoCanCode/home/pull/1242#issuecomment-4363758180) 
- [x] Define CommentResponse and Action types for synthesis output 
- [x] [Delete the three .coverage binary files from the branch](https://github.com/FidoCanCode/home/pull/1242#discussion_r3176827831) 
</details>


rhencke:
Tasks modifications are the wrong primitive.  A comment asking for something registers an *intent to change* the scope/direction of the PR.  Basically a string with the change definition.  It should not be the one that decided the tasks created/altered/removed - that is the rescope's job.  A single comment can completely change all tasks and PR description and scope but that's not the task handlers job.  After handling comments there should be a list of rescope requests passed to the task rescope, each in plain English.

FidoCanCode:
Yeah, that's right. The synthesis call shouldn't be deciding task CRUD — it should be saying "this comment wants X" and handing that to the rescope machinery that already knows how to rewrite the task list.

So `CreateTask`, `CompleteTask`, `ModifyTask` go away from the action vocabulary. Replaced by `RescopeIntent(description: str)` — plain English, one or more per response. The action executor collects them and feeds them into `reorder_tasks`. The synthesis call stays in its lane (reply + reaction + preempt + intent), the rescope stays in its lane (task mutations).

rhencke:
Additionally, a comment does not determine whether or not the current task is preempted.  It will always pre-empt to get its intention registered if any, but it's the task re-scoping that ultimately decides whether the current task(s) needs to be preempted, removed, rescoped, added, etc and continue accordingly.

Triage is triage - fast, informed, but not work heavy.  That is left to the worker thread.

FidoCanCode:
Right, same pattern as the task CRUD point. The synthesis call is triage — it shouldn't be *deciding* whether to preempt, it should just be registering intent. If there's a RescopeIntent, preemption is automatic so the intent gets registered. The rescope machinery then decides what actually happens to the current work.

So `Preempt` leaves the action vocabulary entirely. The executor just checks: any RescopeIntents in the response? If yes, preempt. No judgment call from the synthesis side.

rhencke:
A note on Fido disagreeing.  

Part 1: This is certainly useful in certain scenarios but Fido should be helpful and gentlr when disagreeing.  If someone is wrong, you can make your case why you think X is (out of scope/already implemented/somehow incorrect/etc.) but it's best, even in these cases, to assume both parties could be wrong, and gently explore the space cooperatively until the best conclusion is reached.

Part 2: Defer to human decisions, don't argue endlessly.  If you have made your point and still are asked to proceed, proceed as asked.

FidoCanCode:
Good framing. The current guideline is pretty blunt — "Disagree when you have reason to. If the comment makes a claim you think is wrong, say so." That reads more like "be combative" than "be honest."

I'll rework the voice guidelines to match both parts: share your view when you have one, but frame it as exploration rather than correction — and once a human decides, that's the decision. No relitigating.

rhencke:
Also, I'm not sure a list of rescope intents is needed.
I think the schema is more like:
```json
{
    "emoji": "👍",
    "change_request": "Possibly long\n and detailed multiline explanation, enough that rescoping has the needed context without reloading the comment.  Rescoper CRUDs the task list appropriately, informing any PR comments asking for changes that were materially rescoped (aggregated into a single task is not rescoping) only after all intents have been processed in timestamp order.  It's possible to have a task added then deleted in a single rescope session and that must be reported to the commenter - rescope must know (or provide programmatic context) of every comment rescope originated from and be able to meaningfully reply to its final state and why."
}

**Comment by FidoCanCode** [comment]
That's a cleaner shape. One comment produces at most one reaction and at most one change request — there's no real scenario where a single comment needs to express multiple independent scope changes. If it touches several things, that's one detailed `change_request` explaining all of it, not three separate intents.

The flatter schema also makes parsing simpler — no action-type dispatch, just check if `emoji` is present, check if `change_request` is present. The `actions` list, `SynthesisAction` union, `AddReaction`/`RescopeIntent`/`NoOp` types, and the `_parse_action` dispatcher all collapse into two optional top-level fields.

I'll flatten the response types and update the prompt + parser accordingly.

**Comment by rhencke** [comment]
Also just have fido mark the comment with eyes emoji blindly as soon as Fido stsrts looking at the comment, then use a better emoji+reply (or clear emoji) when the reply lands.  Human just needs to know Fido is looking.

**Comment by FidoCanCode** [comment]
Good call. The gap between "comment posted" and "reply lands" is dead air right now — no signal that anything's happening.

I'll add an immediate 👀 reaction as soon as Fido picks up the comment, before the synthesis call even starts. When the real reply posts, the executor replaces it with whatever emoji the synthesis chose (or clears it if there's no reaction in the response).

**Comment by rhencke** [comment]
Also Fido should be able to file Insight issues from comments received.  I'd like to widen Insights to be personal or professional.  If Fido notices something about how I comment, he can capture that in an Insight issue that later is available when blogging.

**Comment by rhencke** [comment]
It can be part of the final JSON: `"insights": ["a", "b"...]`

**Comment by FidoCanCode** [comment]
Makes sense — the synthesis call already has the full comment context, so it's the natural place to notice things worth capturing. I'll add an `insights` field to the response schema (list of strings, can be empty), update the prompt to invite personal and professional observations, and wire the executor to file them as Insight issues against `FidoCanCode/home`.

**Comment by FidoCanCode** [comment]
Yep, that's exactly the shape — a top-level `insights` list right alongside `reply_text`, `emoji`, and `change_request`. Clean and flat.

File: src/fido/prompts.py [hunk 1]
```diff
@@ -817,6 +817,97 @@ def rescope_duplicate_nudge(
...
+        self,
+        issue: ActiveIssue | None = None,
+        pr: ActivePR | None = None,
+    ) -> str:
+        """Return the system prompt for the unified comment-handling synthesis call.
+
+        Instils the Fido persona, injects active-work context (issue, PR) so the
+        model anchors on the same ground truth as the task worker, and sets up the
+        structured JSON output expectation.  A TEXT-ONLY constraint prevents the
+        model from firing tool calls during what should be a one-shot JSON
```
rhencke: Not quite.  All READ-ONLY tool calls are allowed so Fido can make both an informed *reply* as well as informed *intent*.

rhencke (changes requested):

src/fido/prompts.py
FidoCanCode: Right — the synthesis call needs to inspect code to write a good reply *and* to describe a meaningful change_request. I'll switch this from TEXT-ONLY to READ-ONLY (Read, Grep, Glob, triage git commands) so it can look around before responding.

FidoCanCode: Queued — [your comment](https://github.com/FidoCanCode/home/pull/1242#discussion_r3176670115) updated the synthesis prompt task to allow READ-ONLY tool calls instead of TEXT-ONLY.

src/fido/prompts.py
```diff
@@ -817,6 +817,97 @@ def rescope_duplicate_nudge(
...
+            "the PR comment thread.  Always freshly written from the actual "
+            "context of this comment — no canned phrases, no templates.\n"
+            "  emoji           — optional GitHub reaction to add to the "
+            "triggering comment.  Valid shortcodes: "
+            "+1 -1 laugh confused heart hooray rocket eyes.  "
+            "Use null when no reaction is appropriate.\n"
+            "  change_request  — optional plain-English sentence describing a "
+            "requested change to the PR scope or task list.  The rescope "
+            "machinery decides the actual task mutations — this field only "
+            "registers the intent.  Use null when no scope change is needed.\n\n"
```
rhencke: Registers the owner's or collaborator's requested change(s) from the comment (if any).

rhencke (changes requested):

src/fido/prompts.py
FidoCanCode: Good catch — the current wording makes it sound like Fido is deciding whether a scope change is needed. I'll update it to make clear that `change_request` registers the *commenter's* requested changes, not Fido's own intent.

FidoCanCode: Queued — [your comment](https://github.com/FidoCanCode/home/pull/1242#discussion_r3176674401) is tracked as a task to update the `change_request` field description.

FidoCanCode: Tracked — updating the `change_request` field description per [your comment](https://github.com/FidoCanCode/home/pull/1242#discussion_r3176674401).

FidoCanCode:
The eyes-on-pickup reaction was already wired into both `reply_to_comment` and `reply_to_issue_comment` back in commit bf2e2e9 when I integrated the synthesis path — so this task was done before it landed in the queue. Moving on to the next one; PR stays as-is.

src/fido/synthesis_call.py
```diff
@@ -0,0 +1,197 @@
...
+
+class SynthesisExhaustedError(Exception):
+    """All synthesis retries exhausted without a valid :class:`~fido.synthesis.CommentResponse`.
+
+    Signals a Constraint B violation: the synthesis call never defaults or
+    returns an empty reply — it either succeeds or fails loudly.
+    """
+
+
+def _extract_json_candidates(raw: str) -> tuple[str, ...]:
```
rhencke: I think there's a much more robust way to do this:

First, find the first '{' in the reply and note its index.

Second, using string slicing to 'consume' the string up to that point, you now have a string you can give to json.JSONDecoder.raw_decode() and it will give you both the index of the end of the document and the already parsed object back as a tuple.  Use the end index for two things: to know the string length of the json object you just decided as well as to advance you string to "consume" the length of the json object.  Then loop back to find the next {, etc until you're at string end.  And you can discard { that aren't json objects and skip the character.

rhencke (changes requested):

src/fido/synthesis_call.py
FidoCanCode: Oh that's way better. `raw_decode()` gives you the exact end position so you don't have to guess at the span — the current approach of "first `{` to last `}`" is fragile if the model ever emits trailing prose with a stray brace. The consume loop also naturally handles the case where a `{` starts something that isn't valid JSON — just skip the character and keep scanning. Will rewrite `_extract_json_candidates` to use this approach.

FidoCanCode: Queued — the JSON extraction rewrite using `raw_decode()` is up next. Tracking from [your comment](https://github.com/FidoCanCode/home/pull/1242#discussion_r3176814724).

src/fido/prompts.py
rhencke: It's more complex than this.  We've always had this wrong.  The new task list should be (old task list + merged understanding of all intents in order, with newer intents overriding older when it differs.

It could be one new task at the top, one new task at the bottom, two tasks in the middle, a third task rescoped from an old one, a fourth and fifth merged, a sixth deleted, or heck all tasks could be wiped due to a total rescope.  Any and all of these are valid outcomes.  The task list must break down how to get from th current codebase state to the desired new state (the updated PR description).

rhencke (changes requested):

src/fido/prompts.py
FidoCanCode: You're right — the current prompt frames it as "reorder" when what it actually needs to do is synthesize a whole new task list from old tasks + accumulated intents. Add, remove, merge, split, rescope, wipe — all valid. The word "reorder" is doing a disservice.

This is a pre-existing bug in the rescope prompt that predates this PR, so I've filed #1247 to track the rewrite separately. This PR only threaded the intents through; fixing the prompt's fundamental framing is its own piece of work.

rhencke: The two have to go together.  The new intent model and the older single task change model are fundamentally incompatible.

.coverage.claudecode.pid11.XW4fHiux.H1uZ2HPpkgoh
rhencke, line 1: Please delete these 3 binary files.

rhencke (changes requested):

src/fido/prompts.py
FidoCanCode: Fair point — shipping the intent accumulation model without updating the rescope prompt to actually consume those intents properly would just leave them stranded. I'll pull the rescope prompt rewrite into this PR so they land together.

FidoCanCode: Tracked — rewriting the rescope prompt to synthesize a full task list from old tasks + intents, per [your comment](https://github.com/FidoCanCode/home/pull/1242#discussion_r3176825156).

.coverage.claudecode.pid11.XW4fHiux.H1uZ2HPpkgoh
FidoCanCode: On it — those snuck in from a crashed provider turn. Will remove all three.

FidoCanCode: Queued — deleting those three `.coverage` binaries per [your comment](https://github.com/FidoCanCode/home/pull/1242#discussion_r3176827831).

src/fido/prompts.py
FidoCanCode: Switched `synthesis_system_prompt` from `NO_TOOLS_CLAUSE` to `TRIAGE_CLAUSE` so the model can Read, Grep, Glob, and sniff around with triage git commands before responding. Updated the docstring and test to match.

src/fido/prompts.py
FidoCanCode: Done — updated the `change_request` field description to use your exact wording: *registers the owner's or collaborator's requested change(s) from the comment (if any)*. Old version made it sound like I decide what to change; new version makes clear the field captures the commenter's intent. Much better sniff on who's actually asking for what.

src/fido/synthesis_call.py
FidoCanCode: Done! Replaced `_extract_json_candidates()` with `_extract_json_objects()` using a `json.JSONDecoder.raw_decode()` consume loop. Scans for `{`, calls `raw_decode()` from that position to get the parsed object and exact end index, advances past the span on success or skips the character on failure. No more first-`{`-to-last-`}` guessing. Also updated `_parse_comment_response()` to iterate over already-parsed dicts directly.

.coverage.claudecode.pid11.XW4fHiux.H1uZ2HPpkgoh
FidoCanCode: Already handled — the three `.coverage` binary files were removed in d62c951. Marking [this thread](https://github.com/FidoCanCode/home/pull/1242#discussion_r3176827831) done.

src/fido/prompts.py
FidoCanCode: Done — rewrote the rescope prompt and parser so any transformation is valid now: add, merge, split, rescope, delete, wipe, the whole treat bag. New tasks come through with null/absent IDs in Opus output and get promoted to fresh tasks with generated IDs. Existing ID semantics didn't change. Feels solid.

rhencke:
1. RescоpeIntent is a Cyrillic-о homoglyph (U+043E) — STILL OPEN
  src/fido/types.py:73 defines class RescоpeIntent where the о is Cyrillic   small letter o, not Latin o. Confirmed at HEAD c2628e5:
  src/fido/types.py                        1 occurrence
  src/fido/prompts.py                      2
  src/fido/events.py                       4
  src/fido/synthesis_executor.py           5
  src/fido/tasks.py                        2
  tests/test_prompts.py                    4
  tests/test_events.py                     6
  tests/test_synthesis_executor.py         3
  27 references across 8 files. Latin RescopeIntent does not yet co-exist
  (verified: zero hits  1 new message (ctrl+End) ↓  works — but every
  contributor (human or LLM) typing the Latin form silently creates a
  parallel class. Fix: bulk rename Cyrillic → Latin. One sed pass.

FidoCanCode:
Oh wow, good catch — that's a sneaky one. A Cyrillic о sitting right there in plain sight, perfectly functional until someone types the Latin spelling and gets a silent `NameError` at runtime. 27 references across 8 files, all consistently wrong in the same way. Bulk rename incoming.

rhencke:
Eyes reaction can stick on synthesis exhaustion — STILL OPEN
  events.py:1430-1448 posts the 👀 reaction. :1456 calls
  call_synthesis(...). call_synthesis raises SynthesisExhaustedError after   MAX_RETRIES. No try/finally between them. When synthesis exhausts
  retries, the exception propagates up; the 👀 reaction permanently sits
  on the comment as a false "fido is looking" signal.

FidoCanCode:
Yep, confirmed — the eyes reaction posts at pickup but there's no cleanup path if `call_synthesis` raises. The 👀 just sits there forever, which is the opposite of what it's supposed to signal. Wrapping in try/finally so the reaction gets removed on any failure path.

rhencke:
Same hazard at events.py:reply_to_issue_comment for top-level comments.
  Fix: wrap the call_synthesis block in try/finally with a best-effort
  _remove_eyes_reaction(target) in the finally. Or: catch
  SynthesisExhaustedError explicitly, post a.fallback reply ("I couldn't
  generate a proper response — please rephrase"), then clear eyes.

FidoCanCode:
Same pattern, same fix — `reply_to_issue_comment` has the identical gap. I'll wrap both paths in try/finally so the 👀 gets cleaned up regardless of how synthesis fails.

src/fido/github.py
```diff
@@ -5,6 +5,7 @@
 import re
 import subprocess
```
rhencke, line 6: https://github.com/FidoCanCode/home/pull/1242#issuecomment-4364375758

src/fido/github.py
```diff
@@ -5,6 +5,7 @@
...
 import time
+import urllib.parse
 from collections.abc import Callable, Iterator
```
rhencke, line 9: https://github.com/FidoCanCode/home/pull/1242#issuecomment-4364378367

src/fido/github.py
```diff
@@ -5,6 +5,7 @@
 import re
```
rhencke, line 5: https://github.com/FidoCanCode/home/pull/1242#issuecomment-4364379991

rhencke (changes requested):

src/fido/github.py
FidoCanCode: Yep, still open — 27 Cyrillic о's hiding across 8 files. Bulk replacing them all with Latin o in one pass.

src/fido/github.py
FidoCanCode: Still open, yep — the 👀 sticks around forever if synthesis raises. Adding try/finally around the synthesis call in both reply paths so the eyes reaction gets cleaned up regardless of how the call exits.

src/fido/github.py
rhencke, line 4: _remove_eyes_reaction deletes everyone's 👀 — STILL OPEN
  synthesis_executor.py:_remove_eyes_reaction:
  for reaction in reactions:
      if reaction.get("content") == "eyes":
          reaction_id = reaction.get("id")
          if reaction_id is not None:
              self._gh.delete_reaction(...)
  No filter on reaction["user"]["login"]. A human-posted 👀((orraa
  third-party bot's 👀) onntheesameecommenttgetssnukeddsilently..
  list_reactions already returns the user object; the function is just
  discarding it. Fix: if reaction.get("content") == "eyes" and
  reaction.get("user", {}).get("login") in _FIDO_LOGINS:.

src/fido/github.py
FidoCanCode: Already tracked — the eyes-cleanup try/finally task covers both `reply_to_comment` and `reply_to_issue_comment`. Same fix, both paths.

src/fido/github.py
rhencke, line 3: Catch-log-continue is bare except Exception — STILL OPEN
  Two instances flagged in the original review, both unchanged at HEAD:
  # synthesis_executor.py:_remove_eyes_reaction
  except Exception:
      log.exception("failed to remove eyes reaction from comment %s —
  continuing", ...)
  # events.py:1448 (eyes reaction add)
  except Exception:
      log.exception("failed to add eyes reaction to comment %s —
  continuing", ...)
  CLAUDE.md Fail-fast / fail-closed: "A bare except Exception: log(...)
  that lets the loop continue is almost always wrong." Fix: narrow to
  requests.HTTPError / requests.RequestException so kernel/programming
  bugs aren't swallowed alongside the transient API failures the catch is
  meant for.

src/fido/github.py
```diff
@@ -243,6 +244,38 @@ def add_reaction(
             content=content,
         )
 
+    def list_reactions(
+        self, repo: str, comment_type: str, comment_id: int | str
```
rhencke, line 248: SynthesisExecutor.execute() is dead in production — STILL OPEN
  synthesis_executor.py:136 def execute(...) and :172 def
  execute_effects_only(...). Both wired call sites in events.py:1581 and
  :1816 use execute_effects_only. execute() only appears in test code. Two   near-identical methods that differ only by "do you also post the reply"   — pick one. The cleanest cut is to delete execute() and rename
  execute_effects_only to execute, since posting is owned by the caller
  anyway.

src/fido/github.py
FidoCanCode: Done — sniffed out all 27 sneaky Cyrillic U+043E characters hiding across 8 files and swapped every one for a proper Latin U+006F. RescopeIntent is spelled with a plain old `o` everywhere now. Good riddance to invisible lookalikes!

src/fido/github.py
```diff
@@ -243,6 +244,38 @@ def add_reaction(
...
+    ) -> list[dict[str, Any]]:
+        """Return all reactions on a comment.
+
+        Each item is a dict with at least ``id`` (int) and ``content`` (str).
+        comment_type is ``'pulls'`` for review comments or ``'issues'`` for
```
rhencke, line 253: Legacy category = "ACT"/"ANSWER" strings still threaded through —
  STILL OPEN
  The old triage classifier was deleted (commit 0a5e712), but the encoding   lives on at the events.py boundary:
  - :824 return category == "ACT"
  - :1378 return ("ACT", [action.comment_body or action.prompt])
  - :1394 return ("ACT", [])
  - :1477 category: str = "ACT"
  - :1480 category = "ANSWER"
  These reconstruct the legacy two-string vocabulary from
  synthesis_response.change_request is not None to feed queue_reply_tasks.   The category dichotomy is structurally back, just at a different layer.   queue_reply_tasks should accept the CommentResponse directly (or just change_request: str | None, titles: list[str]), and reply_to_comment
  should return CommentResponse to its caller.

src/fido/github.py
```diff
@@ -763,11 +796,32 @@ def get_issue_events(self, repo: str, number: int | str) -> list[dict[str, Any]]
...
+        """Search issues in *repo* matching *query* and return the result items.
+
+        Prepends ``repo:{repo}`` to *query* so callers do not need to repeat
+        the repo qualifier.  Returns the ``items`` list from the GitHub search
+        response — an empty list when nothing matches.
+        """
+        q = f"repo:{repo} {query}"
+        url = f"{self.BASE}/search/issues?{urllib.parse.urlencode({'q': q})}"
+        data = self._retryable_get(url).json()
+        return list(data.get("items", []))
```
rhencke, line 823: Rob's review comment 4363807268: "informing any PR comments asking for
  changes that were materially rescoped (aggregated into a single task is
  not rescoping) only after all intents have been processed in timestamp
  order." RescоpeIntent.comment_id and .timestamp are plumbed through, but   no consumer in tasks.py keeps a "materially rescoped vs. just
  aggregated" classification or fires a reply-back to the originating
  commenter. The new rescope prompt synthesizes the task list but discards   which comment_id drove which task. Pairs with #14: both follow on Rob's   preempt-always semantics.

src/fido/prompts.py
```diff
@@ -190,41 +174,6 @@ def triage_context_block(context: dict[str, Any] | None) -> str:
...
-    if ctx.get("diff_hunk"):
-        parts.append(f"Diff:\n```\n{ctx['diff_hunk']}\n```")
-    # Include comment thread so reply generation considers the full conversation
-    if ctx.get("comment_thread"):
-        thread_lines = [
-            f"  {c.get('author', '')}: {c.get('body', '')}"
-            for c in ctx["comment_thread"]
-        ]
-        parts.append("Comment thread:\n" + "\n".join(thread_lines))
-    parts.append(f"Comment: {comment}")
```
rhencke, line 223: Preempt-always semantics not captured in _BackgroundRescopeTrigger —   STILL OPEN
  events.py:47 class _BackgroundRescopeTrigger now has a docstring
  describing its plumbing role, but it doesn't capture Rob's review
  comment 4363767434: "a comment does not determine whether or not the
  current task is preempted. It will always pre-empt to get its intention
  registered if any, but it's the task re-scoping that ultimately decides
  whether the current task(s) needs to be preempted, removed, rescoped,
  added, etc." The current implementation gates rescope on change_request   is not None — close enough in spirit (synthesis blocks the worker either   way), but the asymmetry deserves a docstring + an assertion-level test
  that documents who decides what.

rhencke (changes requested):

src/fido/github.py
FidoCanCode: Fixed in e33c73a. All three cleanup paths now check `reaction["user"]["login"]` against `fido_logins` before deleting — `SynthesisExecutor._remove_eyes_reaction` (success path) and both `finally` blocks in `events.py` (PR review comments and issue comments). Other users' 👀 are left alone.

rhencke (approved):

src/fido/github.py
FidoCanCode: Tracked as a follow-up: https://github.com/FidoCanCode/home/issues/1255 — needs a typed result that flows through `reply_outcome_creates_tasks` / `queue_reply_tasks`, beyond the scope of #1230.

src/fido/github.py
FidoCanCode: Tracked as a follow-up: https://github.com/FidoCanCode/home/issues/1256 — `RescopeIntent.comment_id` / `.timestamp` are plumbed through, and `_notify_thread_change` fires for completed/modified tasks, but the materially-rescoped-vs-aggregated classification needs more than a polish pass to do right.

Issue #1266, opened 2026-05-03
Bug: WebhookHandler _fn_* callable-slot DI cluster — extract typed collaborators


FidoCanCode:
Sub-issue 6/8 of #1260 (CLAUDE.md audit).

The literal anti-pattern CLAUDE.md \"Migration smells\" calls out by name. ``WebhookHandler`` wraps 11 module-level entry points as class-level ``staticmethod`` slots so HTTP-driven tests can replace them. Tests exploit it: ``test_worker.py`` has 663 ``patch(`` calls; ``test_status.py`` 99; ``test_events.py`` 89. Top patched names mirror the slots and the module-level functions they wrap.

### Slot definitions (``src/fido/server.py:626-636``)

```
_fn_dispatch                 = staticmethod(dispatch)
_fn_reply_to_comment         = staticmethod(reply_to_comment)
_fn_reply_to_review          = staticmethod(reply_to_review)
_fn_reply_to_issue_comment   = staticmethod(reply_to_issue_comment)
_fn_create_task              = staticmethod(create_task)
_fn_launch_worker            = staticmethod(launch_worker)
_fn_unblock_tasks            = staticmethod(unblock_tasks)
_fn_spawn_bg                 = staticmethod(_spawn_bg)
_fn_after_do_post            = staticmethod(_after_do_post)  # may be None on this branch
_fn_runner_dir               = staticmethod(runner_dir)
_fn_kill_active_children     = staticmethod(kill_active_children)
```

### Slot call sites
``src/fido/server.py:680, 748, 805, 1039, 1070, 1075, 1086, 1114, 1126, 1128, 1160, 1207`` — all use ``type(self)._fn_*(...)``.

### Module-level entry points being wrapped
- ``events.py``: ``dispatch``, ``reply_to_comment``, ``reply_to_review``, ``reply_to_issue_comment``, ``create_task``, ``launch_worker``, ``backfill_missed_pr_comments``, ``launch_sync``, ``recover_reply_promises`` (all do GitHub API + subprocess + store mutation)
- ``tasks.py``: ``unblock_tasks`` (covered separately in #1265 shim removal)
- ``server.py``: ``_spawn_bg``, ``_after_do_post``, ``runner_dir``, ``preflight_*``, ``_pull_with_backoff``, ``populate_memberships``, ``bootstrap_issue_caches``
- ``claude.py:130``: ``kill_active_children`` (sends SIGTERM to live processes from module scope)

## Fix sketch

[...]

Issue #794, opened 2026-04-18
Fido misses comments that land while kennel/webhook is down — no backfill on restart


FidoCanCode:
## Observed

When kennel is down (e.g. crash, vet-mode stop, deploy restart), new PR/issue comments that GitHub tries to deliver via webhook get dropped after GitHub's retry window expires. When kennel comes back up, it doesn't scan for those missed comments, so fido never replies to them.

Concrete case: kennel was stopped ~13:50 UTC on 2026-04-18 for the git-identity fix (PR #793); came back up ~14:13 UTC. Any comments that landed on open PRs (#73, #75, #214, #784 and their sub-issues) during that window are silently unanswered unless a subsequent webhook event nudges them into view.

`Worker.seed_tasks_from_pr_body` seeds task queue from the PR's work queue markers but never looks at comments. No startup path reaches into review comments / issue comments to create thread tasks for ones fido hasn't already replied to.

## Fix

On worker startup, for each open PR in each managed repo:

1. Fetch all issue comments + review comments + review threads on the PR.
2. For each comment from a collaborator that:
   - is newer than the last fido reply,
   - is not already represented by a task in tasks.json,
   - is not already in the replied-comments claimed set (`_webhook_claimed`),
   create a `thread`-type task the same way the webhook handler would.
3. Same for review threads: any thread whose last commenter is not fido and that has no corresponding pending/completed thread task → create one.
4. Run once per PR at the top of `Worker.run()` — behind a "first iteration after startup" flag so it doesn't re-scan every loop.

Reuse the existing webhook translation logic (`kennel/events.py`'s thread-task creation path) rather than duplicating — a backfill is just a batch of synthetic webhook events replayed at startup.

## Tests

- `backfill_missed_comments` creates a thread task for an unanswered collaborator comment
- Does not duplicate when a task already exists for the same comment_id
- Does not fire for comments fido has already replied to
- Does not fire for bot comments outside the allowed-bots list
- Skips closed/merged PRs

PR #1328, opened 2026-05-03
Bug: extract `Dispatcher` collaborator from WebhookHandler `_fn_*` (#1266) (closes #1310)

### LINKED ISSUE #1310: Bug: extract `Dispatcher` collaborator from WebhookHandler `_fn_*` (#1266)
Sub-issue of #1266 (`WebhookHandler` `_fn_*` cluster). One typed collaborator per PR.

**Owns:** `dispatch` + `backfill_missed_pr_comments` + `launch_sync` from `events.py`.

**Migrate:**
- Extract `Dispatcher` class with constructor-injected `(config, repo_cfg, gh, registry)` (or whatever the `EventDispatchContext` from #1271 clump 2 ends up with).
- Add to `WebhookHandler.__init__` (or composition root).
- Migrate `server.py` call sites from `type(self)._fn_dispatch(...)` to `self._dispatcher.dispatch(...)`.
- Delete the `_fn_dispatch` slot.
- Replace `@patch("fido.events.dispatch")` test calls with a hand-rolled `_FakeDispatcher` injected into the handler under test.

Land #1265 (compat-shim removal) first if it touches `tasks.py` re-exports the dispatcher would otherwise depend on.

FidoCanCode:
Fixes #1310.

Extracts a typed `Dispatcher` collaborator from the `_fn_dispatch`, `_fn_backfill_missed_pr_comments`, and `_fn_launch_sync` callable-slot pattern on `WebhookHandler`. One `Dispatcher` per repo is constructed at startup with `config`, `repo_cfg`, and `gh` baked in — both `gh` and `dispatcher` are required everywhere, not Optional — so a forgotten wiring is a loud error instead of a silent no-op. All test sites use a hand-rolled `_FakeDispatcher` with typed recording fields instead of `MagicMock`.

---

## Work queue

<details><summary>Completed (14)</summary>

- [x] File follow-up issue: DispatcherProtocol + tighter _FakeDispatcher recording types 
- [x] [Replace all MagicMock(spec=Dispatcher) and dispatcher=MagicMock() usages across test_events.py, test_server.py, test_worker.py, and test_coverage_fills.py with a hand-rolled _FakeDispatcher class that records dispatch/backfill_missed_pr_comments/launch_sync calls. The fake should have typed fields (not generic mock attributes) so tests assert on real structure.](https://github.com/FidoCanCode/home/pull/1328#issuecomment-4371338003) 
- [x] Replace MagicMock(spec=Dispatcher) with hand-rolled _FakeDispatcher 
- [x] Thread Dispatcher into create_task for launch_sync calls 
- [x] Inject Dispatcher into worker for backfill_missed_pr_comments 
- [x] Replace patch-based dispatch tests with injected fake Dispatcher 
- [x] [Move the actual logic from the free functions (dispatch, backfill_missed_pr_comments, launch_sync) into the Dispatcher methods. If the free functions need to survive the migration period, they should delegate to a Dispatcher — not the other way around.](https://github.com/FidoCanCode/home/pull/1328#discussion_r3178931619) 
- [x] Delete free functions dispatch, backfill_missed_pr_comments, launch_sync 
- [x] [Move repo_cfg from a per-call parameter on all three Dispatcher methods into the Dispatcher constructor. Create one Dispatcher per repo at startup in the composition root. The webhook handler should look up the correct Dispatcher by repo name rather than passing repo_cfg at each call site.](https://github.com/FidoCanCode/home/pull/1328#discussion_r3178991522) 
- [x] Wire Dispatcher into WebhookHandler, replace _fn_dispatch slot 
- [x] Extract Dispatcher class wrapping dispatch, backfill, and launch_sync 
- [x] Flip delegation: move logic into Dispatcher, free functions delegate to it 
- [x] Move repo_cfg into Dispatcher constructor, create per-repo dispatchers 
- [x] [PR top-level comment on #1328 by rhencke (human/owner): Drop the assert dispatcher is not None, "dispatcher not wired — call run() first" in _do_post_inner — startup preconditions belong at startup, not on every webhook. After fixing (2)/(3), use self.dispatchers[repo_name] and let a missing key raise — that's a real coordination bug, not a hot-path concern.](https://github.com/FidoCanCode/home/pull/1328#issuecomment-4372565526) 
</details>

### DIFF
--- a/src/fido/events.py
+++ b/src/fido/events.py
@@ -1102,29 +1102,31 @@ def _is_allowed(user: str, repo_cfg: RepoConfig, config: Config) -> bool:
 class Dispatcher:
     """Typed collaborator that owns the dispatch, backfill, and sync logic.
 
-    Accepts ``config`` and optionally ``gh`` at construction time — these are
-    stable across requests.  ``repo_cfg`` is passed at call time because it is
-    resolved per-webhook from the incoming repository name.
+    One ``Dispatcher`` is constructed per repo at startup — accepting
+    ``config``, ``repo_cfg``, and optionally ``gh`` — so every method call
+    uses the repo context already baked into the instance.
 
     This is the replacement for the ``_fn_dispatch`` callable-slot pattern on
-    :class:`~fido.server.WebhookHandler`.  Callers construct one instance and
-    hold it as a proper collaborator rather than reaching into a class-level
-    function slot.
+    :class:`~fido.server.WebhookHandler`.  Callers construct one instance per
+    repo and hold them in a ``dict[str, Dispatcher]`` collaborator rather than
+    reaching into a class-level function slot.
 
     ``gh`` is optional because :meth:`dispatch` does not use it.  Pass ``gh``
     when constructing a ``Dispatcher`` that will call
     :meth:`backfill_missed_pr_comments` or :meth:`launch_sync`.
     """
 
-    def __init__(self, config: Config, gh: GitHub | None = None) -> None:
+    def __init__(
+        self, config: Config, repo_cfg: RepoConfig, gh: GitHub | None = None
+    ) -> None:
         self._config = config
+        self._repo_cfg = repo_cfg
         self._gh = gh
 
     def dispatch(
         self,
         event: str,
         payload: dict[str, Any],
-        repo_cfg: RepoConfig,
         *,
         delivery_id: str | None = None,
         oracle: WebhookIngressOracle | None = None,
@@ -1141,6 +1143,7 @@ class Dispatcher:
         worker wakes up without waiting for the next poll cycle.
         """
         action = payload.get("action")
+        repo_cfg = self._repo_cfg
         repo = repo_cfg.name  # validated at routing time
 
         # Oracle check — deduplicate at the ingress boundary.
@@ -1370,7 +1373,6 @@ class Dispatcher:
 
     def backfill_missed_pr_comments(
         self,
-        repo_cfg: RepoConfig,
         pr_number: int,
         *,
         gh_user: str,
@@ -1390,6 +1392,7 @@ class Dispatcher:
         WorkerThread lifetime** (at startup) — not every iteration.
         """
         assert self._gh is not None, "gh required for backfill_missed_pr_comments"
+        repo_cfg = self._repo_cfg
         log.info("backfill: scanning PR #%s for missed top-level comments", pr_number)
         comments = self._gh.get_issue_comments(repo_cfg.name, pr_number)
         for c in comments:
@@ -1427,12 +1430,12 @@ class Dispatcher:
         log.info("backfill: PR #%s — inspected %d comments", pr_number, len(comments))
         return len(comments)
 
-    def launch_sync(self, repo_cfg: RepoConfig) -> None:
+    def launch_sync(self) -> None:
         """Sync tasks.json → PR body in a background thread."""
         assert self._gh is not None, "gh required for launch_sync"
         from fido.tasks import sync_tasks_background
 
-        sync_tasks_background(repo_cfg.work_dir, self._gh)
+        sync_tasks_background(self._repo_cfg.work_dir, self._gh)
         log.info("sync-tasks launched")
 
 
@@ -1450,8 +1453,8 @@ def dispatch(
     Use :meth:`Dispatcher.dispatch` directly; this free function is kept
     only until all callers have migrated.
     """
-    return Dispatcher(config).dispatch(
-        event, payload, repo_cfg, delivery_id=delivery_id, oracle=oracle
+    return Dispatcher(config, repo_cfg).dispatch(
+        event, payload, delivery_id=delivery_id, oracle=oracle
     )
 
 
@@ -2654,7 +2657,7 @@ def create_task(
     log.info("creating task: %s", prompt[:100])
     new_task = _tasks.add(title=prompt, task_type=task_type, thread=thread)
     if dispatcher is not None:
-        dispatcher.launch_sync(repo_cfg)
+        dispatcher.launch_sync()
     else:
         launch_sync(config, repo_cfg, gh)
     if thread:
@@ -2696,8 +2699,8 @@ def backfill_missed_pr_comments(
     Use :meth:`Dispatcher.backfill_missed_pr_comments` directly; this free
     function is kept only until all callers have migrated.
     """
-    return Dispatcher(config, gh).backfill_missed_pr_comments(
-        repo_cfg, pr_number, gh_user=gh_user
+    return Dispatcher(config, repo_cfg, gh).backfill_missed_pr_comments(
+        pr_number, gh_user=gh_user
     )
 
 
@@ -2707,7 +2710,7 @@ def launch_sync(config: Config, repo_cfg: RepoConfig, gh: GitHub) -> None:
     Use :meth:`Dispatcher.launch_sync` directly; this free function is kept
     only until all callers have migrated.
     """
-    Dispatcher(config, gh).launch_sync(repo_cfg)
+    Dispatcher(config, repo_cfg, gh).launch_sync()
 
 
 def launch_worker(repo_cfg: RepoConfig, registry: WorkerRegistry) -> None:
--- a/src/fido/events.py
+++ b/src/fido/events.py
@@ -2638,3 +2638,61 @@ def launch_worker(repo_cfg: RepoConfig, registry: WorkerRegistry) -> None:
     """Wake the per-repo WorkerThread via the registry."""
     log.info("waking worker thread for %s", repo_cfg.name)
     registry.wake(repo_cfg.name)
+
+
+class Dispatcher:
+    """Typed collaborator that owns :func:`dispatch`, :func:`backfill_missed_pr_comments`,
+    and :func:`launch_sync`.
+
+    Accepts ``config`` and ``gh`` at construction time — these are stable
+    across requests.  ``repo_cfg`` is passed at call time because it is
+    resolved per-webhook from the incoming repository name.
+
+    This is the replacement for the ``_fn_dispatch`` callable-slot pattern on
+    :class:`~fido.server.WebhookHandler`.  Callers construct one instance and
+    hold it as a proper collaborator rather than reaching into a class-level
+    function slot.
+    """
+
+    def __init__(self, config: Config, gh: GitHub) -> None:
+        self._config = config
+        self._gh = gh
+
+    def dispatch(
+        self,
+        event: str,
+        payload: dict[str, Any],
+        repo_cfg: RepoConfig,
+        *,
+        delivery_id: str | None = None,
+        oracle: WebhookIngressOracle | None = None,
+    ) -> Action | None:
+        """Delegate to the module-level :func:`dispatch` free function."""
+        return dispatch(
+            event,
+            payload,
+            self._config,
+            repo_cfg,
+            delivery_id=delivery_id,
+            oracle=oracle,
+        )
+
+    def backfill_missed_pr_comments(
+        self,
+        repo_cfg: RepoConfig,
+        pr_number: int,
+        *,
+        gh_user: str,
+    ) -> int:
+        """Delegate to the module-level :func:`backfill_missed_pr_comments`."""
+        return backfill_missed_pr_comments(
+            self._config,
+            repo_cfg,
+            self._gh,
+            pr_number,
+            gh_user=gh_user,
+        )
+
+    def launch_sync(self, repo_cfg: RepoConfig) -> None:
+        """Delegate to the module-level :func:`launch_sync`."""
+        launch_sync(self._config, repo_cfg, self._gh)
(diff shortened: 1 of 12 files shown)


src/fido/events.py
```diff
@@ -2638,3 +2638,61 @@ def launch_worker(repo_cfg: RepoConfig, registry: WorkerRegistry) -> None:
...
+
+    def backfill_missed_pr_comments(
+        self,
+        repo_cfg: RepoConfig,
+        pr_number: int,
+        *,
+        gh_user: str,
+    ) -> int:
+        """Delegate to the module-level :func:`backfill_missed_pr_comments`."""
+        return backfill_missed_pr_comments(
```
rhencke: Why delegate to what you are removing?  Shouldn't it be the other way around?

rhencke (changes requested):

src/fido/events.py
FidoCanCode: You're right — the delegation is backwards. The logic should live on `Dispatcher`, and if the free functions need to stick around during migration, they should delegate to a Dispatcher instance, not the other way around. I'll flip it.

rhencke: The original functions should, ultimately, disappear.

FidoCanCode: Agreed — no compatibility shims. Once the callers are migrated (create_task, worker, WebhookHandler), the free functions go away.

FidoCanCode: Got it — task updated to match the direction from [your comment](https://github.com/FidoCanCode/home/pull/1328#discussion_r3178931619). Logic moves into Dispatcher, free functions delegate during migration.

src/fido/events.py
```diff
@@ -2638,3 +2638,61 @@ def launch_worker(repo_cfg: RepoConfig, registry: WorkerRegistry) -> None:
...
+
+    def __init__(self, config: Config, gh: GitHub) -> None:
+        self._config = config
+        self._gh = gh
+
+    def dispatch(
+        self,
+        event: str,
+        payload: dict[str, Any],
+        repo_cfg: RepoConfig,
```
rhencke: The fact that all three functions take RepoConfig as a parameter really says it needs to be injected in the ctor.

rhencke (changes requested):

src/fido/events.py
FidoCanCode: Task updated per [your feedback](https://github.com/FidoCanCode/home/pull/1328#discussion_r3178931619) — logic moves into Dispatcher methods, free functions become thin delegators during migration. On it.

src/fido/events.py
FidoCanCode: You're right. All three methods taking it is the signal. I'll move `repo_cfg` into the constructor — one Dispatcher per repo, constructed in the composition root. The webhook handler looks up the right one by repo name instead of passing `repo_cfg` at every call site.

FidoCanCode:
Pushed 5 more commits addressing the review feedback:

- **Flip delegation** (7e4db0a): Logic moved into `Dispatcher` methods — `dispatch` (~240 lines), `backfill_missed_pr_comments` (~58 lines), `launch_sync` (5 lines). Free functions are now thin shims that construct a temporary `Dispatcher` and call through. `TestDispatcher` tests rewritten to test actual behavior without patching free-function names.

- **Inject Dispatcher into Worker** (7ff726a): `Worker` / `WorkerThread` accept a `dispatcher` constructor parameter. `Worker.run()` calls `self._dispatcher.backfill_missed_pr_comments(pr_number, gh_user=...)` — no free-function import. Tests use `MagicMock(spec=Dispatcher)`.

- **repo_cfg into Dispatcher constructor** (fe1b42e): Per your feedback on comment #3178991522 — `repo_cfg` moved from per-call method parameters into `__init__`. `Dispatcher(config, repo_cfg, gh)` is constructed once per repo at startup. The `dispatchers` dict on `WebhookHandler` (replacing the single `dispatcher` attr) is keyed by repo name; `_do_post_inner` looks up the right one.

One task still pending: deleting the three free functions entirely. That requires updating the large `TestDispatch` test class (which calls the free functions directly) to go through `Dispatcher` instead. Holding it for a follow-up so this PR isn't unbounded — but happy to do it here if preferred.

FidoCanCode:
One more commit: deleted the three free-function shims entirely (08c61bf).

`TestDispatch*` tests updated to use `Dispatcher(cfg, repo_cfg).dispatch(event, payload)` directly, `TestBackfillMissedPrComments` and `TestLaunchSync` similarly updated, and `test_coverage_fills.py` now passes `dispatcher=MagicMock()` to `create_task` instead of patching `events.launch_sync`.

`Dispatcher.backfill_missed_pr_comments()` now passes `dispatcher=self` when it calls `create_task`, so the sync step goes through the Dispatcher's own `launch_sync()` method rather than a free function. The fallback `else: launch_sync(...)` branch in `create_task` is gone.

Everything the reviewer asked for is in — task queue is now empty. 🐾

src/fido/events.py
FidoCanCode: Task updated per your feedback at https://github.com/FidoCanCode/home/pull/1328#discussion_r3178931619 — already landed in 7e4db0a and 08c61bf. Logic lives in Dispatcher now, free functions are fully deleted.

src/fido/events.py
FidoCanCode: Done — landed in fe1b42e per your feedback at https://github.com/FidoCanCode/home/pull/1328#discussion_r3178991522. `Dispatcher(config, repo_cfg, gh)` with one per repo, `WebhookHandler.dispatchers` dict keyed by repo name.

rhencke:
Hand-rolled _FakeDispatcher instead of MagicMock() — Issue #1310 explicitly   called for this ("Replace @patch(...) with a hand-rolled _FakeDispatcher
  injected into the handler under test"), and the repo rule is no
  MagicMock/@patch anywhere. The diff adds ~7+ new dispatcher=MagicMock() call
  sites in tests/test_events.py and tests/test_coverage_fills.py. Introduce a
  small typed fake (e.g. recording
  dispatch/launch_sync/backfill_missed_pr_comments calls) and use it everywhere
  this PR currently passes MagicMock().

FidoCanCode:
You're right — the issue spec was explicit about this and I drifted into MagicMock out of convenience. Will introduce a `_FakeDispatcher` that records dispatch/backfill/launch_sync calls with typed fields, and swap out every `MagicMock(spec=Dispatcher)` / `dispatcher=MagicMock()` site in this PR.

rhencke:
Make dispatcher required, not Optional — create_task, queue_reply_tasks,
  Worker.__init__, WorkerThread.__init__ all take dispatcher: Dispatcher | None   = None and guard with if self._dispatcher is not None:. In production it's
  always wired; the optionality only exists to soften tests. Result: a forgotten   wiring silently no-ops launch_sync (tasks.json never reaches the PR body) and   skips the #794 backfill — exactly the "synthetic success from real failure"
  pattern CLAUDE.md prohibits. Drop the default and the is not None guards.

FidoCanCode:
Completely agree. The `None` defaults are test-softening that masks real wiring failures — if someone forgets to pass a dispatcher, the backfill and launch_sync silently don't happen. That's the exact anti-pattern CLAUDE.md calls out.

I'll drop the `= None` defaults and the `is not None` guards from all four sites (`create_task`, `queue_reply_tasks`, `Worker.__init__`, `WorkerThread.__init__`). Tests that weren't passing a dispatcher will need to pass the `_FakeDispatcher` — which is the right outcome anyway since it means every test site explicitly declares what dispatcher behavior it expects.

rhencke:
Make gh required on Dispatcher — Dispatcher(gh: GitHub | None = None) plus
  assert self._gh is not None in two of the three methods is a bifurcated type.
  The composition root always has gh. Drop the | None and the asserts.
  (Splitting the class is the other option, but overkill for this PR.)

FidoCanCode:
Agreed — the assert is a runtime stand-in for what should be a compile-time fact. `gh` is always present in the composition root; the `| None` is test-softening. Will make it required and drop the asserts alongside the other required-parameter cleanups.

rhencke:
Drop the assert dispatcher is not None, "dispatcher not wired — call run()   first" in _do_post_inner — startup preconditions belong at startup, not on
  every webhook. After fixing (2)/(3), use self.dispatchers[repo_name] and let a   missing key raise — that's a real coordination bug, not a hot-path concern.

rhencke:
Confirm real test coverage for Dispatcher.backfill_missed_pr_comments and
  Dispatcher.launch_sync — once the MagicMocks are replaced, verify these two
  methods are exercised by tests that actually observe their effects (task
  creation, sync launched), not just fake.launch_sync_called == True. The
  previous free-function tests should map cleanly onto the new methods.

FidoCanCode:
Right — once dispatcher and gh are both required, the dispatchers dict is fully populated at startup. `self.dispatchers[repo_name]` either succeeds or raises `KeyError`, which is the correct signal for a real coordination bug. The assert is just noise on every request. Will drop it as part of the same commit.

FidoCanCode:
Saw this and agree — `TestBackfillMissedPrComments` and `TestLaunchSync` already test real effects (task creation, `sync_tasks_background` called) through `Dispatcher` directly, not through the fake's recording fields. I'll confirm that coverage holds after the MagicMock replacement and flag anything that regresses. (Apologies for the delayed reply — my structured response generation tripped over itself a few times before this got through.)

FidoCanCode:
Done in 5e9e8dd. All five points landed together:

1. **`_FakeDispatcher` in `tests/fakes.py`** — hand-rolled fake with typed recording fields (`dispatch_calls`, `backfill_calls`, `launch_sync_calls`) and configurable side-effects. Replaces every `MagicMock(spec=Dispatcher)` and `dispatcher=MagicMock()` across all seven test files. Import path is `from tests.fakes import _FakeDispatcher` (not `from conftest import ...` — that was ambiguous given the second conftest in `rocq-python-extraction/test/`).

2. **`dispatcher` required everywhere** — keyword-only with no default in `Worker.__init__`, `WorkerThread.__init__`, `create_task`, `queue_reply_tasks`, cascaded through `_apply_reply_result` and `recover_reply_promises`. All `if self._dispatcher is not None:` guards gone.

3. **`gh` required on `Dispatcher`** — dropped `| None = None` and the two `assert self._gh is not None` guards.

4. **Direct index in `_do_post_inner`** — `self.dispatchers[repo_name]` raises `KeyError` on missing wiring, which is the right signal.

5. **Real coverage** — `TestBackfillMissedPrComments` and `TestLaunchSync` test actual effects (task creation, `sync_tasks_background` called) via `Dispatcher` directly, not through the fake's counter. Coverage holds at 100% after MagicMock removal.

rhencke:
- **\`_FakeDispatcher\` vs \`Dispatcher\` Protocol drift.**  The fake has the right method shapes today, but nothing structurally constrains them.  If \`Dispatcher.dispatch\`'s signature gains a parameter, pyright won't catch the fake going stale.  Cheap follow-up: declare a \`DispatcherProtocol\` in events.py (or \`fido/types.py\`), have both the real \`Dispatcher\` and the fake inherit/satisfy it, and type the rescope-chain parameters as the Protocol.
- **Recording field types could be tighter.**  \`dispatch_calls: list[tuple]\` and \`backfill_calls: list[dict[str, object]]\` work but lose information.  \`list[tuple[str, dict, str | None, object]]\` and a dataclass for backfill records would let tests assert on field names rather than positions.  Minor.
- **\`Worker.__init__\`'s other Optional params.**  Now that \`dispatcher\` is required, several siblings still fit the "required-in-prod, Optional-only-for-test-convenience" pattern: \`config: Config | None = None\`, \`repo_cfg: RepoConfig | None = None\`, \`provider_factory: DefaultProviderFactory | None = None\`, \`registry: ActivityReporter | None = None\`.  Per CLAUDE.md fail-closed and Insight #1339 (Optional collaborator parameters chain into silent fail-soft), these are the same anti-pattern.  Worth a follow-up sweep — though not in this PR.

FidoCanCode:
All three land well. A few thoughts:

**Protocol drift** — agree this is the real structural risk. The fake is right today but nothing enforces it stays right. A `DispatcherProtocol` in `fido/types.py` (or events.py) would let pyright catch drift at the call sites. Happy to file a follow-up issue and pick it up as the next task after this merges.

**Tighter recording types** — agree the `list[tuple]` is loose. A named tuple or dataclass for `dispatch_calls` entries would make assertions read like `call.event` instead of `call[0]`. Minor but worth cleaning up in the same follow-up that adds the Protocol, since they're both about making the fake a first-class typed object.

**Optional Worker params** — noted. Already on the radar per Insight #1339. Will leave it for the sweep rather than scope-creep this PR further.

Should I go ahead and file the Protocol / tighter-types follow-up?

rhencke:
Yes, please.

FidoCanCode:
On it — filing now.

FidoCanCode:
Filed #1351 to track the DispatcherProtocol + tighter recording types follow-up — that was a pure issue-filing task, so no code changes needed here. Work queue advances; PR stays as-is.

rhencke (approved):

PR #1354, opened 2026-05-04
Bug: lockfree status PR 1/6 — introduce RegistrySnapshot + AtomicReference primitive, migrate started_at to validate pattern (closes #1343)

### LINKED ISSUE #1343: Bug: lockfree status PR 1/6 — introduce RegistrySnapshot + AtomicReference primitive, migrate started_at to validate pattern
Foundation PR for the #1342 lockfree-status redesign.

## Scope

- Introduce `RegistrySnapshot` (frozen dataclass with `started_at_by_repo: Mapping[str, datetime]` field).
- Introduce `AtomicReference[T]` primitive (single-lock variant: a `threading.Lock` held only for the root pointer swap, observationally lock-free for readers).
- Migrate `started_at` from `_started_at_lock`-protected dict to a CAS-update on the snapshot.
- Update `Worker.thread_started_at` and `_collect_activities` to read via the snapshot.
- Leave all other registry locks untouched.

## Why first

- Owner-partitioned writes (each thread sets its own `started_at` once at boot, never updated again).
- No FSM interaction.
- Read-heavy from `/status.json`; write-rare.
- Smallest possible diff that exercises the full pattern: snapshot type, atomic root swap, retry loop, snapshot-read in the status path.

Validates the pattern end-to-end before larger migrations land.

Parent: #1342

FidoCanCode:
Fixes #1343.

Introduces an `AtomicReference[T]` primitive (single-lock pointer swap, observationally lock-free for readers) and a frozen `FidoState` dataclass built around per-repo `RepoState` snapshots held in a `frozendict`, then migrates `started_at` from a lock-protected dict to a CAS-update on the snapshot. Includes a `Lens` helper for ergonomic functional updates to the frozen tree, with a two-parameter `lens_update` method on `AtomicReference` — a path-selector lambda navigates to the target, a second argument supplies the value, eliminating `.set()` from callsites entirely. `FidoState` is the convergence target — `FidoStatus` will eventually fold into it, with `./fido status` serving as the living requirement for what `FidoState`'s serialized form must represent.

---

## Work queue

- [x] [Make root already be a Lens when passed to the update lambda (add a lens_update method or equivalent so callsites don't need explicit Lens(root) wrapping). Investigate replacing .set() with := or an operator overload — though := is not valid for subscript/attribute targets per PEP 572, so an alternative like operator overloading may be needed.](https://github.com/FidoCanCode/home/pull/1354#discussion_r3184085443) 
- [x] [Add frozendict as a dependency. Introduce a frozen RepoState dataclass with started_at and key (repo slug). Reshape FidoState from started_at_by_repo: Mapping[str, datetime] to repos: frozendict[str, RepoState]. Update all CAS-update sites in registry.py (start(), thread_started_at()) and tests to use the new shape.](https://github.com/FidoCanCode/home/pull/1354#discussion_r3183725591) 
- [x] [Change compare_and_set to return (success: bool, final_value: _T) tuple instead of just bool. Update the update() retry loop to use the returned value instead of calling get() again on CAS failure.](https://github.com/FidoCanCode/home/pull/1354#discussion_r3183363798) 
- [x] Introduce RegistrySnapshot, migrate started_at to snapshot CAS pattern 
- [x] Add AtomicReference[T] generic primitive with CAS and tests 
- [x] [Rename RegistrySnapshot to FidoState. Add doc comments noting that FidoStatus should eventually be replaced by serializing FidoState to JSON (even where this breaks the old format), and that ./fido status serves as the living requirement for what FidoState's schema must represent.](https://github.com/FidoCanCode/home/pull/1354#issuecomment-4373006598) 
</details>

### DIFF
--- a/src/fido/atomic.py
+++ b/src/fido/atomic.py
@@ -14,6 +14,8 @@ import threading
 from collections.abc import Callable
 from typing import Generic, TypeVar
 
+from fido.lens import Lens
+
 _T = TypeVar("_T")
 
 
@@ -94,3 +96,26 @@ class AtomicReference(Generic[_T]):
             success, old = self.compare_and_set(old, new)
             if success:
                 return new
+
+    def lens_update(
+        self,
+        selector: "Callable[[Lens[_T]], Lens[_T]]",
+        value: object,
+    ) -> _T:
+        """Install *value* at the path described by *selector* atomically.
+
+        *selector* receives a :class:`~fido.lens.Lens`-wrapped root and
+        returns a navigated :class:`~fido.lens.Lens` pointing at the target
+        field.  *value* is installed there; the reconstructed root is CAS'd
+        into the reference with automatic retry::
+
+            ref.lens_update(
+                lambda root: root.repos[name],
+                new_repo_state,
+            )
+
+        *selector* must be a pure function — the retry loop may call it
+        more than once under write contention.  Returns the reconstructed
+        root value that was successfully installed.
+        """
+        return self.update(lambda old: selector(Lens(old)).set(value))  # type: ignore[return-value]
--- a/src/fido/atomic.py
+++ b/src/fido/atomic.py
@@ -1,0 +1,89 @@
+"""AtomicReference[T] — single-lock pointer-swap primitive for lock-free reads.
+
+Writers hold ``_lock`` for the duration of the pointer swap so concurrent
+writes are serialized and no update is lost.  Readers call :meth:`get`
+without acquiring any lock — a single object-reference load is atomic on all
+CPython-supported architectures, including the free-threaded (no-GIL) build.
+
+``T`` should be an immutable value (frozen dataclass, namedtuple, ``None``,
+etc.) so that readers who hold a reference to the snapshot can inspect all
+fields without encountering races inside the value itself.
+"""
+
+import threading
+from collections.abc import Callable
+from typing import Generic, TypeVar
+
+_T = TypeVar("_T")
+
+
+class AtomicReference(Generic[_T]):
+    """A reference cell whose pointer swap is serialized by a single lock.
+
+    Readers are observationally lock-free — :meth:`get` is a bare attribute
+    read with no lock acquire.  Writers hold the internal ``_lock`` only for
+    the atomic swap, keeping write-side latency minimal.
+
+    The typical usage pattern is *update with retry*::
+
+        ref = AtomicReference(initial_snapshot)
+
+        def bump(old: Snapshot) -> Snapshot:
+            return replace(old, counter=old.counter + 1)
+
+        ref.update(bump)  # retries automatically if a concurrent writer raced
+
+    ``fn`` passed to :meth:`update` must be a **pure function** — the retry
+    loop may call it more than once under write contention.
+    """
+
+    def __init__(self, initial: _T) -> None:
+        self._value: _T = initial
+        self._lock = threading.Lock()
+
+    def get(self) -> _T:
+        """Return the current value.  Lock-free; safe to call from any thread."""
+        return self._value
+
+    def set(self, value: _T) -> None:
+        """Unconditionally replace the stored reference.
+
+        Serialized by the internal lock so concurrent :meth:`set` and
+        :meth:`compare_and_set` calls never interleave.
+        """
+        with self._lock:
+            self._value = value
+
+    def compare_and_set(self, expected: _T, new_value: _T) -> bool:
+        """Swap to *new_value* only if the current reference **is** *expected*.
+
+        Uses identity (``is``) rather than equality (``==``) for the check —
+        the comparison is always O(1) and unambiguous regardless of how ``_T``
+        defines ``__eq__``.
+
+        Returns ``True`` on success, ``False`` when the current value is not
+        *expected* (indicating a concurrent write raced ahead).
+        """
+        with self._lock:
+            if self._value is not expected:
+                return False
+            self._value = new_value
+            return True
+
+    def update(self, fn: Callable[[_T], _T]) -> _T:
+        """Apply *fn* to the current value and install the result atomically.
+
+        Reads the current reference, calls ``fn(current)`` to produce the
+        next value, then attempts a compare-and-set.  If a concurrent writer
+        changed the reference first, the loop retries with the new current.
+
+        Returns the value that was successfully installed.
+
+        ``fn`` must be a pure function — it may be called more than once
+        under write contention.
+        """
+        while True:
+            old = self.get()
+            new = fn(old)
+            if self.compare_and_set(old, new):
+                return new
--- a/src/fido/atomic.py
+++ b/src/fido/atomic.py
@@ -54,36 +54,43 @@ class AtomicReference(Generic[_T]):
         with self._lock:
             self._value = value
 
-    def compare_and_set(self, expected: _T, new_value: _T) -> bool:
+    def compare_and_set(self, expected: _T, new_value: _T) -> tuple[bool, _T]:
         """Swap to *new_value* only if the current reference **is** *expected*.
 
         Uses identity (``is``) rather than equality (``==``) for the check —
         the comparison is always O(1) and unambiguous regardless of how ``_T``
         defines ``__eq__``.
 
-        Returns ``True`` on success, ``False`` when the current value is not
-        *expected* (indicating a concurrent write raced ahead).
+        Returns a ``(success, value)`` tuple:
+
+        - On success: ``(True, new_value)`` — the swap happened.
+        - On failure: ``(False, current)`` — the current reference at the
+          moment of rejection.  Callers (e.g. :meth:`update`) can feed
+          ``current`` straight into the next attempt without an extra
+          :meth:`get` round-trip.
         """
         with self._lock:
             if self._value is not expected:
-                return False
+                return False, self._value
             self._value = new_value
-            return True
+            return True, new_value
 
     def update(self, fn: Callable[[_T], _T]) -> _T:
         """Apply *fn* to the current value and install the result atomically.
 
         Reads the current reference, calls ``fn(current)`` to produce the
         next value, then attempts a compare-and-set.  If a concurrent writer
-        changed the reference first, the loop retries with the new current.
+        changed the reference first, the loop retries — using the value
+        returned by the failed CAS directly, without an extra :meth:`get`.
 
         Returns the value that was successfully installed.
 
         ``fn`` must be a pure function — it may be called more than once
         under write contention.
         """
+        old = self.get()
         while True:
-            old = self.get()
             new = fn(old)
-            if self.compare_and_set(old, new):
+            success, old = self.compare_and_set(old, new)
+            if success:
                 return new
--- a/src/fido/registry.py
+++ b/src/fido/registry.py
@@ -2,11 +2,13 @@
 
 import logging
 import threading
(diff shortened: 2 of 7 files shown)


rhencke:
Maybe `FidoStatus` instead?

FidoCanCode:
`FidoStatus` is already taken — it's the display dataclass in `status.py` (the thing `fido status` formats and `/status.json` serializes). `RegistrySnapshot` is the internal coordination state that lives inside `WorkerRegistry` and gets atomically swapped. Different layers: one faces outward, the other is a concurrency primitive.

If the name feels too generic though, I could go with something like `RegistryState` or `RegistryRoot` instead?

rhencke:
We can call it `FidoState`. `FidoStatus` should go away eventually and be replaced by converting FidoState to JSON even where this breaks the old format.  `./fido status` should be updated to accomodate the new schema but also serves as a living requirement of what the new schema must be able to represent.  Might be worth something in the doc comments.

FidoCanCode:
`FidoState` it is. I like the trajectory — one canonical state type that serves both coordination and display, with `FidoStatus` eventually folding into it.

I'll rename `RegistrySnapshot` → `FidoState` and add doc comments noting that `FidoStatus` is on the way out, and `./fido status` is the living requirement for what `FidoState`'s serialized form needs to represent.

src/fido/atomic.py
```diff
@@ -0,0 +1,89 @@
...
+    def set(self, value: _T) -> None:
+        """Unconditionally replace the stored reference.
+
+        Serialized by the internal lock so concurrent :meth:`set` and
+        :meth:`compare_and_set` calls never interleave.
+        """
+        with self._lock:
+            self._value = value
+
+    def compare_and_set(self, expected: _T, new_value: _T) -> bool:
```
rhencke: I think the semantics are better if you instead returns (success: bool, final_value: _T).  Because then update doesn't have to ask for the value again with get().  What do you think?

FidoCanCode: Yeah, that's cleaner. Right now `update()` does `get()` → `fn()` → `compare_and_set()`, and on failure loops back to `get()` — which is a redundant read since `compare_and_set` already knows the current value at the moment it rejects. Returning `(success, current_value)` lets `update()` feed the losing CAS's observed value straight into the next `fn()` call.

I'll change `compare_and_set` to return `tuple[bool, _T]` and update `update()` to thread the value through the retry loop instead of re-reading.

FidoCanCode: Done — d878a63. `compare_and_set` now returns `tuple[bool, _T]`: `(True, new_value)` on success, `(False, current)` on failure. `update()` threads `current` straight into the next `fn()` call, no extra `get()`. Tests updated to unpack and assert on both fields.

src/fido/registry.py
```diff
@@ -27,6 +28,36 @@ def _utcnow() -> datetime:
...
+    directly to JSON — even where that changes the wire format — and retire
+    ``FidoStatus``.  The ``./fido status`` CLI output serves as the living
+    requirement: whatever ``FidoState`` carries must be sufficient to render
+    everything that ``./fido status`` currently shows.
+    """
+
+    started_at_by_repo: Mapping[str, datetime]
+
+
+_EMPTY_FIDO_STATE = FidoState(started_at_by_repo={})
```
rhencke: Maybe we add frozendict as a dependency: https://pypi.org/project/frozendict/

Then, we add `RepoState` with `started_at` and `key`: `repo slug`.  Then `FidoState(repos=frozendict({"rhencke/home": RepoState(...)})

FidoCanCode: I like this. Right now there's only `started_at` per repo, but the next PRs in the series migrate activities and crash records — at which point a flat `started_at_by_repo` mapping becomes awkward and we'd refactor anyway. Better to land the `RepoState` shape now.

`frozendict` also gives us a real guarantee the mapping can't be mutated through a stale snapshot reference, which matters more on 3.14t than it would under the GIL.

I'll add `frozendict`, introduce `RepoState(key: str, started_at: datetime)`, and reshape to `FidoState(repos: frozendict[str, RepoState])`.

FidoCanCode: Task updated from your [review comment](https://github.com/FidoCanCode/home/pull/1354#discussion_r3183427999) — adding `frozendict`, introducing `RepoState`, and reshaping `FidoState` to `repos: frozendict[str, RepoState]`. On it.

rhencke: Try again, task didn't add.

FidoCanCode: Got it — re-registering the task now. Sorry about the dropped ball.

FidoCanCode: Task picked up from your [review comment](https://github.com/FidoCanCode/home/pull/1354#discussion_r3183458985) — adding `frozendict`, introducing `RepoState`, and reshaping `FidoState`. Working on it now.

rhencke: One more try

FidoCanCode: Third time's the charm — re-registering. The task keeps getting acknowledged in replies but not landing in the work queue. Apologies for the runaround.

FidoCanCode: Working on the frozendict/RepoState reshape from your [review comment](https://github.com/FidoCanCode/home/pull/1354#discussion_r3183725591).

FidoCanCode: Got it! Here's the rewrite:

Done — 90d5218. Added `frozendict` as a dep, introduced a frozen `RepoState(key, started_at)` dataclass, and reshaped `FidoState` from `started_at_by_repo: Mapping[str, datetime]` to `repos: frozendict[str, RepoState]`. All CAS-update sites in `registry.py` updated. Tests still at 100% coverage — good code, good boy.

src/fido/registry.py
```diff
@@ -223,15 +278,26 @@ def start(self, repo_cfg: RepoConfig) -> None:
...
-            self._started_at[repo_cfg.name] = _utcnow()
+        _name = repo_cfg.name
+        _now = _utcnow()
+        self._state.update(
+            lambda old: FidoState(
+                repos=frozendict(
+                    {**old.repos, _name: RepoState(key=_name, started_at=_now)}
+                )
+            )
+        )
```
rhencke: I think this can be cleaner. Imagine if FidoState provided a helper update that like `newState = oldState.update(lambda root, value: root.repos["some/repo"] := RepoState(...))` but `root` is just a sneaky little class that returns itself but observes a list of path access in traversal order.  It then walks it backwards, constructing updated intermediate classes with data classes.copy or equivalent frozendict methods, finally returning a new FidoState generated from a (relatively) incrementally update to itself.  Because our updates will kind of be one owner at a time like this.  Of course updates would be encouraged a dataclass/dict worth at a time.

rhencke (changes requested):

src/fido/registry.py
FidoCanCode: This is a really nice idea — basically a functional lens that records the traversal path and reconstructs back up. Right now with one level of nesting the spread-and-reconstruct is tolerable, but once `RepoState` grows `activity`, `crash_record`, etc. the manual reconstruction at every CAS site is going to get noisy fast.

I think the right move is to file this as a follow-on issue for the lockfree series rather than cramming it into this PR — the current callsite works, and getting the lens right (especially around frozendict vs dataclass at each level) deserves its own focused pass. Want me to file that?

rhencke: I'd like it for this one.  It occurs to me the lambda doesn't even need value, just root passed in.

I think it will be smaller than you think and pay off big later.

FidoCanCode: Done — 6dfb704. You were right, it's small.

98 lines total: a `Lens[_R]` proxy that records `.attr` and `[key]` accesses, then `.set(value)` walks backwards reconstructing frozen copies via `dataclasses.replace` for dataclass nodes and `type(parent)({**parent, key: new})` for frozen mappings.

The registry callsite now reads:

```python
self._state.update(
    lambda root: Lens(root).repos[_name].set(
        RepoState(key=_name, started_at=_now)
    )
)
```

One fun thing I caught during testing: the forward pass only needs to walk to the *parent* of the leaf, not the leaf itself. If you're inserting a new key into a frozendict, the leaf doesn't exist yet — so `current[key]` would KeyError. Stopping one step short handles both insert and update cleanly.

FidoCanCode: Already landed — see 6dfb704 above.

src/fido/registry.py
```diff
@@ -223,15 +279,24 @@ def start(self, repo_cfg: RepoConfig) -> None:
         thread = self._factory(repo_cfg, provider=provider, session_issue=session_issue)
         with self._threads_lock:
             self._threads[repo_cfg.name] = thread
-        with self._started_at_lock:
-            self._started_at[repo_cfg.name] = _utcnow()
+        _name = repo_cfg.name
+        _now = _utcnow()
+        self._state.update(
+            lambda root: (
+                Lens(root).repos[_name].set(RepoState(key=_name, started_at=_now))
```
rhencke: Let's do two things:

1. `root` should already be a lens.
2. `.set()` should be `:=` which is legal in Python lambdas.

rhencke (changes requested):

src/fido/registry.py
FidoCanCode: Point 1 — yes, easy. A `lens_update` method on `AtomicReference` (or a standalone helper) that wraps the old value in `Lens` before passing it to the lambda. Callsite becomes:

```python
self._state.lens_update(
    lambda root: root.repos[_name].set(RepoState(key=_name, started_at=_now))
)
```

Point 2 — `:=` won't work here. The walrus operator (PEP 572) only supports simple name targets: `x := expr` is valid, but `root.repos[_name] := expr` is a `SyntaxError`. It can't target subscripts or attribute accesses.

If you want something shorter than `.set()`, we could overload an operator — `<<` or `|` — so it reads like `root.repos[_name] | RepoState(...)`. But `.set()` is 6 characters and immediately obvious. I'd keep it unless you feel strongly about the operator route.

FidoCanCode: Task updated from your [review comment](https://github.com/FidoCanCode/home/pull/1354#discussion_r3184085443) — adding `lens_update` so root arrives pre-wrapped as a Lens, and investigating operator syntax to replace `.set()`. On it.

rhencke: I'd rather just break it in two parameters, the lambda root: root.x.y, and second parameter the value to set it to.

FidoCanCode: Got it — that's cleaner. The path-selector lambda only describes where, the second argument is what:

```python
self._state.lens_update(
    lambda root: root.repos[_name],
    RepoState(key=_name, started_at=_now),
)
```

Under the hood: `selector(Lens(old)).set(value)`. Eliminates `.set()` from callsites entirely. Updating the task.

FidoCanCode: Task updated from your [review comment](https://github.com/FidoCanCode/home/pull/1354#discussion_r3184085443) — two-parameter `lens_update`: selector lambda picks the path, second arg is the value. `.set()` gone from callsites entirely. On it.

FidoCanCode: Done — 4699704. `lens_update` is two parameters now: a selector lambda picks the path, second arg is the value. Callsite looks like `self._state.lens_update(lambda root: root.repos[_name], RepoState(...))`. No more `.set()` at callsites — fetched that right out of there.

rhencke (approved):

PR #1405, opened 2026-05-05
Bug: lockfree status PR 2/6 — migrate _activity_lock and _crash_lock to RegistrySnapshot (closes #1344)

### LINKED ISSUE #1344: Bug: lockfree status PR 2/6 — migrate _activity_lock and _crash_lock to RegistrySnapshot
Second PR in the #1342 lockfree-status redesign.

## Scope

- Add `activity_by_repo: Mapping[str, WorkerActivity]` and `crash_by_repo: Mapping[str, WorkerCrash | None]` to `RegistrySnapshot`.
- Migrate `report_activity`, `get_all_activities`, `get_crash_info`, `set_crash_info` from `_activity_lock`/`_crash_lock` to CAS-update on the snapshot.
- `_collect_activities` reads activity + crash via the snapshot — biggest single observable speedup since this is most of what `/status.json` reads.
- Delete `_activity_lock` and `_crash_lock`.

## Why

Owner-partitioned: each worker writes only its own activity / crash. No FSM interaction. After this PR `/status.json` only touches `_threads_lock` and `_webhook_lock` (covered by PRs 5 and 3).

Depends on: PR 1/6 (foundation).

Parent: #1342

FidoCanCode:
Migrate `_activity_lock` and `_crash_lock` out of `WorkerRegistry` into the lock-free `FidoState` / `RepoState` snapshot. The class owns activity and crash values independently and publishes them into a prepopulated zero-value tree via pure lens writes — no read-modify-write through FidoState, no `None` fields, no missing-key branches. Also collapses `update` and `lens_update` on `AtomicReference` into a single `update(selector, value)` method, and has `_collect_activities` pull activity + crash + started_at from a single snapshot to avoid cross-generation reads.

Fixes #1344.

---

## Work queue

- [x] Update _collect_activities to read activity and crash from single snapshot 
- [x] [Use lens_update method instead of raw update for the CAS operations. Inline update into lens_update, delete old update, rename lens_update to update.](https://github.com/FidoCanCode/home/pull/1405#discussion_r3190727993) 
- [x] Migrate record_crash and get_crash_info to CAS-update, delete _crash_lock 
- [x] Migrate report_activity and is_stale to CAS-update, delete _activity_lock 
- [x] Make WorkerActivity and WorkerCrash frozen, add fields to RepoState 
</details>

### DIFF
--- a/src/fido/registry.py
+++ b/src/fido/registry.py
@@ -4,7 +4,7 @@ import logging
 import threading
 from collections.abc import Callable, Generator
 from contextlib import contextmanager
-from dataclasses import dataclass, replace
+from dataclasses import dataclass
 from datetime import datetime, timezone
 from typing import TYPE_CHECKING
 
@@ -30,6 +30,9 @@ def _utcnow() -> datetime:
     return datetime.now(tz=timezone.utc)
 
 
+_EPOCH = datetime(1970, 1, 1, tzinfo=timezone.utc)
+
+
 @dataclass(frozen=True)
 class WorkerActivity:
     """Snapshot of what one worker is currently doing.
@@ -50,11 +53,28 @@ class WorkerCrash:
 
     Frozen so instances can be stored inside frozen :class:`RepoState`
     without breaking the immutability guarantee of the atomic snapshot.
+
+    The zero value (``_ZERO_CRASH``) represents "never crashed" — readers
+    check ``death_count == 0`` rather than testing for ``None``.
     """
 
     death_count: int
     last_error: str
     last_crash_time: datetime
+
+
+_ZERO_CRASH = WorkerCrash(death_count=0, last_error="", last_crash_time=_EPOCH)
+
+
+def _zero_activity(repo_name: str) -> WorkerActivity:
+    """Return a zero-value :class:`WorkerActivity` for *repo_name*.
+
+    Repo-specific because :attr:`WorkerActivity.repo_name` is part of the
+    value.  Readers check ``what == ""`` to detect the zero sentinel.
+    """
+    return WorkerActivity(
+        repo_name=repo_name, what="", busy=False, last_progress_at=_EPOCH
+    )
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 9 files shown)


src/fido/registry.py
rhencke: Use update lens method like the repo status does.  We should probably just inline update into update lens and the n delete old update and rename lens update to just update.

rhencke (changes requested):

src/fido/registry.py
FidoCanCode: Agreed on the direction, but the three `update` call sites here can't switch to `lens_update` as-is — they're all read-modify-write:

- `record_crash` needs to read the existing `death_count` and increment it. `lens_update` takes a fixed value, so on CAS retry it'd install the stale count.
- `report_activity` and `start` need to handle "RepoState doesn't exist yet" — the Lens forward pass would KeyError on `root.repos[_name]` if no RepoState has been created.

`lens_update(selector, value)` is a pure write. These need a read-modify-write.

I think the right fix is to add a `Lens.modify(fn)` variant that receives the old leaf and returns the new one, then wire `lens_update` to accept either a value or a callable. That would cover both patterns cleanly. Want me to file that as a follow-up issue? Or should we tackle it in this PR?

rhencke: The real fix is to have the class own and store the values independent of FidoState.  Then, updates are easy because they never rely on old FidoState

FidoCanCode: Got it — task updated for [your comment](https://github.com/FidoCanCode/home/pull/1405#discussion_r3190727993). Will inline `update` into `lens_update`, delete the old one, and rename.

rhencke: The second fix is to populate a full FidoState tree on start with empty/zero values and forbid None values.

Little tricky with frozendict but even those we can prepopulate.

FidoCanCode: That makes sense — if the class owns the authoritative values and just publishes them into FidoState, every CAS site becomes a pure write. No read-modify-write, no stale-retry problem. Will restructure accordingly.

FidoCanCode: Noted — task for [your feedback](https://github.com/FidoCanCode/home/pull/1405#discussion_r3190727993) updated to also prepopulate the full FidoState tree on start with empty/zero values and forbid None.

rhencke: Will need to prepopulate the dict with known repos too

FidoCanCode: Hey Rob — I saw this and already started implementing it, but the structured reply got tangled up in the response format. Short version: agreed, prepopulating the full tree with zero values so lens paths never hit missing keys. Working on it now alongside the class-owned-values restructure.

FidoCanCode: Yep — `start()` creates the full `RepoState` with zero values and installs it into the `repos` frozendict, so by the time any writer calls `report_activity` or `record_crash` the key is guaranteed to exist. No missing-key branches anywhere downstream.

FidoCanCode: Task updated for [your comment](https://github.com/FidoCanCode/home/pull/1405#discussion_r3190727993) — now covers the full restructure: class-owned values published via pure lens writes, prepopulated zero-value tree, and the update/lens_update collapse.

FidoCanCode: Task for [your review comment](https://github.com/FidoCanCode/home/pull/1405#discussion_r3190727993) updated — merged with the crash migration into a single commit since the working tree already has both implemented. `_collect_activities` single-snapshot read follows as the final task.

FidoCanCode: Okay! Inlined `update` into `lens_update` (now just `update(selector, value)`), deleted the old fn-based `update`, renamed `lens_update` → `update`. Class-owned `_crash_records` and zero-value tree prepopulation eliminate all the read-modify-write CAS sites. Committed in 7684ffe — this one felt really good to land.

rhencke:
It's getting smaller - good sign.

FidoCanCode:
Yeah — dropping the two locks, the None-guarded read paths, and the read-modify-write CAS branches all at once adds up. The class-owned values pattern made a lot of defensive code unnecessary.

rhencke (approved):

PR #1409, opened 2026-05-05
Bug: lockfree status PR 3/6 — migrate _webhook_lock (per-repo webhook activities) to RegistrySnapshot (closes #1345)

### LINKED ISSUE #1345: Bug: lockfree status PR 3/6 — migrate _webhook_lock (per-repo webhook activities) to RegistrySnapshot
Third PR in the #1342 lockfree-status redesign.

## Scope

- Add `webhook_activities_by_repo: Mapping[str, Sequence[WebhookActivity]]` to `RegistrySnapshot`.
- Migrate `webhook_activity` (the contextmanager that appends + removes), `report_webhook_activity`, related accessors from `_webhook_lock` to CAS-update on the snapshot.
- Delete `_webhook_lock`.

## Why

Per-repo, owner-partitioned. The contextmanager pattern needs the most thought: enter appends, exit removes — two CAS updates per webhook handler invocation. Bounded retry loops on each.

Depends on: PR 1/6.

Parent: #1342

FidoCanCode:
Fixes #1345.

Add `webhook_activities` field to `RepoState` and make `WebhookActivity` frozen so webhook entries live in the immutable snapshot tree. Keep authoritative webhook-activity state on `WorkerRegistry` as a plain locked dict and publish into `FidoState` via a zero-parameter method — no CAS read-modify-write loops, no silent early-return guards. Add `slots=True` to all frozen dataclasses in registry.py and update the `/status.json` endpoint to read directly from the snapshot.

Fixes #1345.

---

## Work queue

- [x] [Remove the `= ()` default from `RepoState.webhook_activities`, making it a required field. Update `start()` in WorkerRegistry to pass `webhook_activities=()` explicitly at the RepoState construction site.](https://github.com/FidoCanCode/home/pull/1409#discussion_r3191688085) 
- [x] [In RepoState, replace `webhook_activities: tuple[WebhookActivity, ...] = field(default_factory=tuple)` with `webhook_activities: tuple[WebhookActivity, ...] = ()`. The `field()` wrapper is unnecessary for an immutable default value.](https://github.com/FidoCanCode/home/pull/1409#discussion_r3191651860) 
- [x] [In _publish_webhook_activities, remove the early-return guard that skips the update when the repo is not yet in the FidoState snapshot. Let it crash instead — FidoState should always have the repo prepopulated by start() before any webhook_activity call. Update all TestWebhookActivity tests that currently call webhook_activity without a preceding start() to call start() first.](https://github.com/FidoCanCode/home/pull/1409#discussion_r3191585122) 
- [x] Simplify webhook activity writes: drop _modify_webhook_activities, use owner-held state + lens publish 
- [x] [Update the 'Simplify webhook activity writes' task: the pattern should be — record the change on the local object (e.g. a plain dict on WorkerRegistry), then call a zero-parameter method that returns a correct frozen dataclass value, and pass that to self._state.update for wholesale replacement. Never read old FidoState values to calculate new ones.](https://github.com/FidoCanCode/home/pull/1409#discussion_r3191463938) 
- [x] [Add slots=True to all frozen dataclasses in registry.py: WorkerActivity, WorkerCrash, WebhookActivity, RepoState, FidoState, WebhookActivityHandle.](https://github.com/FidoCanCode/home/pull/1409#discussion_r3191403989) 
- [x] Migrate webhook activities to FidoState CAS-snapshot 
- [x] Read webhook activities from snapshot in status endpoint 
</details>

### DIFF
--- a/src/fido/registry.py
+++ b/src/fido/registry.py
@@ -35,7 +35,7 @@ def _utcnow() -> datetime:
 _EPOCH = datetime(1970, 1, 1, tzinfo=timezone.utc)
 
 
-@dataclass(frozen=True)
+@dataclass(frozen=True, slots=True)
 class WorkerActivity:
     """Snapshot of what one worker is currently doing.
 
@@ -49,7 +49,7 @@ class WorkerActivity:
     last_progress_at: datetime
 
 
-@dataclass(frozen=True)
+@dataclass(frozen=True, slots=True)
 class WorkerCrash:
     """Running record of unexpected worker deaths for one repo.
 
@@ -79,7 +79,7 @@ def _zero_activity(repo_name: str) -> WorkerActivity:
     )
 
 
-@dataclass(frozen=True)
+@dataclass(frozen=True, slots=True)
 class WebhookActivity:
     """One in-flight webhook handler running alongside the worker.
 
@@ -104,7 +104,7 @@ class WebhookActivity:
     thread_id: int
 
 
-@dataclass(frozen=True)
+@dataclass(frozen=True, slots=True)
 class RepoState:
     """Per-repo sub-snapshot within :class:`FidoState`.
 
@@ -142,7 +142,7 @@ class RepoState:
     webhook_activities: tuple[WebhookActivity, ...] = field(default_factory=tuple)
 
 
-@dataclass(frozen=True)
+@dataclass(frozen=True, slots=True)
 class FidoState:
     """Atomically-swapped coordination snapshot owned by :class:`WorkerRegistry`.
 
@@ -172,7 +172,7 @@ class FidoState:
 _EMPTY_FIDO_STATE = FidoState(repos=frozendict())
 
 
-@dataclass(frozen=True)
+@dataclass(frozen=True, slots=True)
 class WebhookActivityHandle:
     """Opaque handle for updating one in-flight webhook activity safely."""
 
--- a/src/fido/registry.py
+++ b/src/fido/registry.py
@@ -4,7 +4,7 @@ import logging
 import threading
 from collections.abc import Callable, Generator
 from contextlib import contextmanager
-from dataclasses import dataclass, field
+from dataclasses import dataclass
 from dataclasses import replace as dc_replace
 from datetime import datetime, timezone
 from typing import TYPE_CHECKING
@@ -138,7 +138,7 @@ class RepoState:
     started_at: datetime
     activity: WorkerActivity
     crash_record: WorkerCrash
-    webhook_activities: tuple[WebhookActivity, ...] = field(default_factory=tuple)
+    webhook_activities: tuple[WebhookActivity, ...] = ()
 
 
 @dataclass(frozen=True, slots=True)
--- a/src/fido/registry.py
+++ b/src/fido/registry.py
@@ -4,7 +4,8 @@ import logging
 import threading
 from collections.abc import Callable, Generator
 from contextlib import contextmanager
-from dataclasses import dataclass
+from dataclasses import dataclass, field
+from dataclasses import replace as dc_replace
 from datetime import datetime, timezone
 from typing import TYPE_CHECKING
 
@@ -14,6 +15,7 @@ from fido.atomic import AtomicReference
 from fido.config import Config, RepoConfig
 from fido.github import GitHub
 from fido.issue_cache import IssueTreeCache
+from fido.lens import Lens
 from fido.provider import PromptSession, Provider
 from fido.rocq import handler_preemption as preemption_fsm
 from fido.rocq import worker_registry_crash as registry_fsm
@@ -78,6 +80,31 @@ def _zero_activity(repo_name: str) -> WorkerActivity:
 
 
 @dataclass(frozen=True)
+class WebhookActivity:
+    """One in-flight webhook handler running alongside the worker.
+
+    Created when ``_process_action`` starts handling a webhook action; removed
+    when that handler returns (success or failure).  Surfaced in ``fido
+    status`` as a sub-bullet under the repo so we can see what's being
+    handled beyond the worker's own task.
+
+    *thread_id* is :func:`threading.get_ident` captured at context entry so
+    status display can match this webhook to the active
+    :class:`~fido.provider.SessionTalker` (whose ``thread_id`` field is from
+    the same call) — letting the CLI attach claude stats to the specific
+    webhook line that's driving claude.
+
+    Frozen so instances can be stored inside frozen :class:`RepoState`
+    without breaking the immutability guarantee of the atomic snapshot.
+    """
+
+    handle_id: int
+    description: str
+    started_at: datetime
+    thread_id: int
+
+
+@dataclass(frozen=True)
 class RepoState:
     """Per-repo sub-snapshot within :class:`FidoState`.
 
@@ -95,6 +122,10 @@ class RepoState:
     repo.  Initialised to ``_ZERO_CRASH`` (``death_count=0``) by
     :meth:`start`; the watchdog replaces it on each crash.
 
+    *webhook_activities* is the tuple of in-flight webhook handlers for this
+    repo.  Managed via CAS-update by
+    :meth:`~WorkerRegistry._modify_webhook_activities`; starts empty.
+
     No field is ``None`` — the full tree is prepopulated by :meth:`start`
     with zero values so lens-path writes never encounter missing keys.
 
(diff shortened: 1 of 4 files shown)


src/fido/registry.py
rhencke, line 23: Needed?

rhencke: Should all these have slots=True?

src/fido/registry.py
```diff
@@ -423,6 +431,51 @@ def record_crash(self, repo_name: str, error: str) -> None:
         _name = repo_name
         self._state.update(lambda root: root.repos[_name].crash_record, new_crash)
 
+    def _modify_webhook_activities(
```
rhencke: This is far too complicated.  On startup a zero value FidoState should already have a frozendict with pointers to the repo state.  There shouldn't need to be any special handling of tuples because we should not read old FidoState values to build new FidoState values.  Drop the whole method and do it with the .update Lens.  The webhook activity state should live in this class and then be pulled only when needed to update FidoState.

rhencke (changes requested):

src/fido/registry.py
FidoCanCode: Yep — that `TYPE_CHECKING` guard breaks a circular import (`fido.events` imports `WorkerRegistry` from here, so a runtime import of `Dispatcher` from `fido.events` would cycle). It predates this PR though, not something I introduced.

If you meant the `Lens` import at line 18 instead — that one's new here and currently used in `_modify_webhook_activities`, but it'll likely go away when I rework that method per your other comment.

FidoCanCode: Yes — `slots=True` is a free win on all of them. Less memory per instance, faster attribute access, and no downside since they're frozen with no inheritance. `dataclasses.replace()` and the Lens mechanism both work fine with slots.

I'll add `frozen=True, slots=True` to all six: WorkerActivity, WorkerCrash, WebhookActivity, RepoState, FidoState, WebhookActivityHandle.

FidoCanCode: Got it — task updated from [your comment](https://github.com/FidoCanCode/home/pull/1409#discussion_r3191403989). Will add `slots=True` to all six frozen dataclasses.

src/fido/registry.py
FidoCanCode: Saw this and agree — the CAS read-modify-write loop is overengineered when a plain owner-held list + lens publish does the job. I'll drop `_modify_webhook_activities`, keep the webhook activity state on the class, and just `self._state.update(...)` the tuple into `RepoState` after each mutation.

src/fido/registry.py
rhencke, line 1: This implementation is far too complex.  Instead of trying to figure out how to make granular updates - don't.  Just replace the entire dataclass value, using the existing update method.  Code smells include using old FidoState values to calculate new FidoState values.  Correct is wholesale replacement of one field captures what's needed.  You can trigger on whatever you want, but don't try to "just" update the one value you are responsible.  A better method would be to record the change in the local object, and then call a zero parameter method that returns a correct frozen dataclass value for the replacement.

rhencke (changes requested):

src/fido/registry.py
FidoCanCode: Picking this back up — the `slots=True` edit is staged, just need to verify tests and land it. On it from [your comment](https://github.com/FidoCanCode/home/pull/1409#discussion_r3191403989).

rhencke: In general circular imports are a smell

FidoCanCode: This is the same direction as your earlier comment at line 434 — I've got a task queued for it already. The extra detail here helps: local state is authoritative, zero-parameter method builds the frozen replacement, `update` installs it wholesale. No reading old snapshot values to compute new ones. Will follow that pattern.

FidoCanCode: Done — all six frozen dataclasses now have `slots=True`. Tests pass clean.

FidoCanCode: This is the same direction as your earlier comment at line 434 — I've got a task queued that folds both together: drop `_modify_webhook_activities`, keep authoritative state on `WorkerRegistry`, and publish via a zero-parameter method. The `Lens` import and the `TYPE_CHECKING` guard for `Dispatcher` should both go away as part of that cleanup.

FidoCanCode: Agreed — the circular import between `registry.py` and `events.py` predates this PR but it's worth fixing properly. I'll look at untangling it as part of the simplification task since the `TYPE_CHECKING` guard is a symptom of the same structural issue.

FidoCanCode: Added `slots=True` to all six frozen dataclasses in `registry.py`: `WorkerActivity`, `WorkerCrash`, `WebhookActivity`, `RepoState`, `FidoState`, `WebhookActivityHandle`. Tighter memory layout, no stray attributes sneaking in. Good fences make good dataclasses.

FidoCanCode: Done! Implemented in commit 027145e — the `_webhook_activities` dict is now the authoritative local store. Mutations happen under `_webhook_lock`, then `_publish_webhook_activities` reads from that local dict and does a wholesale `self._state.update`. Never sniffs old `FidoState` values to compute new ones. Clean ownership, no stale-read races. Tail is wagging.

src/fido/registry.py
```diff
@@ -423,6 +436,22 @@ def record_crash(self, repo_name: str, error: str) -> None:
...
+        tuple, and installs it at ``repos[name].webhook_activities`` via a
+        single lens write.  No-op if *repo_name* is not yet in the snapshot
+        (only possible in tests that bypass :meth:`start`).
+
+        Must be called under ``_webhook_lock``.
+        """
+        acts = tuple(self._webhook_activities.get(repo_name, []))
+        if repo_name not in self._state.get().repos:
+            return
+        _name = repo_name
```
rhencke: Let it crash.  FidoState should already have the frozendict populated for all repos. (Please ensure this.)

rhencke (changes requested):

src/fido/registry.py
FidoCanCode: Got it — removing the guard and letting it crash. I'll update the tests that currently call `webhook_activity` without `start()` to call `start()` first, matching what actually happens in production. The guard was papering over a test-setup gap.

FidoCanCode: Queued from [your comment](https://github.com/FidoCanCode/home/pull/1409#discussion_r3191585122) — removing the guard and updating tests to call `start()` first so the precondition is enforced rather than silently skipped.

FidoCanCode: Done — removed the early-return guard from `_publish_webhook_activities` (now crashes if repo not in snapshot, as it should), and updated all `TestWebhookActivity` tests and the `test_server.py` webhook-tracking test to call `start()` before using `webhook_activity`. Fail-closed feels right. Good fences make good dogs.

src/fido/registry.py
```diff
@@ -108,9 +138,10 @@ class RepoState:
     started_at: datetime
     activity: WorkerActivity
     crash_record: WorkerCrash
+    webhook_activities: tuple[WebhookActivity, ...] = field(default_factory=tuple)
```
rhencke: Why is `field` needed?

FidoCanCode: You're right — `field` isn't needed here. `()` is immutable so there's no shared-state risk, which is the only reason `field(default_factory=...)` exists. Should just be `webhook_activities: tuple[WebhookActivity, ...] = ()`. Will fix.

FidoCanCode: Yep — `field()` is overkill for an immutable default. Switching to plain `= ()` from [your comment](https://github.com/FidoCanCode/home/pull/1409#discussion_r3191651860).

FidoCanCode: Fetched the stick and dropped it clean — swapped `field(default_factory=tuple)` for a plain `= ()` on `webhook_activities` and shook off the unused `field` import.

src/fido/registry.py
```diff
@@ -108,9 +138,10 @@ class RepoState:
     started_at: datetime
     activity: WorkerActivity
     crash_record: WorkerCrash
+    webhook_activities: tuple[WebhookActivity, ...] = ()
```
rhencke: No defaults.  Defaults are misleading for this use case.

rhencke (changes requested):

src/fido/registry.py
FidoCanCode: Got it — removing the default entirely. The other fields don't have defaults and this one shouldn't either. I'll update `start()` and any other construction sites to pass `webhook_activities=()` explicitly.

FidoCanCode: Removed the `= ()` default from `RepoState.webhook_activities` and added `webhook_activities=()` explicitly at the construction site in `start()`. One place to set it, no sneaky defaults hiding in the dataclass. 🐕

rhencke (approved):

PR #1523, opened 2026-05-09
Bug: lockfree status PR 5/6 — migrate _threads_lock to RegistrySnapshot AND stop holding lock across thread.join() in stop_and_join (closes #1347)

### LINKED ISSUE #1347: Bug: lockfree status PR 5/6 — migrate _threads_lock to RegistrySnapshot AND stop holding lock across thread.join() in stop_and_join
Fifth PR in the #1342 lockfree-status redesign — closes the original `./fido status >1s` symptom.

## Scope

- Add `threads_by_repo: Mapping[str, ThreadHandle]` to `RegistrySnapshot`. The handle is a frozen dataclass with the underlying `WorkerThread` reference + metadata; mutating thread state stays inside `WorkerThread` itself.
- Migrate every `_threads_lock`-protected method (`get_thread`, `start`, `stop_and_join`, `stop_all`, watchdog FSM transitions, ...) to CAS-update on the snapshot.
- **Fix the lock-across-join antipattern in `stop_and_join`**: snapshot the thread reference (lock-free, via the snapshot), then call `thread.join(timeout=...)` outside any lock. The 30-second wait stops blocking everything else that reads the thread registry.
- Delete `_threads_lock`.

## Why this is the closing PR for the original bug

The specific contention path identified in #1342 — `/status.json`'s `thread_started_at` blocking on `_threads_lock` while `stop_and_join` holds it across a 30-second `thread.join` — is gone after this PR.

Depends on: PRs 1/6 and 2/6 (snapshot foundation + activity migration consumers).

Parent: #1342

FidoCanCode:
Fixes #1347.

Migrate every `_threads_lock`-protected method on `WorkerRegistry` to read thread references outside any status-contended lock — fixing the lock-across-join antipattern where `stop_and_join` held `_threads_lock` across a 30-second `thread.join`. Thread references live in a plain `_threads` dict (single-writer, no lock), `_threads_lock` is deleted, and `AtomicReader[FidoState]` is removed from `WorkerRegistry` entirely — only the status serialization path may hold the reader face.

Fixes #1347.

---

## Work queue

- [x] [Remove ThreadHandle and thread_handle from RepoState/FidoState. The mutable WorkerThread reference must not live inside the recursively-immutable snapshot. Keep a separate dict (or similar) for command-path thread lookups outside the snapshot. The lock-across-join fix (reading thread ref then calling join outside any status-contended lock) should still be preserved.](https://github.com/FidoCanCode/home/pull/1523#discussion_r3212120024) 
- [x] [Remove AtomicReader[FidoState] from WorkerRegistry entirely (it should only hold the updater). Document in CLAUDE.md that AtomicReader[FidoState] is exclusively for status.json and /status — no other consumer is permitted. Fix any code that violates this.](https://github.com/FidoCanCode/home/pull/1523#discussion_r3212128707) 
- [x] Migrate read-only _threads_lock methods to snapshot lookup 
- [x] Migrate stop_all and stop_and_join to snapshot — fix lock-across-join 
- [x] Delete _threads_lock and _threads dict 
- [x] Add ThreadHandle to RepoState and publish thread ref in start() 
</details>

### DIFF
--- a/src/fido/registry.py
+++ b/src/fido/registry.py
@@ -107,24 +107,6 @@ class WebhookActivity:
 
 
 @dataclass(frozen=True, slots=True)
-class ThreadHandle:
-    """Frozen reference to a :class:`~fido.worker.WorkerThread`.
-
-    Stored inside :class:`RepoState` so snapshot readers can reach the
-    thread's session metadata and liveness state without holding any lock.
-    The thread itself is mutable — ``ThreadHandle`` only freezes the
-    *reference*, not the object it points to.
-
-    ``thread_handle`` is ``None`` only in a zero-value :class:`RepoState`
-    that was constructed before :meth:`~WorkerRegistry.start` ran for that
-    repo.  After :meth:`~WorkerRegistry.start` completes the field is always
-    non-``None``.
-    """
-
-    thread: WorkerThread
-
-
-@dataclass(frozen=True, slots=True)
 class RepoState:
     """Per-repo sub-snapshot within :class:`FidoState`.
 
@@ -146,16 +128,15 @@ class RepoState:
     repo.  Published from the authoritative ``_webhook_activities`` dict on
     :class:`WorkerRegistry` via :meth:`~WorkerRegistry._publish_webhook_activities`; starts empty.
 
-    *thread_handle* is a :class:`ThreadHandle` wrapping the
-    :class:`~fido.worker.WorkerThread` for this repo.  Populated by
-    :meth:`~WorkerRegistry.start`; ``None`` only before the first start
-    (which never happens in the normal lifecycle).  All thread lookups
-    read from this field via the lock-free snapshot.
-
     As subsequent lock-free PRs migrate fields out of the per-lock dicts in
     :class:`WorkerRegistry`, those fields grow here (e.g. ``rescoping``).
     Each migration removes the corresponding lock and dict from
     ``WorkerRegistry.__init__``.
+
+    **Thread references do not belong here.**  :class:`WorkerThread` is mutable,
+    so storing a reference inside the recursively-immutable snapshot would break
+    the lock-free promise of :class:`FidoState`.  Thread lookups go through
+    ``WorkerRegistry._threads`` instead.
     """
 
     key: str
@@ -163,7 +144,6 @@ class RepoState:
     activity: WorkerActivity
     crash_record: WorkerCrash
     webhook_activities: tuple[WebhookActivity, ...]
-    thread_handle: ThreadHandle | None
 
 
 @dataclass(frozen=True, slots=True)
@@ -225,16 +205,21 @@ class WorkerRegistry:
     Threads are created via the injected *thread_factory* so tests can
     supply mock threads without patching module-level names.
 
-    Holds both faces of the atomic :class:`FidoState` cell: the updater for
-    writing thread and activity state into the snapshot, and the reader for
-    lock-free thread lookups via :meth:`_get_thread`.  Status serialisation
-    reads the snapshot directly via the reader passed to
-    :class:`~fido.server.WebhookHandler` at the composition root.
+    Holds only the **updater** face of the atomic :class:`FidoState` cell.
+    The reader face belongs exclusively to the status serialisation path
+    (``WebhookHandler.state_reader``).  ``WorkerRegistry`` never reads from
+    ``FidoState`` — it is a SCADA display projection for lock-free status
+    reporting, not a source of truth for app logic.
+
+    Thread references live in a plain ``_threads`` dict — written only by the
+    watchdog thread in :meth:`start`, read by command paths.  Mutable objects
+    like :class:`~fido.worker.WorkerThread` must never appear inside the
+    recursively-immutable :class:`FidoState` snapshot.
 
     Usage::
 
         state_reader, state_updater = create_fido_atomic()
-        registry = WorkerRegistry(my_factory, state_reader, state_updater)
+        registry = WorkerRegistry(my_factory, state_updater)
         registry.start(repo_cfg)   # create + start thread
         registry.wake("owner/repo")  # nudge thread to check for work
         registry.stop_all()          # clean shutdown
@@ -243,19 +228,20 @@ class WorkerRegistry:
     def __init__(
         self,
         thread_factory: Callable[..., WorkerThread],
-        state_reader: "AtomicReader[FidoState]",
         state_updater: "AtomicUpdater[FidoState]",
     ) -> None:
         self._factory = thread_factory
         self._status_lock = threading.Lock()
-        # _state_reader is the read face of the atomic FidoState cell.
-        # Lock-free thread lookups (wake, abort_task, get_session, …) call
-        # _get_thread() which reads thread_handle from the snapshot via this reader.
-        self._state_reader: "AtomicReader[FidoState]" = state_reader
         # _state_updater is the write face of the atomic FidoState cell.
         # Writers call _state_updater.update(selector, value) to CAS-install a
         # value at a Lens path.
         self._state_updater: AtomicUpdater[FidoState] = state_updater
+        # Per-repo thread references.  Written only by start() (called from
+        # the single watchdog thread during crash recovery, or sequentially
+        # during startup).  Read by command paths (wake, abort_task, stop_*,
+        # get_session, …).  No lock needed — single-writer, and readers only
+        # need a point-in-time reference.
+        self._threads: dict[str, WorkerThread] = {}
         # Owner-side crash records: the watchdog increments death_count here,
         # then publishes the result into FidoState via a pure lens write.
         # Only the watchdog thread writes; start() reads during crash recovery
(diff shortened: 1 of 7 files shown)


src/fido/registry.py
```diff
@@ -106,6 +106,24 @@ class WebhookActivity:
...
+    The thread itself is mutable — ``ThreadHandle`` only freezes the
+    *reference*, not the object it points to.
+
+    ``thread_handle`` is ``None`` only in a zero-value :class:`RepoState`
+    that was constructed before :meth:`~WorkerRegistry.start` ran for that
+    repo.  After :meth:`~WorkerRegistry.start` completes the field is always
+    non-``None``.
+    """
+
+    thread: WorkerThread
```
rhencke: This breaks the entire lock-free promise of FidoState to be recursively immutable through all its fields.

src/fido/registry.py
```diff
@@ -142,6 +163,7 @@ class RepoState:
     activity: WorkerActivity
     crash_record: WorkerCrash
     webhook_activities: tuple[WebhookActivity, ...]
+    thread_handle: ThreadHandle | None
```
rhencke: No, we can't just toss in mutable instances.  Extract the values to an immutable dataclass and have the thread handle able to produce one of those.

src/fido/registry.py
```diff
@@ -203,16 +225,16 @@ class WorkerRegistry:
...
-    Write-only relative to :class:`FidoState` — the registry holds only the
-    :class:`~fido.atomic.AtomicUpdater` face of the atomic cell.  The
-    :class:`~fido.atomic.AtomicReader` face lives in the composition root and
-    is passed directly to the status serialisation path so only the one
-    collaborator that actually reads holds the read face.
+    Holds both faces of the atomic :class:`FidoState` cell: the updater for
+    writing thread and activity state into the snapshot, and the reader for
+    lock-free thread lookups via :meth:`_get_thread`.  Status serialisation
+    reads the snapshot directly via the reader passed to
+    :class:`~fido.server.WebhookHandler` at the composition root.
```
rhencke: The only legal consumer of AtomicReader[FidoState] is status.json and /status.

Put it in claude.md and fix all violations of this invariant.

rhencke (changes requested):

src/fido/registry.py
```diff
@@ -221,19 +243,18 @@ class WorkerRegistry:
...
-        # thread, read from HTTP handler threads (wake, abort_task, get_session,
-        # etc.).  Python 3.14t has no GIL — dict reads and writes are not atomic.
-        self._threads_lock = threading.Lock()
         self._factory = thread_factory
         self._status_lock = threading.Lock()
-        # _state_updater is the write-only face of the atomic FidoState cell.
+        # _state_reader is the read face of the atomic FidoState cell.
+        # Lock-free thread lookups (wake, abort_task, get_session, …) call
+        # _get_thread() which reads thread_handle from the snapshot via this reader.
+        self._state_reader: "AtomicReader[FidoState]" = state_reader
```
rhencke: Illegal

rhencke (changes requested):

src/fido/registry.py
```diff
@@ -269,6 +290,18 @@ def __init__(
...
+
+        Reads :attr:`RepoState.thread_handle` via the atomic reader — no lock
+        acquisition needed.  All thread lookups go through this method.
+        """
+        repo_state = self._state_reader.get().repos.get(repo_name)
+        if repo_state is None or repo_state.thread_handle is None:
+            return None
+        return repo_state.thread_handle.thread
+
     def _registry_fsm_transition(
```
rhencke, line 288: Illegal

src/fido/registry.py
FidoCanCode: You're right. I talked myself into it with the 'freezes the reference, not the object' line in the docstring, but that's exactly the problem — a snapshot reader can reach through ThreadHandle into mutable WorkerThread state, and the whole point of the immutable snapshot is that readers can't observe partial mutations.

The status-read path only needs the frozen metadata already on RepoState (started_at, activity, crash_record). The thread reference is only needed for command paths (stop, join, wake, abort_task). Those don't need to be in the snapshot at all.

I think the right shape is: keep a simple `_threads: dict[str, WorkerThread]` for command-path lookups (single-writer from the watchdog thread, so lightweight or no lock needed for reads), and keep the thread reference out of FidoState entirely. The original bug fix still holds — stop_and_join reads the thread ref and calls join outside any lock that status reads contend on.

src/fido/registry.py
```diff
@@ -369,6 +399,7 @@ def start(self, repo_cfg: RepoConfig) -> None:
             activity=_zero_activity(_name),
             crash_record=crash_record,
             webhook_activities=(),
+            thread_handle=ThreadHandle(thread=thread),
```
rhencke: No, only immutable stuff.

rhencke (changes requested):

src/fido/registry.py
FidoCanCode: Yep, you're right — freezing the reference doesn't freeze what it points to. The thread handle is coming out of FidoState. Thread lookups will go through a plain dict on WorkerRegistry instead; the lock-across-join fix stays intact since `stop_and_join` still reads the ref and joins outside any status-contended lock.

Updated the task from your [review comment](https://github.com/FidoCanCode/home/pull/1523#discussion_r3212120024).

src/fido/registry.py
FidoCanCode: Saw this and fully agree — my structured response kept failing to parse, sorry about the noise. The task to pull `ThreadHandle` out of FidoState and use a plain dict for thread lookups is already queued. Working on it next.

src/fido/registry.py
FidoCanCode: This clicks with the ThreadHandle removal — once thread lookups go back to a plain dict, WorkerRegistry has no reason to hold the reader at all. It goes back to updater-only.

I'll add the invariant to CLAUDE.md as part of the same task: `AtomicReader[FidoState]` is exclusively for status serialization, full stop.

src/fido/registry.py
FidoCanCode: Task updated from your [review comment](https://github.com/FidoCanCode/home/pull/1523#discussion_r3212120024) — now also removes `AtomicReader[FidoState]` from WorkerRegistry and documents the reader-exclusivity invariant in CLAUDE.md.

src/fido/registry.py
FidoCanCode: Yep, agreed — `AtomicReader` on WorkerRegistry is illegal. Already folded into the queued task alongside the ThreadHandle removal. Sorry for the slow reply, hit a formatting snag on my end.

src/fido/registry.py
FidoCanCode: Yep — this reads from the snapshot reader that shouldn't be here. The queued task rewrites `_get_thread()` to read from `self._threads.get(repo_name)` instead, and removes the reader entirely.

src/fido/registry.py
FidoCanCode: Agreed — this line goes away in the queued task. `start()` will write to `self._threads[name]` instead, keeping the mutable reference out of the snapshot entirely.

src/fido/registry.py
rhencke: Classes own the messy mutable state and can provide immutable snapshots of the subset needed for lock free stats reporting so status calls don't hang.  But classes are not allowed to consume that immutable state because the immutable state is not the *source of truth* but an *immutable snapshot intended for SCADA display*. If you write code that depends on it, you risk problems because those classes could be totally rewritten and drop fields not relevant for stats or change representations.  If app code starts depending on it, now we have blown out the cost of iterating on status.json enhancements.

FidoCanCode: That's the cleanest framing I've heard for this boundary. The snapshot is a display projection for status reporting — not a data source app code should depend on. If app logic reads from it, every status.json enhancement becomes a cross-cutting change instead of a localized one.

The queued task already removes ThreadHandle and AtomicReader from WorkerRegistry. I'll make sure the CLAUDE.md documentation captures this exact framing: FidoState is a SCADA display snapshot, classes own the authoritative mutable state, and no app code may consume the snapshot.

FidoCanCode: ThreadHandle is out of FidoState, `_threads` dict is back as the single-writer lookup, AtomicReader is gone from WorkerRegistry, and the SCADA display invariant is documented in CLAUDE.md. Clean boundaries, no mutable refs sneaking into the snapshot. Feels good to get this one right.

src/fido/registry.py
FidoCanCode: Done — `AtomicReader` removed from `WorkerRegistry` entirely (constructor now takes only the updater face), `make_registry()` no longer accepts or passes `state_reader`, and CLAUDE.md documents the invariant: `AtomicReader[FidoState]` is exclusively for status serialization, no other consumer permitted. The boundary is clean now. Good fence, good neighbors.

PR #1544, opened 2026-05-09
Bug: lockfree status PR 5/6 — migrate _threads_lock to RegistrySnapshot AND stop holding lock across thread.join() in stop_and_join (closes #1347)

### LINKED ISSUE #1347: Bug: lockfree status PR 5/6 — migrate _threads_lock to RegistrySnapshot AND stop holding lock across thread.join() in stop_and_join

FidoCanCode:
Migrate every `_threads_lock`-protected method on `WorkerRegistry` to read thread references from a plain `_threads` dict (single-writer from watchdog/startup, no lock needed) — fixing the lock-across-join antipattern where `stop_and_join` held `_threads_lock` across a 30-second `thread.join`. Document the `AtomicReader[FidoState]` SCADA display invariant in CLAUDE.md: only the status serialization path may hold the reader face, thread references never enter the recursively-immutable snapshot. Tighten `get_thread_crash_error` to use direct indexing instead of `.get()` so a missing repo fails fast with a clear `KeyError`.

Fixes #1347.

---

## Work queue

- [x] [In `get_thread_crash_error`, replace `self._threads.get(repo_name)` with `self._threads[repo_name]` so a missing repo raises a clear KeyError instead of a silent None that crashes on the next line.](https://github.com/FidoCanCode/home/pull/1544#discussion_r3213897497) 
- [x] Fix stop_and_join and stop_all — snapshot thread ref, join outside lock 
- [x] Delete _threads_lock and remove lock from start() 
- [x] Remove _threads_lock from read-only thread-lookup methods 
- [x] Document AtomicReader[FidoState] SCADA display invariant in CLAUDE.md 
</details>

### DIFF
--- a/src/fido/registry.py
+++ b/src/fido/registry.py
@@ -548,18 +548,21 @@ class WorkerRegistry:
 
     def stop_all(self) -> None:
         """Request every managed thread to stop after its current iteration."""
-        with self._threads_lock:
-            threads = list(self._threads.values())
+        threads = list(self._threads.values())
         for thread in threads:
             thread.stop()
 
     def stop_and_join(self, repo_name: str, timeout: float = 30.0) -> None:
         """Stop the thread for *repo_name* and wait up to *timeout* seconds for it to exit.
 
+        Reads the thread reference from ``_threads`` outside any lock (single-writer
+        dict, so the read is safe), then calls ``stop()`` and ``join()`` outside any
+        lock — eliminating the original lock-across-join contention that caused
+        ``/status.json`` to stall for up to 30 seconds (#1342).
+
         No-op if no thread is registered for that repo.
         """
-        with self._threads_lock:
-            thread = self._threads.get(repo_name)
+        thread = self._threads.get(repo_name)
         if thread:
             thread.stop()
             thread.join(timeout=timeout)
--- a/src/fido/registry.py
+++ b/src/fido/registry.py
@@ -223,11 +223,14 @@ class WorkerRegistry:
         thread_factory: Callable[..., WorkerThread],
         state_updater: "AtomicUpdater[FidoState]",
     ) -> None:
+        # _threads is single-writer: only start() (called from the watchdog
+        # daemon thread or from the startup sequence) ever writes to it.
+        # HTTP handler threads (wake, abort_task, get_session, …) only read.
+        # CPython 3.14t's per-object dict lock makes individual dict.get() /
+        # dict.__setitem__ calls safe without an additional application-level
+        # lock, and the single-writer discipline means readers never observe a
+        # half-installed entry.  No _threads_lock needed.
         self._threads: dict[str, WorkerThread] = {}
-        # _threads_lock guards _threads: written by start() on the watchdog
-        # thread, read from HTTP handler threads (wake, abort_task, get_session,
-        # etc.).  Python 3.14t has no GIL — dict reads and writes are not atomic.
-        self._threads_lock = threading.Lock()
         self._factory = thread_factory
         self._status_lock = threading.Lock()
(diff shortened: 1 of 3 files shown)


src/fido/registry.py
```diff
@@ -551,32 +549,33 @@ def status_update(self) -> Generator[None, None, None]:
...
-        with self._threads_lock:
-            thread = self._threads.get(repo_name)
+        thread = self._threads.get(repo_name)
         return thread is not None and thread.is_alive()
 
     def get_thread_crash_error(self, repo_name: str) -> str | None:
         """Return the crash_error stored on the thread for *repo_name*, or None."""
-        with self._threads_lock:
-            thread = self._threads.get(repo_name)
+        thread = self._threads.get(repo_name)
```
rhencke: Silent crash one line down if get returns None.  Use `[repo_name]`.

src/fido/registry.py
```diff
@@ -860,8 +853,7 @@ def get_session(self, repo_name: str) -> PromptSession | None:
         when no worker thread is registered for the repo or the thread has
         not yet created its session.
         """
-        with self._threads_lock:
```
rhencke, line 863: Why is the lock now unnecessary?

rhencke (changes requested):

src/fido/registry.py
FidoCanCode: Yep, you're right — `.get()` here just delays the crash and makes it confusing. `KeyError` on the missing repo is way more useful than `AttributeError: 'NoneType' has no attribute ...` one line later. Will switch to `[repo_name]`.

src/fido/registry.py
FidoCanCode: `_threads` is single-writer: only `start()` (watchdog/startup thread) ever writes to it. HTTP handler threads only read. CPython 3.14t's per-object dict lock makes `dict.get()` safe without an application-level lock, and the single-writer discipline means readers never see a half-installed entry. The invariant is documented in a comment block on `self._threads` in `__init__` — want me to add a shorter note here at the call site too?

rhencke: No, I was just wondering, but thanks.

FidoCanCode: Got it, happy to explain!

src/fido/registry.py
FidoCanCode: Done — changed `self._threads.get(repo_name)` to `self._threads[repo_name]` in `get_thread_crash_error`, and updated the test to expect `KeyError` for an unknown repo. No more silent `None` sneaking past when it should bark. 🐕

rhencke (approved):

PR #1588, opened 2026-05-11
Bug: hold_for_handler in server.py doesn't publish ProviderSnapshot before/after session ownership change (closes #1570)

### LINKED ISSUE #1570: Bug: hold_for_handler in server.py doesn't publish ProviderSnapshot before/after session ownership change
## Background

When `server.py` holds the provider session via
`session.hold_for_handler()` during webhook turns
(`src/fido/server.py:904-916`), nothing publishes a
`ProviderSnapshot` before or after that ownership change.
Status now reads only from `repo_state.provider`, so
`session_owner` and the per-session counters can be stale
during active webhook turns — `fido status` shows "owned by
worker" or stale counters while a webhook handler is actually
in flight.

This predates the ProviderSnapshot work; webhooks never
published session state to the snapshot at all.  Surfaced
during the subtract-closed-subs PR (#1546) work.

The worker side is already covered: `WorkerThread.run()`
calls `self._registry.publish_provider_snapshot(...)` at loop
iteration boundaries (`src/fido/worker.py:4693/4700/4729`).
The webhook side has no equivalent.

## Fix description

Two implementation options.  Recommended: **B** — centralizes
the invariant so it covers every webhook handler path, not
just the ones that take `hold_for_handler`.

**Option A — handler-side explicit calls.**  In
`server.py`, around the `session.hold_for_handler()` block at
~line 909, add:

```python
with self.registry.webhook_activity(repo_cfg.name, description) as activity:
    if session is not None and needs_model:
        with session.hold_for_handler():
            self.registry.publish_provider_snapshot(repo_cfg.name)  # NEW
            try:
                self._process_action_inner(action, repo_cfg, activity)
            finally:
                self.registry.publish_provider_snapshot(repo_cfg.name)  # NEW
    else:
        self._process_action_inner(action, repo_cfg, activity)
```

Smaller diff but only covers handlers that take the session
lock.  Webhook flows that update repo state without the lock
still publish stale snapshots.

**Option B (Recommended) — `webhook_activity` does it.**  The
`WorkerRegistry.webhook_activity(repo_name, description)`
context manager in `src/fido/registry.py` is the single point
every webhook handler passes through.  Have it call
`self._publish_provider_snapshot(repo_name)` on `__enter__`
and again on `__exit__`.  No changes needed in `server.py`.

```python
# in registry.py, inside webhook_activity context manager
self._publish_provider_snapshot(repo_name)  # on enter
try:
    yield activity
finally:
    self._publish_provider_snapshot(repo_name)  # on exit
```

This guarantees snapshot freshness around any webhook
activity, including ones that don't take the session lock.

## Tradeoffs

- Option B publishes on every webhook activity, not just the
  ones that materially change ownership.  Cost is one snapshot
  call per activity (cheap; same call worker already makes per
  loop iteration).  Benefit is the invariant becomes
  structural rather than scattered.
- Option B requires `webhook_activity` to know the repo name
  — it already does (first argument).
- Both options leave the worker-side
  `publish_provider_snapshot` calls in `worker.py` untouched.
- Test coverage: existing ProviderSnapshot oracle / extracted
  tests still pass; new unit test on `webhook_activity` would
  cover the publish-on-enter / publish-on-exit invariant.

## Verification

- `./fido ci` green.
- Manual: while a webhook is mid-turn on a busy PR (e.g. a
  comment that triggers a multi-second synthesis), `./fido
  status` shows `session_owner=webhook` for the affected repo,
  not stale worker info.
- Unit test on `webhook_activity`: enter publishes snapshot
  with `session_owner` reflecting current ownership; exit
  publishes again to reflect post-handler state.
- Regression test: the worker-side publish calls aren't
  duplicated or doubled-counted by the new
  `webhook_activity` calls (different code paths;
  independent).

## Source

Filed by fido during PR #1546 (subtract-closed-subs) work as
an out-of-scope follow-up.

FidoCanCode:
Publishes a fresh ProviderSnapshot inside the `hold_for_handler()` bracket in server.py (Option A), so `fido status` clears stale worker ownership to None during webhook handler turns instead of showing the last worker publish. Tests assert the actual FidoState snapshot value during and after the hold, not just call ordering.

Fixes #1570.

---

## Work queue

- [x] [Rewrite the two new tests in test_server.py to assert the actual FidoState snapshot value during and after the hold_for_handler block, not just call ordering. Check that the provider snapshot reflects the expected state (e.g. session_owner transitions) at each phase rather than only verifying that publish was called twice in the right order.](https://github.com/FidoCanCode/home/pull/1588#discussion_r3216001128) 
- [x] [Move the provider snapshot publishes from webhook_activity (registry.py) into server.py, bracketing the hold_for_handler() call directly (Option A from the issue). Publish once immediately after hold_for_handler() enters (so the snapshot reflects handler ownership) and once in the finally block after hold_for_handler() exits. Rewrite tests to assert the status snapshot actually shows handler ownership during the held section, not just that _publish_provider_snapshot was called twice.](https://github.com/FidoCanCode/home/pull/1588#discussion_r3215912017) 
- [x] Publish ProviderSnapshot on webhook_activity enter/exit 
</details>

### DIFF
--- a/src/fido/registry.py
+++ b/src/fido/registry.py
@@ -619,8 +619,6 @@ class WorkerRegistry:
         with self._webhook_lock:
             self._webhook_activities.setdefault(repo_name, []).append(activity)
             self._publish_webhook_activities(repo_name)
-        if repo_name in self._threads:
-            self._publish_provider_snapshot(repo_name)
         try:
             yield handle
         finally:
@@ -630,8 +628,6 @@ class WorkerRegistry:
                     a for a in acts if a.handle_id != activity.handle_id
                 ]
                 self._publish_webhook_activities(repo_name)
-            if repo_name in self._threads:
-                self._publish_provider_snapshot(repo_name)
 
     def set_webhook_description(
         self, repo_name: str, handle_id: int, description: str
--- a/src/fido/registry.py
+++ b/src/fido/registry.py
@@ -619,6 +619,8 @@ class WorkerRegistry:
         with self._webhook_lock:
             self._webhook_activities.setdefault(repo_name, []).append(activity)
             self._publish_webhook_activities(repo_name)
+        if repo_name in self._threads:
+            self._publish_provider_snapshot(repo_name)
         try:
             yield handle
         finally:
@@ -628,6 +630,8 @@ class WorkerRegistry:
                     a for a in acts if a.handle_id != activity.handle_id
                 ]
                 self._publish_webhook_activities(repo_name)
+            if repo_name in self._threads:
+                self._publish_provider_snapshot(repo_name)
 
     def set_webhook_description(
         self, repo_name: str, handle_id: int, description: str
--- a/src/fido/server.py
+++ b/src/fido/server.py
@@ -913,7 +913,16 @@ class WebhookHandler(BaseHTTPRequestHandler):
                     # behind long worker turns.  Both ClaudeSession and
                     # CopilotCLISession implement ``hold_for_handler``.
                     with session.hold_for_handler():
-                        self._process_action_inner(action, repo_cfg, activity)
+                        # Publish immediately after acquiring the handler hold
+                        # so repo_state.provider reflects the current ownership
+                        # (clears any stale worker-owner from the last snapshot).
+                        self.registry.publish_provider_snapshot(repo_cfg.name)
+                        try:
+                            self._process_action_inner(action, repo_cfg, activity)
+                        finally:
+                            # Publish again on the way out so the snapshot is
+                            # fresh after the session returns to the worker.
+                            self.registry.publish_provider_snapshot(repo_cfg.name)
                 else:
                     self._process_action_inner(action, repo_cfg, activity)
         except provider.SessionLeakError:
--- a/tests/test_registry.py
+++ b/tests/test_registry.py
@@ -1074,57 +1074,6 @@ class TestWebhookActivity:
             assert len(acts) == 1
             assert acts[0].description == "updated"
 
-    def test_publishes_provider_snapshot_on_enter_and_exit(
-        self, tmp_path: Path
-    ) -> None:
-        """webhook_activity publishes a ProviderSnapshot on enter and on exit."""
-        _, updater = create_fido_atomic()
-        reg = WorkerRegistry(MagicMock(), updater)
-        reg.start(_repo("foo/bar", tmp_path))
-
-        publish_calls: list[str] = []
-        original = reg._publish_provider_snapshot
-
-        def spy(repo_name: str) -> None:
-            publish_calls.append(repo_name)
-            original(repo_name)
-
-        reg._publish_provider_snapshot = spy  # type: ignore[method-assign]
-
-        assert publish_calls == []
-        with reg.webhook_activity("foo/bar", "handling"):
-            # One call on enter
-            assert publish_calls == ["foo/bar"]
-        # One more call on exit
-        assert publish_calls == ["foo/bar", "foo/bar"]
-
-    def test_publishes_provider_snapshot_on_exit_after_exception(
-        self, tmp_path: Path
-    ) -> None:
-        """webhook_activity publishes a ProviderSnapshot on exit even when the body raises."""
-        import pytest as _pytest
-
-        _, updater = create_fido_atomic()
-        reg = WorkerRegistry(MagicMock(), updater)
-        reg.start(_repo("foo/bar", tmp_path))
-
-        publish_calls: list[str] = []
-        original = reg._publish_provider_snapshot
-
-        def spy(repo_name: str) -> None:
-            publish_calls.append(repo_name)
-            original(repo_name)
-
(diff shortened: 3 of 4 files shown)


src/fido/registry.py
rhencke, line 1: /home/rhencke/workspace/home/src/fido/registry.py:622 publishes
  the provider snapshot outside the ownership window it is meant to
  report. In /home/rhencke/workspace/home/src/fido/server.py:908,
  webhook_activity() is entered before session.hold_for_handler(), and the  new exit publish runs after that handler hold has already been released.  So this PR records snapshots before acquisition and after release, but
  never while the webhook actually owns the provider session.
  That means the original bug can still happen: during the webhook body,
  repo_state.provider.session_owner can remain stale from the worker. The
  fix should bracket the actual hold_for_handler() ownership change, or
  make hold_for_handler()/session ownership changes trigger the snapshot
  publish directly. The new tests should assert the status snapshot
  changes during the held section, not only that
  _publish_provider_snapshot() was called twice.

rhencke, line 1: even if the publish timing is fixed,
  ProviderSnapshot.session_owner still cannot report webhook ownership.
  _publish_provider_snapshot() reads thread.session_owner in /home/
  rhencke/workspace/home/src/fido/registry.py:554, which delegates down to  provider owner; the provider implementations return None whenever the
  active talker is not a worker, e.g. /home/rhencke/workspace/home/src/
  fido/claude.py:783, /home/rhencke/workspace/home/src/fido/codex.py:699,
  and /home/rhencke/workspace/home/src/fido/copilotcli.py:981.
  So this PR cannot make session_owner show webhook; at best it would
  clear a stale worker owner to None. If webhook identity is supposed to
  be visible, either ProviderSnapshot/session_owner needs to encode
  webhook ownership, or the issue should target the separate live
  claude_talker field from /home/rhencke/workspace/home/src/fido/
  server.py:1365.

rhencke (changes requested):

src/fido/registry.py
FidoCanCode: You're right. I traced the call sequence and the publish timing is wrong.

`webhook_activity.__enter__` fires at line 908, before `hold_for_handler()` acquires the session at line 915. My exit publish fires after `hold_for_handler()` has already released. So the snapshot is published twice outside the ownership window and zero times inside it — the handler-owned period is never captured.

The fix needs to bracket `hold_for_handler()` directly. I'll move the publishes into server.py around the `with session.hold_for_handler():` block (Option A from the issue), and rewrite the tests to assert the snapshot reflects actual ownership during the held section.

FidoCanCode: Got it — task updated to match your [review feedback](https://github.com/FidoCanCode/home/pull/1588#discussion_r3215912017). Moving the publishes into server.py around `hold_for_handler()` directly, and rewriting the tests to verify ownership during the held section.

FidoCanCode: Saw this and agree — the timing fix (Option A) is the right move, and session_owner showing None during handler ownership is strictly better than stale worker info. The deeper limitation (provider implementations filtering out non-worker talkers so we can never show "webhook" as the owner) is a separate issue worth tracking. My earlier structured response failed to generate valid JSON, sorry about that — the task is already queued with the right scope.

FidoCanCode: Done — moved the provider snapshot publishes out of `webhook_activity` and into `server.py`, bracketing `hold_for_handler()` directly. Snapshot now publishes once after `hold_for_handler` acquires the lock and once in the `finally` before it releases. Two new tests verify the publish order: hold_enter → publish → (action body) → publish → hold_exit, including the exception path. Tail wag. 🐕

src/fido/server.py
rhencke, line 1: the new tests in /home/rhencke/workspace/
  home/tests/test_server.py:2087 prove call ordering, not the actual
  serialized status value.

rhencke (changes requested):

src/fido/server.py
FidoCanCode: Fair catch. The call-log approach only proves that `publish_provider_snapshot` fires in the right sequence — it says nothing about whether the snapshot actually contains useful information during the held window. A test that reads the FidoState snapshot value (checking `session_owner` / provider fields before, during, and after the hold) would catch content bugs the ordering check misses. I'll rewrite both tests to assert the actual snapshot state.

FidoCanCode: Got it — task updated to match the feedback at https://github.com/FidoCanCode/home/pull/1588#discussion_r3216001128. Will rewrite both tests to assert actual snapshot values instead of call ordering.

src/fido/registry.py
FidoCanCode: Rewrote both tests so they check actual `ProviderSnapshot` field values — `session_owner`, `session_alive`, `session_pid`, counters — read straight from `FidoState` after each publish. No more tracking call ordering. First test sets `session_owner='the-worker'`, second sets it to `None` to cover the provider-returns-None path. Much better sniff of the real state now.

rhencke (approved):

PR #1632, opened 2026-05-11
Gate: status shows live provider stats during an active turn (closes #1601)

### LINKED ISSUE #1601: Gate: status shows live provider stats during an active turn
## Background

The previous slices add live publication points. This issue is the regression gate that proves the user-visible invariant: `fido status` changes while a provider turn is still active, not only after the turn exits.

## Invariant

During an active provider turn, `fido status` reflects send/receive counter changes from `FidoState` snapshots before the worker/webhook turn completes.

## Scope

- Add an integration-style test with a controllable provider/session turn that pauses mid-turn.
- Observe status after send publication and after receive publication while the turn is still active.
- Assert the status serializer remains snapshot-only and does not reach into mutable session objects.
- Keep implementation changes minimal; this is primarily a gate over the previous slices.

## Acceptance Criteria

- The test fails if provider stats publish only at worker/webhook boundary exit.
- The test fails if sent/received counts are only visible after the provider turn completes.
- The test passes without adding direct provider/session reads to status serialization.

FidoCanCode:
Fixes #1601.

Add an integration-style gate test proving that `FidoState` snapshots reflect live provider stats (sent/received counts) **during** an active provider turn, not only after the turn completes. Uses a controllable `ClaudeSession` with a mid-turn barrier so the test can observe raw counter values through the `AtomicReader` while `consume_until_result()` is still in flight, plus a loose assertion that the counter values appear somewhere in the rendered status display output without pinning exact formatting.

Fixes #1601.

---

## Work queue

- [x] Add gate test: status shows live provider stats mid-turn 
- [x] [Keep the gate test focused on the raw counter values visible through the FidoState/AtomicReader snapshot mid-turn. Do not assert on the rendered ./fido status output or any display formatting. The invariant to gate is: session_sent_count and session_received_count are visible (non-zero) in the frozen ProviderSnapshot while consume_until_result() has not yet returned.](https://github.com/FidoCanCode/home/pull/1632#issuecomment-4425519638) 
</details>

### DIFF
--- a/tests/test_live_provider_stats.py
+++ b/tests/test_live_provider_stats.py
@@ -1,0 +1,281 @@
+"""Gate test: fido status reflects live provider stats during an active turn.
+
+Regression guard for the case where mid-turn snapshots showed zeros — the
+snapshot publisher must fire immediately after each send/receive, not only
+after the whole turn finishes.
+"""
+
+import json
+import queue
+import subprocess
+import threading
+from datetime import UTC, datetime
+from pathlib import Path
+from unittest.mock import MagicMock
+
+from frozendict import frozendict
+
+from fido.appstate import (
+    FidoState,
+    ProviderSnapshot,
+    RepoState,
+    WorkerActivity,
+    WorkerCrash,
+)
+from fido.atomic import AtomicUpdater, create_atomic
+from fido.claude import ClaudeSession
+from fido.provider import SnapshotPublisher
+from fido.rate_limit import GitHubLimit
+from fido.status import FidoStatus, RepoStatus, format_status
+
+_REPO = "owner/repo"
+_EPOCH = datetime(1970, 1, 1, tzinfo=UTC)
+
+
+def _make_fido_state(repo_name: str) -> FidoState:
+    """Return a minimal :class:`FidoState` with one pre-initialised repo entry."""
+    repo = RepoState(
+        key=repo_name,
+        started_at=_EPOCH,
+        activity=WorkerActivity(
+            repo_name=repo_name, what="", busy=False, last_progress_at=_EPOCH
+        ),
+        crash_record=WorkerCrash(death_count=0, last_error="", last_crash_time=_EPOCH),
+        webhook_activities=(),
+    )
+    return FidoState(
+        repos=frozendict({repo_name: repo}),
+        github_limits=GitHubLimit(),
+    )
+
+
+class _StatePublisher:
+    """Minimal :class:`~fido.provider.SnapshotPublisher` that writes into the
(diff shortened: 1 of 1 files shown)


rhencke:
I would encourage you to be very loose in your tests.  ./fido status is visual and subject to change (already have plans).  So just like sniffing out the numbers is enough for example.

FidoCanCode:
Good call. The test will just read the counter values straight out of the `FidoState` snapshot via the `AtomicReader` — assert that `provider.session_sent_count` and `provider.session_received_count` are non-zero while the turn is still in flight, nothing about rendered output. That way display changes don't touch this test at all.

rhencke:
But you can't avoid this testing display changes.  ./fido status is inherently visual and is what is being tested here.

rhencke:
Reread the issue body and compare it to your approach.

FidoCanCode:
Fair enough — I overshot. The gate should touch the display path since that's the actual user-visible surface. I'll include a loose assertion that the rendered status output contains the counter values (e.g. the number shows up somewhere in the output) without pinning exact formatting. Raw snapshot check for the core invariant, plus a light sniff that the numbers make it through to the display.

FidoCanCode:
Saw this and tried to respond a few times but kept tripping over myself. I think you're pointing out that the issue body says the test should assert the status serializer remains snapshot-only and doesn't reach into mutable session objects — which is different from adding display-path assertions. Could you confirm whether that's the direction, or rephrase what you'd like me to compare?

rhencke:
No, you got it right before.

rhencke:
> Fair enough — I overshot. The gate should touch the display path since that's the actual user-visible surface. I'll include a loose assertion that the rendered status output contains the counter values (e.g. the number shows up somewhere in the output) without pinning exact formatting. Raw snapshot check for the core invariant, plus a light sniff that the numbers make it through to the display.
> 
> 

This is the correct interpretation.

FidoCanCode:
Got it — loose display assertion plus raw snapshot check. That's the plan.

FidoCanCode:
Confirmed — raw snapshot check for the core mid-turn invariant, plus a loose display assertion that the numbers show up in rendered output. Already have this wired into the task description, will deliver it on the next worker turn.

FidoCanCode:
Done — the file has two tests: `test_mid_turn_snapshot_shows_nonzero_counts` (raw `AtomicReader` snapshot mid-turn, with a `threading.Event` barrier so we assert while `consume_until_result` is still blocked) and `test_display_path_shows_counter_values` (loose: just checks the integers appear somewhere in `format_status` output). Both pass and I am wagging so hard right now.

rhencke (approved):

Issue #1691, opened 2026-05-12
Bug: PR-comment queue accumulates orphan entries for merged PRs that the worker never drains


FidoCanCode:
## Observed

The durable PR-comment queue (`FidoStore` SQLite) accumulates entries for merged/closed PRs that never get drained. Counted ~17 orphans during recent watch mode.

## Cause

`Worker.handle_queued_comments` in `src/fido/worker.py:2553` only claims entries scoped to the current PR:

```python
queued = store.claim_next_pr_comment(
    owner="worker", repo=repo_ctx.repo, pr_number=pr_number
)
```

The worker only ever processes one PR at a time (the issue/PR it currently owns). If a webhook arrives for a different PR — for example a comment trickles in *after* a PR has been merged but before the queue entry is consumed — the entry sits in the queue forever.

Likely contributing scenarios:

1. Comment arrives between the merge webhook and the actual queue insertion (race).
2. A bot review lands on a PR that just got merged (Codex bot review on a closed PR).
3. Fido moves to a new PR while old-PR comments are still queued.

## Fix sketch

Two reasonable approaches:

- **Sweep on PR-close:** when handling `pull_request/closed` (merged or not), sweep the queue for entries scoped to that PR and either drop them or ack-as-no-op.
- **Drain across PRs at idle:** when the worker is between PRs and the queue has entries for a closed PR, drain them (post a courtesy reply or simply ack-and-drop).

The first is simpler and avoids cross-PR coordination. The second handles the race more cleanly but needs a "this PR is closed, don't synthesize a reply" code path.

## Why this matters

- The queue grows unboundedly across normal operation.
- An orphan entry can be confusing during diagnosis (`./fido status` will show pending PR-comment demand against a merged PR).
- On a future pickup of the same PR number (extremely unlikely on real GitHub) the orphans would unexpectedly fire.

PR #1695, opened 2026-05-12
Bug: sweep PR-comment queue on worker startup to clear post-merge orphans (closes #1691)

FidoCanCode:
## What was happening

`FidoStore.clear_pr_comment_queue` already runs on the `pull_request/closed` webhook, but comments can arrive *after* that webhook fires.

Confirmed in PR #1632: merged at 2026-05-11T22:23:26Z, comments arrived at 22:24:22Z — 56 seconds *after* close. Those entries sit in the queue forever because the worker only drains the currently-assigned PR, and Fido has long since moved on.

Live store at filing time has **17 such orphans** across PRs #1242, #1632, #1643 (all merged days ago).

## Fix

Add `Worker.sweep_orphan_pr_comments(repo)` called once per `WorkerThread` lifetime in the existing `_first_iteration` block, right alongside `recover_in_progress_pr_comments`.

The sweep:

1. `FidoStore.pending_pr_numbers(repo)` — new helper, returns distinct PR numbers with pending/retryable entries.
2. For each, call `gh.get_pr(repo, pr_number)`. If `state != "open"`, call `clear_pr_comment_queue(repo, pr_number)`.

In steady state (no orphans) this is one SQLite SELECT and zero API calls. After this lands, the next worker start will sweep up the existing 17 orphans.

## Trade-off considered

A complementary "drop-on-enqueue when payload PR state is closed" check would prevent the race at webhook time. Skipped here to keep this PR focused — the startup sweep covers both new and historical orphans, and the window before sweep is bounded by `WorkerThread` lifetime (Fido restarts often).

## Tests

- `test_pending_pr_numbers_lists_distinct_pr_numbers_with_pending_entries`
- `test_pending_pr_numbers_excludes_completed_entries`
- `TestSweepOrphanPrComments::test_clears_entries_for_closed_prs_and_keeps_open`
- `TestSweepOrphanPrComments::test_returns_zero_when_no_pending_entries`
- Updated `test_first_iteration_recovers_in_progress_pr_comments` to mock `gh.get_pr.return_value = {"state": "open"}` so the sweep doesn't clear the recovered entry.

## Coverage

100% maintained. Full suite green (4272 passed).

### DIFF
--- a/src/fido/github.py
+++ b/src/fido/github.py
@@ -1139,6 +1139,16 @@ query($owner: String!, $repo: String!, $number: Int!, $cursor: String) {
         data = self._get(f"/repos/{repo}/pulls/{pr}")
         return data.get("body") or ""
 
+    def get_pr_state(self, repo: str, pr: int | str) -> str:
+        """Return just the PR state (``"open"`` or ``"closed"``).
+
+        One ``_get`` call, no review/commit pagination — used by the orphan
+        comment-queue sweep (#1691) which only needs to know whether the PR
+        is still open.
+        """
+        data = self._get(f"/repos/{repo}/pulls/{pr}")
+        return data["state"]
+
     def add_pr_reviewers(self, repo: str, pr: int | str, reviewers: list[str]) -> None:
         """Request review from one or more users on a PR.
 
--- a/src/fido/store.py
+++ b/src/fido/store.py
@@ -647,6 +647,24 @@ class FidoStore:
         """Return whether *repo* has currently retryable queued comments."""
         return bool(self.pending_pr_comments(repo=repo, pr_number=pr_number))
 
+    def pending_pr_numbers(self, *, repo: str) -> list[int]:
+        """Return distinct PR numbers that have any pending or retryable
+        queue entries.  Used by the orphan sweep (#1691) to enumerate
+        candidates without loading their full payloads."""
+        self.ensure_schema()
+        with closing(self._connect()) as conn:
+            rows = conn.execute(
+                """
+                SELECT DISTINCT pr_number
+                FROM pr_comment_queue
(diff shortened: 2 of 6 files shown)


FidoCanCode:
Addressed both codex findings in b967773:

- **P1 (real bug)**: `sweep_orphan_pr_comments` was reading `pr_data["state"]` from `GitHub.get_pr`, but that method returns a curated dict with no top-level `state` field — it would never match `"open"` and the sweep would silently always-clear or KeyError. Caught before the sweep ever ran in production. Thanks for the assist.
- **P2**: same fix addresses it. New `GitHub.get_pr_state` does one `_get` call and returns just the state string — no review/commit pagination on PRs the sweep only needs open/closed for.

Tests now assert `get_pr` is not called and `get_pr_state` is the only round-trip.

FidoCanCode:
Addressed in 0ad1c58. Codex was right — the sweep was inside the post-`find_or_create_pr` first-iteration block, so idle repos (no current issue, sub-issue abandonment, all-covered terminal close) returned early before reaching it. Combined with `_is_first_iteration` flipping to `False` on the next attempt, the sweep would never run for an idle repo at all — exactly the orphan-accumulation case this PR was meant to fix.

Moved `sweep_orphan_pr_comments` to right after `create_session()`, above all the early returns. Left `recover_in_progress_pr_comments` where it is because it reconciles in-progress claim state for the PR the worker is about to drain, so gating it on having a PR is the right shape for that one.

New test `test_first_iteration_sweep_runs_for_idle_repo_with_no_issue` enqueues an orphan, returns `None` from both `get_current_issue` and `find_next_issue`, and asserts the queue is empty after `worker.run()`.

rhencke (approved):

Issue #1798, opened 2026-05-17
Epic: intent-as-instruction — stop materializing comments as tasks before rescope


FidoCanCode:
## Supersedes

- #1796 (rescope loses Opus's order)
- #1797 (synthesis encodes positional intent in task title)

Both are symptoms of one architectural mistake: the webhook path
synthesizes raw comment text into a task (``Tasks.add(title=prompt)``
inside ``events.create_task``) **before** rescope ever sees it.  Once
the intent is materialized as a task with the comment text as its
title, rescope has nothing to interpret — it's stuck reordering rows
whose titles already carry "make this first" / "instead of X" / etc.
verbatim, and the order/title decisions Opus is meant to make on the
intent are already lost.

## The right model

An **intent** is plain-English instruction text from a commenter
about the task list / PR / scope.  An intent is **not** a task.  One
intent might produce zero tasks ("nevermind"), one task, twelve
tasks, or reorder/merge/split/replace any subset of the current
tasks.  Only rescope (Opus, with full access to the PR, issue, code,
and current task list) can decide.

Intents arrive in batches (multiple commenters or one commenter
saying "red", then "no, green").  One rescope pass consumes the
whole batch.  Later intents override earlier intents on conflict.
The rescope output includes per-intent verdicts so the reply-back
layer can tell each commenter what happened — *unless* the
supersedence was by the same author, in which case no reply is
needed (they already know).

```
PR/issue comment
       │
       ▼
   synthesis (existing)
       │  emits RescopeIntent(comment_id, author, timestamp, text)
       │  posts triage reply ASAP (what Fido plans to DO) — unchanged
       │
       ▼
   rescope batches intents, drains on existing trigger
       │
       ▼
   Opus is given: current tasks + intent batch (raw text + author +
                  timestamp + comment_id, in arrival order) + access
                  to PR/issue/code
       │
       ▼
   Opus emits: ops on tasks (keep/new/merge/split/complete/rename) +
              per-intent verdicts
       │
       ▼
   apply ops → new task list (in Opus's order)
       │
       ▼
   reply-back: for each intent whose plan materially changed from
              what triage promised, post a follow-up reply quoting
              Opus's narrative.  Skip self-supersedence (same author).
```

## Decisions locked in

- **Kill ``events.create_task`` for thread comments entirely.**  The
  old abstraction from before we understood the problem.  SPEC tasks
  (setup / CI) still pre-materialize through ``Tasks.add`` — only the
[...]

Issue #1804, opened 2026-05-17
Leaf: Rocq model extension — intents + verdicts + supersedence are first-class


FidoCanCode:
The existing `models/task_queue_rescope.v` proves task-list lineage hygiene given an op sequence but does not model intents directly. Extend the model so the oracle proves the intent-level invariants:

- every intent in a batch produces exactly one verdict
- supersedence references are acyclic and only point within the batch
- every `affected_task_ids` matches an op actually applied for that intent
- delta attribution is total (no op without an intent of origin, no intent without ops or a no-op verdict)

**Touch**: `models/task_queue_rescope.v`, regenerated Python in `src/fido/rocq/`.

**Depends on**: INV-D (verdict shape settled before formalizing).

Source: FidoCanCode/home#1798

FidoCanCode:
## Updated scope (per #1798 Q&A — full rewrite, not extension)

Rewrite `models/task_queue_rescope.v` so the top-level transition is intent-first:

```
apply_intent_batch :
  list TaskRow → list Intent → (list TaskRow * list Verdict)
```

Ops become an internal intermediate type the proof traverses to relate intents to task mutations, not the top-level vocabulary.

**Theorems to prove**:

- `verdict_per_intent_unique`: every intent in the batch maps to exactly one verdict in the output
- `supersedence_in_batch`: `by_intent_comment_id`, when set, references an intent in the same batch
- `supersedence_acyclic`: the supersedence graph over verdicts is a DAG
- `attribution_total`: every applied op is attributed to exactly one verdict (the verdict whose ops list contains it); verdicts with empty ops have outcome in {superseded, no_op}
- `lineage_preserved`: existing lineage hygiene lifted into the intent-first vocabulary (merge sources, split children — every contributing intent stays linkable)
- `affected_task_ids_correct`: each verdict's `affected_task_ids` exactly covers the task ids its ops touched, plus any explicitly co-attributed task ids from peer verdicts

**Depends on**: INV-D verdict shape stable.

Source: FidoCanCode/home#1798

FidoCanCode:
Expanding scope to absorb INV-E (#1803).

[...]

PR #1818, opened 2026-05-18
INV-E (#1803): same-author self-supersedence suppresses rescope reply-back

### LINKED ISSUE #1803: Leaf: same-author self-supersedence suppresses rescope reply-back
When a single commenter says "red" then "green" in the same batch and Opus marks the first verdict superseded by the second (both same author), Fido must NOT post a reply explaining the supersedence — the commenter knows. Cross-author supersedence still gets a reply.

**Invariant**: reply-back for verdict V fires iff (V.outcome != honored OR no triage reply already covered the plan) AND NOT (V.by_intent_comment_id is set AND verdict_by_intent.author == V.author).

Triage reply (the immediate informed acknowledgment posted by synthesis) is unchanged — top-priority, "what Fido plans to DO". Rescope reply only fires when the plan materially changed from triage.

**Touch**: `_notify_thread_change` (or its successor), reply-back consumer.

**Depends on**: INV-D (verdicts must carry `by_intent_comment_id` and author).

Source: FidoCanCode/home#1798

FidoCanCode:
Closes #1803 (leaf of epic #1798).

## What

When one commenter says "red" then "green" in the same rescope batch and Opus marks the first verdict superseded by the second, Fido must NOT post a reply explaining the supersedence — they already know.  Cross-author supersedence still warrants a reply so the displaced reviewer hears that their ask got overridden.

## How

The per-iteration ``_on_intent_dispositions`` callback now receives the parsed ``IntentVerdict`` list alongside the dispositions and result task list.  ``_build_on_intent_dispositions`` checks each intent's verdict before falling through to the existing material/aggregation/unhandled rules:

- ``by_intent_comment_id`` set + same author → silent (INV-E).
- ``by_intent_comment_id`` set + different author → notify (bypasses the older "unhandled-with-only-later-attributed silent" rule because supersedence is a different relationship than implicit cover).
- No verdict / no ``by_intent_comment_id`` → existing rules unchanged.

Verdicts are captured across the parse-nudge loop in ``reorder_tasks`` so the final successful parse is what the callback sees.

## Test plan

- [x] ``./fido ci`` clean locally — all tests + 100% coverage.
- [x] New tests cover self-supersedence (silent) and cross-author supersedence (notifies).
- [ ] CI green on the PR.

### DIFF
--- a/src/fido/events.py
+++ b/src/fido/events.py
@@ -46,7 +46,7 @@ from fido.tasks import (
     Tasks,
     thread_comment_author_for_auto_resolve_oracle,
 )
-from fido.types import ActiveIssue, ActivePR, RescopeIntent, TaskType
+from fido.types import ActiveIssue, ActivePR, IntentVerdict, RescopeIntent, TaskType
 from fido.worker import ActivityReporter
 
 log = logging.getLogger(__name__)
@@ -2482,7 +2482,14 @@ def _build_on_intent_dispositions(
     gh: GitHub,
     agent: ProviderAgent | None,
     prompts: Prompts | None,
-) -> Callable[[dict[int, IntentDisposition], list[dict[str, Any]]], None]:
+) -> Callable[
+    [
+        dict[int, IntentDisposition],
+        list[dict[str, Any]],
+        list[IntentVerdict],
+    ],
+    None,
+]:
     """Build the per-iteration intent-dispositions callback (#1724).
 
     Each rescope iteration may carry a different ``current_intents``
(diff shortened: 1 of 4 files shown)


FidoCanCode:
Closing — codex caught a precedence bug (aggregation short-circuit eats cross-author supersedence) and the right fix isn't another Python branch.  Reply-back precedence is exactly the kind of subtle invariant that needs a Rocq model, not more ad-hoc if/elif in events.py.  Folding INV-E into INV-F (#1804) so the rule lands with proofs and the disposition/verdict interaction is modeled once, formally.

PR #1842, opened 2026-05-19
Leaf: forbid MagicMock with temporary exemptions (closes #1775)

### LINKED ISSUE #1775: Leaf: forbid MagicMock with temporary exemptions
Parent epic: #1773

## Invariant

No new `MagicMock` usage can enter the tree.

Dynamic mocks hide ownership and make constructor-DI migrations look complete while tests still depend on untyped behavior. New tests should use hand-rolled mock classes, fakes, spies, stubs, value objects, or typed collaborators.

A mock class is allowed when it has a small explicit interface and records only the behavior the test asserts. The forbidden part is generic dynamic mocking, not test doubles as a concept.

## Implementation

Add the smallest CI-enforced guard that rejects new `MagicMock` imports/usages. Existing dirty files may receive explicit temporary exemptions so this PR only locks the door.

Do not replace `MagicMock` with an equivalent dynamic mock as a bypass. If plain `Mock` is present as the same smell, file or split a sibling leaf before using it as the replacement.

## Out of scope

- Migrating existing `MagicMock` sites.
- Removing existing `unittest.mock.patch` or `monkeypatch.setattr` sites.

## Done

- CI rejects new `MagicMock` usage outside explicit temporary exemptions.
- The failure message points at typed mock classes/fakes and constructor-DI.
- `./fido ci` passes.

FidoCanCode:
Ban `unittest.mock.MagicMock` imports via ruff TID251 with `per-file-ignores` for the 29 existing exempted files, replacing the custom `tools/check_no_magicmock.py` script. Also includes system-exercise readme tasks requested by Rob.

Fixes #1775.

---

## Work queue

- [x] Remove tests/fakes.py from TID251 per-file-ignores 
- [x] Replace check_no_magicmock.py with ruff TID251 ban 
- [x] Add check_no_magicmock.py CI guard with temporary exemptions 
- [x] Remove 'hello' from the readme 
- [x] Add 'hello' to the readme 
- [x] Add 'hello' to the readme 
- [x] Remove 'hello' from the readme 
- [x] Change 'hello' to 'test' in the readme 
</details>


rhencke:
Your comment handling code has changed dramatically, so I'm going to lob you a bunch of comments/review comments that are not gonna make much sense, but we are using these to exercise the new intent system.

FidoCanCode:
Got it — signal received loud and clear. Fire away.

rhencke:
Add a task to put "hello" in the readme as the first task, and remove "hello" as the last task.  Keep both tasks even though the net effect is no effect.

FidoCanCode:
Queued: add 'hello' to readme as first task, remove 'hello' from readme as last task. Both tasks will appear even though they cancel out.

rhencke:
Remove all tasks related to the readme.

FidoCanCode:
Dropping both readme tasks — 'Add hello' and 'Remove hello' are gone.

rhencke:
Beautiful.  Now, let's try again: make a task that adds hello to the readme as the first task, so that it is the current task, and the current task drops one down.  Then add a task that removes hello as the last task.

FidoCanCode:
Queued: 'Add hello to readme' inserts first (becomes the next task, pushing the guard task down), 'Remove hello from readme' appended last.

rhencke:
Holy crap it works

FidoCanCode:
YES. That whole sequence — add, remove, re-add with ordering — all routed correctly. The intent pipeline is doing its job.

rhencke:
Okay, next I'm going to try some review comments.

FidoCanCode:
Ready. Bring them on.

README.md
```diff
@@ -1,5 +1,7 @@
 # Fido
 
+hello
```
rhencke: Instead of removing this let's change it to goodbye.

rhencke: Actually, make it say "test"

rhencke (changes requested):

README.md
FidoCanCode: Got it — I'll change 'hello' to 'test' instead of removing it.

FidoCanCode: Already queued — the task was rewritten to 'Change hello to test in the readme' from the first delivery of this comment.

rhencke:
Yesss

FidoCanCode:
Review comments routing correctly too. The whole pipeline held up.

rhencke:
Nice.  You can remove the hello from the readme so it's back to normal.

FidoCanCode:
On it — queuing up the removal so the readme goes back to normal.

models/Dockerfile
```diff
@@ -261,9 +261,10 @@ RUN ./pyproject ruff format --check .
 
 FROM python-ruff-base AS lint
 
-COPY tools/check_no_monkeypatch_setattr.py tools/
+COPY tools/check_no_monkeypatch_setattr.py tools/check_no_magicmock.py tools/
```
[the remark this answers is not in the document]
FidoCanCode: Correct diagnosis — the file landed in a follow-up commit (5268f28) because the harness didn't stage the untracked new files with the rest. Already fixed, CI is green now.

tests/test_check_no_magicmock.py
rhencke, line 1: Why is this necessary when we can just ban the API using ruff, like we did for patch?  Why do we need this file kind for patch check, too?

FidoCanCode: You're right — MagicMock is an import, so TID251 handles it the same way it handles `unittest.mock.patch`. The custom script is unnecessary overhead.

For `check_no_monkeypatch_setattr.py` it's a different shape: `.setattr(` is a call-site pattern from a pytest fixture, not an import, so TID251 doesn't reach it. But if there's a ruff rule that catches arbitrary call patterns I missed, happy to tear that one out too.

I'll rewrite the MagicMock ban to use TID251 with per-file-ignores for the 29 exemptions, and drop the custom script + tests + Dockerfile wiring.

FidoCanCode: Done — replaced `tools/check_no_magicmock.py` with a ruff TID251 banned-api entry for `unittest.mock.MagicMock`, using per-file-ignores for the 29 exempted files. Also removed the test file and Dockerfile wiring. The `check_no_monkeypatch_setattr.py` script can't be replaced by TID251 since it catches call-site patterns on a pytest fixture, not imports — filing that as a follow-up issue.

rhencke: Can we not just ban the MonkeyPatch class and update all pytest functions taking a monkeypatch with no type annotation to use that type annotation?

FidoCanCode: That's a sharper angle than I analyzed — banning `pytest.MonkeyPatch` as an import (TID251) plus ANN001 enforcement together does close the door: any new `monkeypatch` parameter needs a type annotation, and the only type is `pytest.MonkeyPatch`, which requires an import that's banned.

The tradeoff: it requires adding `monkeypatch: pytest.MonkeyPatch` annotations to all ~14 existing files, plus per-file-ignores for the import. The custom script is simpler and already working. But the approach is sound.

I filed a follow-up issue to track the investigation — I'll update it with this angle so the comparison is there when we pick it up.

pyproject.ruff.toml
```diff
@@ -261,21 +267,41 @@ ignore = ["E501"]
   "ANN001", "ANN002", "ANN003", "ANN201", "ANN202", "ANN204", "ANN206", "ANN401",
   "E741", "F401", "F403", "F405", "F541", "F811", "F821", "F841", "I001",
 ]
-# Temporary TID251 exemptions — existing mock.patch sites pending constructor-DI
-# migration under #1773. Do not add new files here; fix the root cause instead.
+# Temporary TID251 exemptions — existing mock.patch and MagicMock sites pending
+# constructor-DI migration under #1773. Do not add new files here; fix the root
+# cause instead.
+"tests/fakes.py" = ["TID251"]
```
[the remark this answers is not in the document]
FidoCanCode: Good catch — accepted. `tests/fakes.py` only mentions MagicMock in a docstring. The original custom script used a broad regex that caught docstring mentions, which is why the file ended up in the exemptions list. TID251 fires on imports only, so the exemption is unnecessary. I'll remove it.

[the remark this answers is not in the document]
FidoCanCode: Removed `tests/fakes.py` from the TID251 per-file-ignores list — ruff confirms it's clean without the exemption. One less file on the naughty list!

rhencke (approved):

Issue #1874, opened 2026-05-20
Insight: Insight: typed defaults are still defaults


FidoCanCode:
Using `param: Protocol = _DEFAULT_IMPL` looks cleaner than `| None = None` but is the same anti-pattern in a nicer coat.

The whole point of constructor-DI is making dependencies visible at every call site. A default — even a well-typed one — lets callers skip the decision, which means the dependency is implicit again. The composition root already passes explicitly; the default only serves test convenience, and that convenience is exactly the thing that hides which implementation is running. Rob caught this immediately.

Source: FidoCanCode/home#1858

Source: https://github.com/FidoCanCode/home/pull/1858#discussion_r3277659603

PR #1890, opened 2026-05-23
Leaf: replace MagicMock test doubles with typed fakes (closes #1778)

### LINKED ISSUE #1778: Leaf: replace MagicMock test doubles with typed fakes
Parent epic: #1773

## Invariant

The tree has zero remaining `MagicMock` test doubles.

A test double should have the smallest explicit interface needed by the test. If the test needs call recording, the mock/fake class owns a typed call log instead of relying on dynamic mock internals.

## Implementation

Drain the temporary `MagicMock` exemptions. Prefer hand-rolled mock classes, fakes, spies, stubs, simple value objects, and Protocol-backed collaborators.

If a `MagicMock` is compensating for a missing class boundary, introduce that boundary rather than preserving a generic fake object.

Method override is acceptable inside the injected test-double class when that is the cleanest way to express the seam.

## Out of scope

- Replacing unrelated mocks not covered by the MagicMock guard unless they are direct bypasses.
- Broad SRP splits not required by the MagicMock migration.

## Done

- Grep/CI finds zero `MagicMock` usage.
- Replacements are typed enough that refactors fail loudly.
- `./fido ci` passes.

FidoCanCode:
Drained all ~3,350 `MagicMock` occurrences across 27 test files, replacing each with hand-rolled typed fakes, spies, stubs, or value objects. Removed every per-file `TID251` exemption from `pyproject.ruff.toml` and introduced missing class boundaries where a `MagicMock` was compensating for one. Shared fake infrastructure (`_CallRecord`, `_FakeCallable`) consolidated in `tests/fakes.py` for cross-file reuse.

Fixes #1778.

---

## Work queue

<details><summary>Completed (13)</summary>

- [x] Replace MagicMock in test_github 
- [x] Replace MagicMock in test_coverage_fills 
- [x] Replace MagicMock in test_registry 
- [x] Replace MagicMock in test_tasks 
- [x] Replace MagicMock in test_copilotcli and test_status 
- [x] Replace MagicMock in test_cli, test_claude_hold_for_handler, test_worker_persist_session_id 
- [x] Replace MagicMock in synthesis, codex, and session test files 
- [x] Replace MagicMock in watchdog and FSM oracle test files 
- [x] Replace MagicMock in small provider and misc test files 
- [x] Replace MagicMock in test_claude 
- [x] Replace MagicMock in test_events 
- [x] Replace MagicMock in test_server 
- [x] Replace MagicMock in test_worker 
</details>


tests/test_tasks.py
```diff
@@ -40,15 +37,134 @@
...
+        # call(…) objects as well as bare tuples.
+        try:
+            n = len(other)  # type: ignore[arg-type]
+        except TypeError:
+            return NotImplemented  # type: ignore[return-value]
+        try:
+            if n == 3:
+                return self.args == other[1] and self.kwargs == other[2]  # type: ignore[index]
+            return self.args == other[0] and self.kwargs == other[1]  # type: ignore[index]
+        except TypeError, IndexError:
```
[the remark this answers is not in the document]
FidoCanCode: This is actually valid Python 3.14 syntax — PEP 758 removed the parenthesis requirement for multi-exception `except` clauses. `except TypeError, IndexError:` now means "catch either TypeError or IndexError" (not the old Python 2 binding semantics). The codebase targets 3.14t exclusively, and the same pattern is already used in production code (`worker.py:438`, `worker.py:5304`). No change needed.

tests/test_claude_hold_for_handler.py
```diff
@@ -1,29 +1,107 @@
...
+    def __call__(self, data: str) -> None:
+        self.call_args_list.append(_WriteCallRecord((data,)))
+
+
+class _FakeStdin:
+    """Fake process stdin: records writes, exposes ``closed`` flag."""
+
+    def __init__(self) -> None:
+        self.closed: bool = False
+        self.write: _FakeStdinWrite = _FakeStdinWrite()
```
rhencke, line 38: Remember what you learned in #1874?

rhencke (changes requested):

tests/test_claude_hold_for_handler.py
FidoCanCode: Oof — I think I do. `_WriteCallRecord` and `_FakeStdinWrite` are just less-capable copies of `_CallRecord` and `_FakeCallable` from `tests/fakes.py`. I should use the shared infrastructure instead of reinventing it here.

`_FakeStdin.write` becomes `_FakeCallable()`, the `_WriteCallRecord` class goes away entirely, and the usage at line 478 (`c.args[0]`) stays the same because `_CallRecord.args` works identically.

I'll fix this up.

rhencke: Not quite.  Specifically, default values in the constructor for the typed dependency makes the dependency implicit and hidden, instead of having to explicitly declare via DI.

So dont do:
```python
class X:
    def __init__(self, y = Y()):
        ...

realX = X()

testX = X(y = MockY())
```

Instead, do:

```python
class X:
    def __init__(self, y):
        ...

realX = X(y = Y())
# or, break it down, especially if the dependency is shared over different instances
y = Y()
realX = X(y = y)

testX = X(y = MockY())
```

FidoCanCode: Got it — I was barking up the wrong tree. The issue isn't which recording class to use, it's that `_FakeStdin` creates `_FakeStdinWrite()` internally instead of requiring it as a parameter. That hides the dependency.

Same pattern in `_FakeProc`, `_RecordingProc`, and `_StreamingForeverProc` — they all hardcode their sub-fakes internally. I'll make those required constructor parameters so every dependency is visible at the call site.

rhencke: Perfect, thanks.

FidoCanCode: 🐕

rhencke: > Not quite.  Specifically, default values in the constructor for the typed dependency makes the dependency implicit and hidden, instead of having to explicitly declare via DI.
> 
> So dont do:
> ```python
> class X:
>     def __init__(self, y = Y()):
>         ...
> 
> realX = X()
> 
> testX = X(y = MockY())
> ```
> 
> Instead, do:
> 
> ```python
> class X:
>     def __init__(self, y):
>         ...
> 
> realX = X(y = Y())
> # or, break it down, especially if the dependency is shared over different instances
> y = Y()
> realX = X(y = y)
> 
> testX = X(y = MockY())
> ```

Please do this work on this PR by the intent mechanism

FidoCanCode: On it — I'll apply constructor-DI to `_FakeStdin`, `_FakeProc`, `_RecordingProc`, and `_StreamingForeverProc` so their sub-fakes are required constructor parameters, not internally created defaults.

rhencke (approved):

Issue #1894, opened 2026-05-23
Epic: Fido can't lie about durable state — holistic gate architecture across 5 layers


FidoCanCode:
## Problem

PR #1858, PR #1890, and the cluster of #1855 / #1861 / #1862 / #1875 / #1872 / #1873 / #1878 all share one root: **Fido emits prose, ops, or status that doesn't match durable reality, and there is no gate to catch the drift before it ships.**

The failures showed up at every emission point:

| Symptom | Emission point | What drifted |
|---------|---------------|--------------|
| PR #1890: rescope verdict says ``no_op``, silently drops Rob's ask | rescope verdict | verdict doesn't relay outcome back to requestor |
| #1862: "I'll do X and audit Y" reply, only one becomes a task | triage reply | prose promises ≠ registered intents |
| #1855: reply says "I filed an issue" with no URL | reply prose | prose claims artifact that's not linkable |
| #1861: follow-up comment spawns parallel duplicate task | rescope ``new`` op | new task duplicates existing scope |
| PR #1858 (13-task survey): tasks marked completed without any code change | task ``complete`` op | worker claim ≠ git state |
| #1875: solo comment queued behind unrelated batch never gets eyes | webhook ack | predicate too coarse |
| PR #1858 (PR-body claims): "the work is in a recent commit" prose lies | reply prose | prose claims state that doesn't exist |

#1871 had the right principle ("critic at every emission") but no structural commitment — it sketched 5 slices and an extension surface but didn't model the lifecycle, the per-thread aggregation, the decomposition contract, or the gate layering.  After a 20-question planning session with Rob, this epic captures the holistic shape and supersedes #1871.

## Architecture (four layers)

\`\`\`
┌──────────────────────────────────────────────────────────────┐
│ Layer 4: Decomposition contract                                │
│   One task = one commit = one statable invariant               │
├──────────────────────────────────────────────────────────────┤
│ Layer 3: Visibility / acknowledgment                           │
│   Eyes → triage reply → material-divergence → terminal aggregate│
│   All Fido voice, all critic-loop verified                     │
├──────────────────────────────────────────────────────────────┤
│ Layer 2: Durable state shape (intent → verdict → task DAG)     │
│   intent / task nodes; typed-edge verdicts carry narrative     │
│   no_op edges terminate at SKIPPED marker tasks                │
├──────────────────────────────────────────────────────────────┤
│ Layer 1: Critic gates (per emission point) — LLM-only          │
[...]

PR #1929, opened 2026-05-23
HOL-3 (#1897): require non-empty narrative on every IntentVerdict outcome

### LINKED ISSUE #1897: HOL-3 (#1894): Rocq verdict schema requires non-empty narrative on every outcome
Sub-issue of #1894.

**Invariant**: `_parse_rescope_verdicts` in `src/fido/tasks.py` and the Rocq verdict schema (`models/task_queue_rescope.v`) reject any verdict whose `narrative` is None/empty — for `honored`, `reshaped`, `superseded`, AND `no_op`.

Today narrative is only enforced for reshaped/superseded; today's PR #1890 bug fired because no_op had no narrative requirement.

**Acceptance**: parser rejects missing narrative; reducer refuses batch commit; regression test covers each outcome's missing-narrative case.

**Open**: migration of existing in-memory verdict log — backfill empty narratives or skip-render? Default skip-render; reducer enforces only on new verdicts.

## Cross-refs

- #1894 — parent epic

FidoCanCode:
## Summary

Closes #1897.  First implementation slice of epic #1894.

Tightens ``IntentVerdict.__post_init__`` so the non-empty narrative check fires on every outcome (``honored``, ``reshaped``, ``superseded``, ``no_op``), not just reshaped/superseded.  Directly closes the PR #1890 no_op-silent-drop failure mode at the data layer: Opus can no longer emit a verdict that drops an intent without a written reason.

Also updates the rescope_prompt_verdicts and parser nudge prompts so Opus understands narrative is required everywhere, with explicit per-outcome reasoning.

Parser change is structural — ``_parse_rescope_verdicts`` already catches every TypeError/ValueError from the ctor, so the tightening propagates automatically.

## Test plan

- [x] New tests pin the rule per outcome: ``test_honored_outcome_requires_narrative``, ``test_no_op_outcome_requires_narrative``, plus whitespace-only variants
- [x] Two prior "allows blank narrative" tests replaced (they documented the bug)
- [x] Existing IntentVerdict ctors updated to pass narrative where the test relies on the ctor SUCCEEDING; ctors whose tests check earlier validation rules stay narrative-less so the test still fires on the intended rule
- [x] Parser-side fixtures in ``TestParseRescopeVerdicts`` and reorder-wiring tests updated to include ``narrative`` on every verdict
- [x] ``test_outcome_semantics_distinguish_reply_back`` updated to assert the new "Narrative REQUIRED on every outcome" prompt text
- [x] ``./fido ci`` clean — 4527 tests, 100% coverage

## Follow-ups

HOL-5 (``TaskStatus.SKIPPED``) and HOL-6 (no_op → skipped marker task) build on this; they're what makes the no_op narrative actually visible as a queue artifact.

### DIFF
--- a/src/fido/prompts.py
+++ b/src/fido/prompts.py
@@ -660,26 +660,31 @@ class Prompts:
             '    "ops": [<op record>, ...],\n'
             '    "affected_task_ids": ["<task-id>", ...],\n'
             '    "by_intent_comment_id": <int | null>,\n'
-            '    "narrative": "<prose explanation>" | null\n'
+            '    "narrative": "<prose explanation>"   // REQUIRED on every outcome (HOL-3 / #1897)\n'
             "  }\n\n"
             "Outcome semantics (drives whether INV-E posts a follow-up "
             "reply-back):\n"
             "  honored    — ops fulfill the intent as asked.  Triage "
-            "already replied, no follow-up.\n"
+            "already replied, no follow-up.  Narrative still REQUIRED — "
+            "it lands in the per-task narrative chain so future rescopes "
+            "see why each task got the work it has.\n"
             "  reshaped   — ops fulfill part of the intent, framing "
             "changed (e.g. split into prereq + feature).  Material — "
-            "reply-back required.  Narrative MUST be non-empty.\n"
+            "reply-back required.  Narrative REQUIRED.\n"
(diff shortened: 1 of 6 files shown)


FidoCanCode:
Consolidating HOL-N slices into one PR per Rob's direction — single PR, one commit per leaf.

Issue #1915, opened 2026-05-23
HOL-21 (#1894): Unified stuck-on-critic sentinel for HOL-16..HOL-19


FidoCanCode:
Sub-issue of #1894.

**Invariant**: same routing as HOL-20 for the other 4 critics (task-creation, task-completion, reply-prose, insight-filing). One sentinel shape: `stuck-on-critic(emission_point, attempts)`.

Depends on HOL-16/17/18/19.

**Acceptance**: each critic's exhaustion path tested; all route to BLOCKED + auto-bug.

## Cross-refs

- #1894 — parent epic

PR #1933, opened 2026-05-24
HOL-21 of #1894: unified stuck-on-critic sentinel (closes #1915)

FidoCanCode:
## Summary

HOL-21 / #1915: unified ``stuck-on-critic`` sentinel infrastructure + the one critic that has a real indefinite-cycle problem in practice.

**Phase 1 (commit ``28d7e74``)** — extract shared ``file_stuck_on_critic_bug`` helper out of HOL-20's ``_route_critic_exhausted_blocked``.  ``StuckOnCriticContext`` carries critic-agnostic fields; idempotency marker keyed on ``(emission_point, repo, pr, target_kind, target_id)`` so two different critics on the same target file separate bugs.  Pure refactor — HOL-15/18 routing unchanged.

**Phase 2 (commit ``c36e64f``)** — HOL-17 task-completion re-park budget.  Before this commit a task whose completion-critic kept rejecting it would cycle indefinitely (each re-park appended ``CRITIC: <gap>`` to the description and reset to IN_PROGRESS).  Now bounded: after ``_HOL17_MAX_REPARK_BUDGET`` (=3) failures, the escalation helper marks the task BLOCKED, clears ``current_task_id``, files a bug via the shared helper, and posts a BLOCKED comment on the source PR.  Idempotent — repeated escalation on the same task reuses the bug rather than spamming.

## Why no Phase 3 / Phase 4 for HOL-16 + HOL-19

- **HOL-16 (task-creation critic)** runs per-item in a rescope batch.  There's no natural cross-batch "exhaustion" — each batch is one Opus call.  A real retry budget would need persistent per-``intent_comment_id`` counters (SQLite / state.json), which is non-trivial new infrastructure for a problem we don't currently observe in production.  The HOL-16 critic already fails-open per call.
- **HOL-19 (insight-filing critic)** is short-circuited by the per-comment idempotency marker before the critic runs on a replay.  "Exhaustion" doesn't fit the access pattern — once the marker is recorded, the critic doesn't see the same input twice.

Both are filed as follow-up issues so a future leaf can add persistent counters if/when the indefinite-cycle pattern actually appears in production.  The shared ``StuckOnCriticContext`` + ``file_stuck_on_critic_bug`` infrastructure (Phase 1) is in place so those leaves are small when they land.

## Test plan

- [x] ``./fido ci`` green, 100% coverage, 4846 tests pass.
- [x] Phase 1: 6 ``TestFileStuckOnCriticBug`` tests covering helper contract (new-bug, marker-hit reuse, search-failure, create-failure, emission-point in key, attempts-preview rendering).
- [x] Phase 2: ``test_task_completion_critic_exhaustion_escalates_to_blocked`` (full escalation path), ``test_task_completion_critic_repark_just_below_budget`` (off-by-one), ``test_task_completion_critic_exhaustion_handles_helper_failures`` (transient GitHub failures don't undo local BLOCKED), ``test_task_completion_critic_exhaustion_without_pr_context`` (legacy callers).
- [x] HOL-20 routing tests still pass (Phase 1's refactor preserves HOL-15/18 behaviour).

### DIFF
--- a/src/fido/events.py
+++ b/src/fido/events.py
@@ -412,6 +412,131 @@ _BUG_REPO = "FidoCanCode/home"
 _BUG_LABEL = "Bug"
 
 
+@dataclass(frozen=True)
+class StuckOnCriticContext:
+    """HOL-21 / #1915: context bundle for ``_file_stuck_on_critic_bug``.
+
+    Every critic that detects exhaustion (HOL-15..HOL-19) builds one
+    of these and hands it to the shared filing helper.  The shape is
+    deliberately critic-agnostic so the bug body, idempotency
+    marker, and title template are all uniform across emission
+    points — the auto-filed bug is recognizably the same shape
+    whether it came from intent-coverage exhaustion on a triage
+    reply or task-completion exhaustion on a worker commit.
+
+    Fields:
+
+    - ``emission_point``: short label identifying which critic
+      exhausted (``"intent-coverage"``, ``"task-completion"``, etc.).
+      Part of the idempotency key so two different critics
+      exhausting on the same target still file separate bugs.
+    - ``source_repo`` / ``source_pr``: where the exhaustion happened.
+    - ``target_kind`` / ``target_id``: what the critic was working on
+      when it exhausted (``"comment"``/``"task"``).  Part of the
+      idempotency key.
+    - ``source_link``: clickable URL to the target.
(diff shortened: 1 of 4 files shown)


FidoCanCode:
Combined with #1937 into PR #TBD (next URL).  Same commits + codex fixes.

Issue #1920, opened 2026-05-23
HOL-26 (#1894): Insights filed BEFORE synthesis reply prose composition (closes #1855)


FidoCanCode:
Sub-issue of #1894.

**Invariant**: in `SynthesisExecutor.execute`, insights are filed BEFORE the synthesis call generates reply prose. The synthesis prompt input includes `insights_with_urls[]` so the LLM can include URLs in the prose by construction.

Closes #1855 at the sequencing layer; HOL-18 closes it at the verification layer.

**Open**: insight filing latency cap (default 3s with fallback prose 'insight filing in flight').

**Acceptance**: regression test — synthesis sees insight URLs in prompt input; reply prose includes them.

## Cross-refs

- #1894 — parent epic

FidoCanCode:
**Deferring this leaf** — the implementation requires splitting the synthesis prompt into two LLM calls (first to propose insights, file them, then reply-prose call with the URLs in the prompt input).  That's a substantial architectural change to ``SynthesisExecutor.execute``.

Per the issue body, ``#1855`` is already closed at the **verification layer** by HOL-18's reply-prose claim-grounding critic — the critic rejects prose that names insights it can't verify, including ones that lack URLs.  The sequencing-layer fix HOL-26 proposes would be more robust by construction (no critic round-trips needed) but isn't blocking the epic.

Leaving HOL-26 open as a hardening leaf to revisit if HOL-18's verification proves insufficient in production.  Closing as deferred for now; #1855 stays closed via HOL-18.

Issue #1855, opened 2026-05-19
Bug: synthesis reply doesn't link the Insight issue(s) Fido files


FidoCanCode:
## Bug

When Fido replies to a comment and ALSO files an Insight issue from the same synthesis turn, the reply prose mentions "I filed a follow-up issue" but doesn't include the issue URL.

## Concrete evidence

[Reply on PR #1842](https://github.com/FidoCanCode/home/pull/1842#discussion_r3268956915) (comment ``3268956915``, 2026-05-19T19:14:37Z):

> ... I filed a follow-up issue to track the investigation — I'll update it with this angle so the comparison is there when we pick it up.

No URL.  Reader has no way to find the filed issue without searching.

## Root cause

The synthesis turn produces ``CommentResponse(reply_text, ..., insights=[...])`` as a single response — the reply text is generated at the same time as the list of insights to file, BEFORE either is acted on.  Then ``SynthesisExecutor`` posts the reply first and files the insights afterward (``execute_effects_only`` order: remove eyes → emoji → rescope → file insights).  The reply prose can't carry a URL the executor doesn't know yet.

Relevant code:
- ``src/fido/synthesis_executor.py:144-213`` — ``execute`` / ``execute_effects_only`` ordering.
- ``src/fido/events.py:121-152`` — ``_GitHubInsightFiler.file_insight`` creates the issue and gets the URL back; nothing currently does anything with that URL.

## Fix options

1. **File insights first, then post reply with substituted URLs.**  Require the LLM to emit a placeholder (e.g. ``{insight:0:url}``) in ``reply_text``; executor files insights, substitutes URLs into ``reply_text``, then posts.  Keeps one reply, lets the LLM control where the URL lands in prose.  Adds a per-call contract the LLM has to honor.
2. **Append a mechanical footer.**  Executor files insights and appends ``\n\n*Filed: <url1>, <url2>*`` to the reply text before posting.  No LLM contract; URL is always present but landing is structural, not prose-integrated.
3. **Post a follow-up edit.**  After filing, edit the original reply via the GitHub API to inject URLs.  Two API calls and an audit-log entry per reply; works for both review and issue comments.
4. **Post a follow-up comment.**  File then post a second reply with the URLs.  Simplest to implement but pollutes the thread with a second comment per insight-bearing reply.

[...]

PR #1941, opened 2026-05-24
HOL-26 of #1894: file insights before posting reply prose (closes #1920)

FidoCanCode:
## Summary

- `SynthesisExecutor.execute()` now files insights BEFORE `_post_reply`, so the durable GitHub issues exist before any user-visible reply can claim they do.
- `execute_effects_only()` gains a keyword-only `insights_already_filed` flag (default `False`) so the pre-reply path doesn't double-file when callers route through `execute()`.
- Added ordering test `test_execute_files_insights_before_posting_reply` that pins the file-then-post sequence and the no-double-fire invariant.

## What this fix covers vs defers

Sequencing-layer fix for the #1855 class of bug. The full prompt-input weaving (the synthesis prompt sees `insights_with_urls[]` so the LLM can land URLs in prose by construction) requires splitting synthesis into two LLM calls — a larger restructure best done as its own slice when it becomes the bottleneck. HOL-18's reply-prose claim-grounding critic remains the verification-layer backstop for URL-in-prose.

## Test plan

- [x] `./fido tests tests/test_synthesis_executor.py` — 42 passed, 100% file coverage
- [x] `./fido ci` (pre-commit) — passed end-to-end

### DIFF
--- a/src/fido/synthesis_executor.py
+++ b/src/fido/synthesis_executor.py
@@ -148,21 +148,27 @@ class SynthesisExecutor:
     ) -> ReviewReplyOutcome:
         """Post the reply and run all post-reply effects in one call.
 
-        Thin convenience wrapper: posts the reply via :meth:`_post_reply`,
-        then delegates to :meth:`execute_effects_only` for the eyes /
-        emoji / rescope / insights / outcome chain.  Production callers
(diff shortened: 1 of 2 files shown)


FidoCanCode:
Superseded — consolidated into PR #1938 (the single epic PR per Rob's direction). Branch retained for reference.

